PR #172, opened 2013-09-20
Update soap.js WSSecurity to use toISOString method on dates

tracker1:
Date.prototype.toISOString() is available in the node.js execution environment and should be used in favor of the formatting method that was used.


herom:

PR #176, opened 2013-09-30
Added check for data types, valid SOAP error message

lbox-admin:
- Allow undefined/empty parameter only if attribute nillable is set to true
- Check elements for correct data types
- Send valid SOAP message with error


herom:

PR #190, opened 2013-11-26
Optionally use funcback pattern if callback is omitted

MatAtBread:
Create remote methods with "funcbank" pattern to reduce code complexity and make available for use with NoDent.

The pattern is optional (only used if "callback" is omitted) and used only if the callback is NOT specified to the rmeote method.

Note: This also fixes a bug which lost the "options" if the "args" were omitted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+/.project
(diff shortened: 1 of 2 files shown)


herom:

PR #209, opened 2014-01-14
 Fixed: error while there is no connection to url, which may cause read  property "headers" of undefined variable "response".

PavelVdovin:
The error may be reproduced in the following way.
The client is created from the .wsdl file in local file system. Then the endpoint is specified (with functionality provided by the local wsdl) and the client function is called. If there is no connection to the url specified, the variable response is undefined, and instead of calling callback, the exception is thrown (cannot read property "headers" of undefined).


herom:

PR #210, opened 2014-01-15
Add support for xml attributes in requests and responses

evs-chris:
I have a few soap services that I access that require attributes instead of nested elements. This commit sets any field in an argument object that starts with '@' as an attribute on its parent element instead of as a nested element. It also makes any attributes that come back in response available in the response object.

I'm not positive that this is the best way to handle this, but it seemed much better than having a blessed name, like attrs, that may overlap with actual parameters.

In summary:
```javascript
client.someMethod({ name: 'value', '@attr': 'attrval' }, function(err, response) { });
```
has a body with
```xml
<urn:someMethod attr="attrval"><urn:name>value</urn:name></urn:someMethod>
```
instead of the current (sans @)
```xml
<urn:someMethod><urn:attr>attrval</urn:attr><urn:name>value</urn:name></urn:someMethod>
```

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -736,9 +736,6 @@ WSDL.prototype.xmlToObject = function(xml) {
                self.definitions.messages[originalName] = self.definitions.messages[name];
             }
 
-            // since objects have no special attribute area, stick attributes straight into the obj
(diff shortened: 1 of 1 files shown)


evs-chris:
Looks like this has been fixed for a while.

PR #214, opened 2014-01-18
Add an option to disable time stamp element generation in WSSecurity

ethankao:
The module works great, until I use it to access charge point web services.

I'd like to add an option to disable the timestamp generation in WSSecurity XML. 

When I use this module to access charge point web service, their service is not able to process the timestamp. 

The service wsdl is here, https://webservices.chargepoint.com/cp_api_4.1.wsdl


herom:

PR #232, opened 2014-01-27
export open_wsdl and add a keepText option so text in WSDL will be kept

jchip:


herom:

Issue #206, opened 2014-01-06
A pure Javascript solution anybody?


christiaanwesterbeek:
Node-soap depends on Node-expat that requires node-gyp. Node-gyp requires you to install Visual Studio on your Windows machine. If your machine is a Windows Server, you definitely don't want to install Visual Studio on it. Seems like Mac users aren't always freed from trouble getting node-gyp to work.

This issue is to propose a replacement of the node-expat dependency with sax.js. Sax doesn't require node-gyp. Effectively making node-soap a pure Javascript solution. https://github.com/isaacs/sax-js

I have created a working fork that implements this. https://github.com/devotis/node-soap/tree/feature-sax

PR #246, opened 2014-02-03
Replace Expat dependency with Sax

christiaanwesterbeek:
@jsdevel I followed your steps. Here's the pull request for my feature branch that's based on the current master of milewise/node-soap. Here I replaced the expat dependency with sax. Refs issue #206

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -7,7 +7,7 @@
 
 "use strict";
 
-var expat = require('node-expat');
+var sax = require('sax');
 var inherits = require('util').inherits;
 var http = require('./http');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


wprater:
why was this replaced? we are having performance issues with SAX and Im curious why node-expat was removed?

herom:
As you may have recognized, this PR is more than 2 years old - I don't know exactly why this was changed, but there are chances that the performance issues you're seeing were maybe not present then... 

Which performance issues have you tracked while using the `sax` package?

wprater:
I assume it was probably less about performance and more about not using a native dependency?

On our larger SOAP responses, performance of saxjs is more than 5 times slower when compared to easysax or node-expat.  We've tried both and now are using a fork with easysax!

`xmlToObject`
saxjs: 98.2ms
node-expat: 29.9ms
saxjs: 17.1ms

herom:
This is really a huge performance boost :+1: Are you interessted in contributing and send in a PR with the replaced `easysax` dependency? This would be awesome :dancer:

Issue #348, opened 2014-05-13
`input.$type` is undefined in `Client.prototype._invoke`


phGitUser:
I have a problem with the `$type` parameter of my messages: 

When my `WSDL` looks like this: 

dummy.wsdl: 

``` wsdl
<?xml version="1.0" encoding="UTF-8"?>
<wsdl:definitions xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/" 
xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/" 
xmlns:xs="http://www.w3.org/2001/XMLSchema" 
xmlns:tns="http://www.Dummy.com"  xmlns:n="http://www.Dummy.com/Name/Types"  xmlns:ns="http://schemas.xmlsoap.org/soap/encoding/" targetNamespace="http://www.Dummy.com">
    <wsdl:types>
        <xs:schema>
            <xs:import namespace="http://www.Dummy.com/Common/Types" schemaLocation="./Common.xsd"/>
            <xs:import namespace="http://www.Dummy.com/Name/Types" schemaLocation="./Name.xsd"/>
        </xs:schema>
    </wsdl:types>
    <wsdl:message name="DummyRequest">
        <wsdl:part name="DummyRequest" element="n:DummyRequest"/>
    </wsdl:message>
    <wsdl:message name="DummyResponse">
        <wsdl:part name="DummyResponse" element="n:DummyResponse"/>
    </wsdl:message>
    <wsdl:portType name="DummyPortType">
        <wsdl:operation name="Dummy">
            <wsdl:input message="tns:DummyRequest"/>
            <wsdl:output message="tns:DummyResponse"/>
        </wsdl:operation>
    </wsdl:portType>
    <wsdl:binding name="DummyBinding" type="tns:DummyPortType">
        <soap:binding style="document" transport="http://schemas.xmlsoap.org/soap/http"/>
        <wsdl:operation name="Dummy">
            <soap:operation soapAction="http://www.Dummy.com#Dummy" style="document"/>
            <wsdl:input>
                <soap:body use="literal"/>
            </wsdl:input>
            <wsdl:output>
                <soap:body use="literal"/>
            </wsdl:output>
        </wsdl:operation>
    </wsdl:binding>
    <wsdl:service name="DummyService">
        <wsdl:port name="DummyPortType" binding="tns:DummyBinding">
            <soap:address location="http://www.Dummy.com/"/>
        </wsdl:port>
    </wsdl:service>
[...]

PR #357, opened 2014-05-22
Introduced `$lookupType` and `$lookupTypes` on a `message` `part` elemen...

herom:
...t which has declared an `element` instead of a `type`.

Within `Client.prototype._invoke()` the `input.$lookupType`  property will be used if `input.$type` is not available, providing a first hook for further parameter look ups to move in the right direction and resolve the desired property with the already existing/implemented code.

The `$inputLooupTypes` `Array` provides an `object` for each referenced `complexType` along the line of possible available children, setting its `$name` `$type` and `$namespace` attributes so that `WSDL.prototype.findParameterObject()` and `WSDL.prototype.findChildParameterObject()` could be used without changing the way how all other `object`s are processed.

This pull request closes https://github.com/vpulim/node-soap/issues/348 without introducing the discussed `DeferredCache` implementation, as everything what's needed is already read and available in the`MessageElement.prototype.postProcess` method.


mmcarolos:
I am having some issues with this fix not working properly for a service I am trying to use. I believe there is a mistake in the code and the reason why the test passes is because message's part has a matching element and name value. For example in the UpdateProfile__correct_nanespaces_in_sequence_with_imports test the soap.wsdl file has the following message:

```xml
    <wsdl:message name="UpdateProfileRequest">
        <wsdl:part name="UpdateProfileRequest" element="n:UpdateProfileRequest"/>
    </wsdl:message>
```

Note that the name and element are UpdateProfileRequest which makes `WSDL.prototype.findParameterObject` method get the correct `parameterTypeObj` from `schema.elements.UpdateProfileRequest` using `$lookupType`.

Now if for example the the message's part had a different name such as the following example:

```xml
    <wsdl:message name="UpdateProfileRequest">
        <wsdl:part name="parameter" element="n:UpdateProfileRequest"/>
    </wsdl:message>
```

Then the `WSDL.prototype.findParameterObject` method no longer gets the correct `parameterTypeObj` using `$lookupType` since `schema.elements.parameter` does not exist since it is still named `schema.elements.UpdateProfileRequest`.

It looks like changing the method `MessageElement.prototype.postProcess` method that sets `$lookupType` from:

```javascript
this.element.$lookupType = part.$name;
```

to:
```javascript
this.element.$lookupType = nsName.name;
```

will make it work correctly in both cases.

Is there something I am missing here or does this sound right and I should do a pull request? Sorry if I was wrong in posting on this thread instead of creating a new one.

herom:
Thanks for your time and effort to track this issue down @mmcarolos - it would be awesome if you could provide a PR with your fix (please see our [Guidelines on "Submitting a Pull Request"](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request)) :+1:

PR #413, opened 2014-07-09
Look up the parameterTypeObject based on method name

wankdanker:
In `WSDL.findParameterObject`, if the `parameterTypeObj` is not located by the
`parameterType` in either the `complexTypes` or `elements` objects, then also
check to see if it can be found by the name of the method in the `elements`
object.


herom:
I'm closing this PR as there was no activity for nearly a year now...

PR #425, opened 2014-07-30
Support authenticate async callback 

aduyng:
Change method authenticate(security) to authenticate(security, callback) to support async logic (e.x. checking username/password stored in database).
```javascript
soap = server.listen(....);
soap.authenticate = function(security, callback) {
    'use strict';

    var password, username, token;
    token = security.UsernameToken;
    username = token.Username;
    password = token.Password;
    //perform database lookup
   databaseLookup(username, function(err, doc){
       if( err ){
           callback(false);
           return;
       }
       callback(doc);
   });
}

```


jsdevel:
Closing as it's been over a year since there was any activity here.

PR #433, opened 2014-08-12
PasswordText bug

nodejsbspa:
Tag Nonce EncodingType must be appear only where _passwordType is differente from 'PasswordText'


herom:
I'm closing this PR as there was no activity for nearly a year now...

Issue #460, opened 2014-09-16
Cannot read property 'attributes' of null


justerik:
The object that I want to send to the server contains some values that are null, something like this:
{value:{val1:"val1", val2: null}} 

Then I get the error "Cannot read property 'attributes' of null".

Is it not possible to send null? or is this unwanted behavior?

PR #468, opened 2014-10-04
Added extra null check

justerik:
Null check provides a fix for the bug in issue #460


herom:

PR #483, opened 2014-10-24
Adding XMLSchema to the xmlns for the soapenvelope

javm:
Hi I was using node-soap to consume some web service using the namespace 'xsd'
but executing queries the server was responding that it doesn't recognize that namespace,
so I just needed to add the following line in the soapenvelope in lib/client.js

I don't know it is possible include it dynamically, anyway it is just a hardcoded line:
 "xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\" " +

And here is the wsdl:
https://www.integracionesqapx.com.mx/wsuniversaldummy/pxuniversal.asmx?WSDL

Thanks


herom:

PR #486, opened 2014-11-01
added enhancement for new wsdl option useJsonNamespace

eric-clark:
by setting the option of useJsonNamespace: true for options passed in to
createClient the wsdl function objectToXml will not add a namespace to
the xml elements.  this way you can hard code the namespace into your
json.  if you don't set useJsonNamespace, it will default to false and
run as normal.


herom:

Issue #68, opened 2012-06-25
TypeError: Cannot set property 'targetNSAlias' of undefined


andrewkro:
Hey Guys,

After finally getting expat compiled and installed on windows, i've tried loading a wsdl into node-soap to create a web service server.

Ive loaded into SoapUI tool to test wsdl/xsd compliance and sent test request/response ok.

when i start a simple node http server, i get the following:

C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:346
        this.element.targetNSAlias = ns;
                                   ^
TypeError: Cannot set property 'targetNSAlias' of undefined
    at [object Object].postProcess (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:346:36)
    at [object Object].postProcess (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:378:17)
    at [object Object].postProcess (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:395:15)
    at [object Object].postProcess (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:408:14)
    at [object Object].postProcess (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:440:21)
    at C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:594:36
    at C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:641:13
    at C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:641:13
    at [object Object]._processNextInclude (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:635:26)
    at [object Object].callback (C:\Documents and Settings\Administrator\My Documents\Web Service Current\node_modules\soap\lib\wsdl.js:640:14)

The WSDL being loaded is here: http://pastebin.com/iECgpUJs
The first import schema which is loaded (aidx/IATA_AIDX_FlightLegNotifRQ.xsd) is here: http://pastebin.com/5V0r3eYM

Any help would be greatly appreciated!

Andrew

andrewbranch:
[...]

PR #526, opened 2014-12-14
fix for issue #68

Himanshu4:
Its a fix for issue: https://github.com/vpulim/node-soap/issues/68
if the element is not defined then it throws error on line lib/wsdl.js:469.


patryk0605:
Can you please merge this as I am running into issues without this line.

herom:
Thanks @Himanshu4 :+1:

PR #529, opened 2014-12-30
Enhance executeMethod by passing SOAP headers as an argument

julb:
For Node SOAP server, it is required for some services to get the SOAP headers from the request.
The server.js pass the headers as the 3rd argument of the method (after args and callback).


julb:
The commit f2503abffc28e8f5f1a820e3f86166085a1e5a18 includes the change that I proposed.
I close the P/R.

herom:
thanks for keeping track of the changes :+1:

PR #538, opened 2015-01-14
Return the soapHeader

nguyenchr:
@srushti

We need to return the Response Soap Header (not http response headers)
So that we can get information like wsUsernameToken

thoughts?


nguyenchr:
whoops meant to PR this to my fork

PR #539, opened 2015-01-14
Return the response soap header

nguyenchr:
I currently have the usecase for needing the Soap `Header` from a response
It appears the current implementation discards everything but the `Body`

This PR adds the response `Header` to the end of the callback

I think it would have been neater to return it in the `result` but that would be a breaking change


leonidez:
:+1: 

Using this fork now in my app. Upgrade was pretty simple since I am using `RSVP.denodeify`. Thanks.

nguyenchr:
hi @herom 
I just read that any PR that is open for over 2 weeks will be automatically closed.
This PR is approaching 2 weeks, so wondering if you had any thoughts?

herom:
@nguyenchr I'm really sorry but I'm actually very short on time and I completely lost track of the open (and passing) PRs. Actually there should be more collaborators but for the last months it seems that I'm own my own :smile: 

Sorry again and thanks a lot for your contribution - we appreciate it a lot :+1:

nguyenchr:
@herom no problem at all, thanks for following it up

Issue #489, opened 2014-11-03
Soap header from incoming request


Zassi:
I have created web service as shown in node-soap documentation, and web service is working OK. My question is, how I can fetch SOAP headers from incoming request? WSDL I am using is valid.

PR #551, opened 2015-02-01
Issue #489: Soap header from incoming request

mdhooge:
I think this commit should correct the Issue #489.

I've added a 3rd argument to the function called by the server to handle an incoming request.
The test I added firstly failed and now passes.

I did a "rebase" as explained in the Contributing file.


herom:
Thanks a lot for your contribution - we appreciate it a lot :+1:

PR #559, opened 2015-02-08
Client instances emit 'soapError' event when error is detected

mdhooge:
Now that the client can send events, it was really easy to add that one. I spent more time writing the tests!

Note: First, I chose the simpler 'error' event. But that one has a special meaning in node and that broke backward compatibility.


herom:
Awesome, thanks a lot @mdhooge :+1:

PR #562, opened 2015-02-10
fix for issue #68. Fix the case where requests are in soap format and… …

Himanshu4:


Himanshu4:
We had a soap api in which response is in json embedded in soap (xml). This will fix that issue.

herom:
Thanks @Himanshu4 - I assume we're good to close #526 now, are we?

PR #569, opened 2015-02-12
Fix soap XML namespace - change from soap to SOAP-ENV

marcus:
I'll preface this by saying I don't know much about SOAP but this is what I had to do to get this library working for me. I was taking the headers generated by node-soap and pasting them in an XML validator, they weren't valid. I changed the namespace from soap to SOAP-ENV based on an example. That validated and fixed my issues.

I've updated the tests, however 3 fail despite the actual and expected looking identical to me. If you're interested in this PR, I can put more effort into fixing those tests.


robertjneal:
Hi @marcus,

 Actually soap and SOAP-ENV are arbitrary namespace prefixes. As long as they point to the same namespace, i.e. http://schemas.xmlsoap.org/soap/envelope/, the prefix in the request and the WSDL do not need to match. One can use 'soap' and another 'SOAP-ENV'. Here's a useful resource: http://docs.oracle.com/cd/E19717-01/819-7757/aeqfk/index.html

 Without more details it's difficult to say what the cause of the issue you ran into might have been, but the library should work with anybody's SOAP requests without the changes in this PR. If you have questions feel free to post an issue and tag me and I'll see if I can help.

herom:
As @robertjneal explained, it shouldn't matter if the namespace is called `soap`, `SOAP-ENV` or whatever. Please open an issue describing your problem first as this is a rather HUGE PR and I don't think that this change will make it into `node-soap` as long as we don't have more information/issues/bug reports on this.

So, please, raise an issue (if you haven't already) for your particular problem and see if it could also be fixed without this huge renaming PR :+1:

PR #571, opened 2015-02-13
Configure for v0.12 and io.js compatibility

### LINKED ISSUE #550: Cannot assign to read only property 'namespace' of xsd:string
Hi there,

Here is my code, which works locally with node 0.11.14:

~~~javascript
return new Promise(function (resolve, reject) {
			//resolve(false);
			if (!iban) {
				reject(new Error('IBAN must be provided'));
			}

			soap.createClient(config.url, function (err, client) {
				var args = {
					iban     : iban,
					user     : config.user,
					password : config.password
				};

				/*jshint camelcase:false*/
				client.validate_iban(args, function (err, result) {
					var blz = false;
					if (_.hasPath(result, 'return.bank_code.$value')) {
						blz = result.return.bank_code.$value;
					}
					resolve(blz);
				});
				/*jshint camelcase:true*/
			});
		});
~~~

This is my test:
~~~javascript
    it('checks correct iban', function (done) {
        ibanrechner.getBlzByIban('PL2510301944****101620901000').then(function(result){
            console(result);
            expect(1).to.equal(1);
            done();
        }).catch(function(err){
            console.log(err);
            expect(0).to.equal(1);
            done();
        });
    });
~~~

And this is what I got when this test performs on my build server:
image

Could someone tell me what is wrong here?
I've tried another variant with the ```chai-as-promised``` lib, but the soap is not working on my build server.

What could be a reason of the error?

Regards,

andrewbranch:
Fixes #550. Sorry about the end-of-line spacing changes, my editor seemed to do that automatically for some reason.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,7 @@ language: node_js
 node_js:
   - 0.8
   - 0.10
+  - 0.12
+  - iojs
 before_install:
   - npm install -g npm@~1.4.6
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -565,8 +565,12 @@ MessageElement.prototype.postProcess = function(definitions) {
       } else {
         this.parts[part.$name] = part.$type;
       }
-      this.parts[part.$name].namespace = nsName.namespace;
-      this.parts[part.$name].xmlns = ns;
+      
+      if (typeof this.parts[part.$name] === 'object') {
+        this.parts[part.$name].namespace = nsName.namespace;
+        this.parts[part.$name].xmlns = ns;
+      }
+      
       this.children.splice(i--, 1);
     }
   }
@@ -869,8 +873,11 @@ ElementElement.prototype.description = function(definitions, xmlns) {
         else {
           elem = element[name] = typeElement.description(definitions, xmlns);
         }
-        elem.targetNSAlias = type.namespace;
-        elem.targetNamespace = ns;
+        
+        if (typeof elem === 'object') {
+          elem.targetNSAlias = type.namespace;
+          elem.targetNamespace = ns;
+        }
 
         definitions.descriptions.types[typeName] = elem;
       }
--- a/test/server-test.js
+++ b/test/server-test.js
@@ -56,7 +56,7 @@ describe('SOAP Server', function() {
 
       //windows return 0.0.0.0 as address and that is not 
       //valid to use in a request
-      if (test.server.address().address === '0.0.0.0') {
+      if (test.server.address().address === '0.0.0.0' || test.server.address().address === '::') {
         test.baseUrl =
           'http://127.0.0.1:' + test.server.address().port;
       }
--- a/test/ssl-test.js
+++ b/test/ssl-test.js
@@ -43,6 +43,11 @@ describe('SOAP Client(SSL)', function() {
       test.soapServer = soap.listen(test.server, '/stockquote', test.service, test.wsdl);
       test.baseUrl =
         'https://' + test.server.address().address + ':' + test.server.address().port;
+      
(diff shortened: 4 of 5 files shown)


mdhooge:
It is no surprise, but this also solves the problem with node 0.12.0.
However, I wonder if this wouldn't be better to explicitly check for "object" instead of an inequality with "string". I could try myself, but right now, I'm back to v0.10.36 ;-)

herom:
Thanks a lot for your contribution @andrewbranch - we appreciate it a lot! This really should make it into a new release as soon as possible :+1: 

Before we can merge this, I'll have to ask you to revert all unnecessary changes (like the `end-of-line` changes, etc.) so that we can keep our Git History clean and don't clutter it with changes which shouldn't be in history (the history should show only real code changes, nothing more).

Also, I guess it would be of further value, if you could add a line in our [.travis.yml](https://github.com/vpulim/node-soap/blob/master/.travis.yml) file to include `node v0.12.0` for testing along with your commit. If you do so, please amend and force push, or squash and force push your changes as we request **1 commit per PR** as stated in our [Guidelines for Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

andrewbranch:
@mdhooge — I agree; changed to `object`.
@herom — I’m not sure how to overwrite commits I already pushed to GitHub. You can see that what I just tried kind of made things messier—my apologies. Would it be easiest to delete this PR and start a fresh one? Sorry for the inconvenience.

andrewbranch:
Fixed the commit history, but it looks like there are other issues arising with newer node versions. Working on adding fixes to the pull.

andrewbranch:
Most of the failing tests from the last build were caused by `server.address().address()` returning `::`. I have no idea what that’s about, but I checked for it and replaced with `127.0.0.1`. However, there are still two failing tests that I don’t know how to fix. Would appreciate help from someone more familiar with the project.

robertjneal:
It looks like you can fix the error by copying the conditional you used in server-test.js to the beforeEach of ssl-test.js, namely:
if (test.server.address().address === '0.0.0.0' || test.server.address().address === '::') {

I didn't checkout your branch to confirm, but that looks to be the issue.

andrewbranch:
Everything looks good now! Thanks to @mdhooge, @herom, and @robertjneal for the input/guidance!

.travis.yml
```diff
@@ -2,5 +2,7 @@ language: node_js
 node_js:
   - 0.8
   - 0.10
+  - 0.12
+  - iojs
 before_install:
   - npm install -g npm@~1.4.6
```
herom: why was this removed?

andrewbranch: Only an addition is shown here—the dashes are part of the YAML syntax itself. :)

simonjosefsson: On a tangential note, is there a specific reason to stay with [npm@1.4.6](https://github.com/npm/npm/releases/tag/v1.4.6)?
It's almost a year old and [npm's](https://www.npmjs.com/package/npm) current requirement of node >= 0.8 seems like a non-issue no?

herom: @andrewbranch oh, sorry - I was totally shure that this was deleted :smile_cat: 

@simonjosefsson we needed `npm v1.4.6` in order to let the tests pass for `node v0.8.0` too while pulling in some new libraries. This could probably make it into another PR where we should discuss about support for the (very, very) old `node v0.8.0` in general.

herom:
Thanks a ton @andrewbranch :dancers:

andrewbranch:
@herom My pleasure—one step closer to using io.js on Heroku!

Issue #386, opened 2014-06-20
xmlToObject can interpret namespace attributes as methods


asentance:
I have a soap message that uses the inline style of namespacing:
```xml
<SOAP-ENV:Envelope 
		xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" 
		xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" 
		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
		xmlns:wmls="http://www.witsml.org/message/120">
		<SOAP-ENV:Body SOAP-ENV:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
			<wmls:WMLS_GetVersion/>
		</SOAP-ENV:Body>
	</SOAP-ENV:Envelope>
```
When server.js tries to get the method name, it uses:
methodName = Object.keys(body)[0];
But, the first key in body is 'attributes' because of the extra encodingStyle attribute in the Body element. As I don't have a method called attributes, it throws an exception.

I have a workaround that filters out methodNames called 'attributes', but that may be treating the symptom rather than the cause (and I'm rather new to this).

robertjneal:
I had the same issue. When attributes are added to the body tag, the initial index of Object.keys(body) is sometimes "attributes" rather than the expected messageElemName. I have a fix for it and plan on opening a PR today.

herom:
As @jsdevel  said, it would be fantastic if you code provide a PR to fix this issue @robertjneal - looking forward to it :+1: 

Please see our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before you start :smile_cat:

robertjneal:
Yeah, it's not going to make it today. Writing a test is more involved than I thought it would be for having just added a simple conditional. But this fix is critical for encrypted WSS requests. I'll try to get it in a PR soon.

herom:
Closing this issue as the provided fix by @robertjneal got merged in today. Feel free to reopen as you see things still going wrong/going wrong again.

PR #574, opened 2015-02-14
Issue #386 This fix adds support for attributes in body element

robertjneal:
There are specific use cases, for instance with wss security, where attributes are added to the soap body element. This fix adds support for the server to parse them and for the client to add them.


herom:
Thanks a lot @robertjneal :+1:

PR #578, opened 2015-02-23
Add support for sending NULL values

DeviaVir:
Currently, it's not possible to send data that looks like this:
```
{
  'object': {
    'string': null
  }
}
```
As you will get an error `undefined has no attributes`. 

This PR sets the correct way of communicating a `NULL` value.


herom:
Hey @DeviaVir - the same as in your previous PR, we really appreciate every single contribution but have to ask people to have a look at our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) and add tests to their code changes.

Thanks a lot! :smiley:

DeviaVir:
@herom Actually, regarding this one I checked other PR's who tried the similar but not following the standards which made me doubt my changes are in the right spot. Would you be able to throw a quick glance and tell me if I should move it before I add the tests? Thanks again!

lib/wsdl.js
```diff
@@ -1558,7 +1558,11 @@ WSDL.prototype.objectToXML = function(obj, name, namespace, xmlns, first, xmlnsA
       }
 
       if (!Array.isArray(child)) {
-        parts.push(['<', ns, name, attr, xmlnsAttrib, '>'].join(''));
+        var insNull = '';
+        if(typeof child === 'object' && child === null) {
```
herom: why do we need an extra check for `typeof child === 'object'`? wouldn't it be sufficient to check for `null` or `undefined` only. or, better to say, wouldn't it be safer (as we could also detect an empty String) to check for the "falsyness" of it by doing `if (!child) {...`?

herom:
I think that, if this approach works in the first place, and all tests are green, there isn't a question of "following the standards" as long as it's not completely out of place :smile: :+1:

lib/wsdl.js
DeviaVir: @herom Let me explain my reasoning for this extra check. SOAP API's are like grammar nazi's, I'm making absolutely sure here that this is in fact a NULL value, since you ONLY want the NULL values to have this kind of behavior. Please note `typeof null === 'object'`

herom: I see, so we need this extra check to guarantee that this is really just `null`? Please don't get me wrong, I'm not criticising your code, I'm just curious about the code others write and I'm wondering why we need to check for `=== 'object'` too if we only want to catch `=== 'null'` for further processing? As far as I know there is no value which could give a truthy check on `typeof val === 'null'` and not for `typeof val === 'object'` at the same time?

DeviaVir: Reconsidering, let me test if it would produce any false results if we check only `child === null`

DeviaVir: You're correct, I'm sure I added it for a reason, let me verify it in my tests! And if I can't figure out why then I'll get rid of it :) Thanks for your feedback, appreciate it!

herom: Thanks a lot for your understanding - I hope you don't get me wrong - I'm not nagging, I'm just up for learning as much as I can from others :smiley:

DeviaVir:
@herom Thanks again for your time and feedback, I have added tests for this PR now as well. Please let me know if you think I'm missing anything!

herom:
Thanks a lot :dancers:

DeviaVir:
Pfiati!

herom:
Pfiati gott :smile:

PR #579, opened 2015-02-23
Respond with security timestamp if request had one

roiemir:
When a request is sent with WSSecurity with a security header with a timestamp, it might be expected in the client to have a timestamp in the response.


herom:
Thanks a lot for your contribution @roiemir and sorry for the delay - I'm really short on time in 2015 :disappointed:

Issue #580, opened 2015-02-26
Returns an error when json is embedded in soap xml


Himanshu4:
Hi we have a web service which returns json embedded in soap body.

Sample response:
<SOAP-ENV:Envelope>
<SOAP-ENV:Body>
<GetStationsResult>
{[{"Source_ID":"9088","Station_Name":"Serndamaram"},{"Source_ID":"9089","Station_Name":"Virasigamani"},{"Source_ID":"9090","Station_Name":"Dodamarg(Goa)"},{"Source_ID":"9093","Station_Name":"Chimakurthy"}]}</GetStationsResult></SOAP-ENV:Body></SOAP-ENV:Envelope>

But, it fails at line 225. in lib/client.js . obj = self.wsdl.xmlToObject(body);

Ideally, it should return the json in response/body.

Himanshu4:
Closing the issue as it has been merged.

PR #581, opened 2015-02-26
If response is json embedded in soap, then error should not be thrown. Fix issue #580

Himanshu4:
Changes included:
1. Changed console.log to debug when soap response is json to avoid unnecessary logs.
2. Fix for issue #580 . When json is embedded in soap body, then currently it throws error. We won't return the error and send the response in body param. That way, it would be a cleaner approach for the client to distinguish it from other errors like "http error"/"soap error".

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,6 +15,7 @@ var http = require('./http'),
   assert = require('assert'),
   events = require('events'),
   util = require('util'),
+  debug = require('debug')('node-soap'),
   url = require('url');
 
 var Client = function(wsdl, endpoint, options) {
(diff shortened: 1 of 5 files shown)


herom:
Thanks a lot for your contribution @Himanshu4 - and sorry for the long delay :disappointed:. I'm getting the feeling of being the only contributor who cares for this project at the moment :smiley:

Personally, I'm totally happy with the addition of the `debug` library and I hope we make use of it more often so that this library can "speak about it's work" to the developer which would help a lot! So, whoever reads this and want to get a hand on pulling in `debug`  wherever necessary/helpful, this would make a great PR :+1: 

In order to get your PR merged in, I also have to ask you for accompanying tests, so we can make shure that your code changes/additions don't get overwritten accidently. Please have a look at our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) in advance :smiley:

Himanshu4:
@herom : Apologies for so long delay. I have added a test wsdl in which response element is not defined as it is a json and also added a test case for the same. Combined both the previous commits into a single one. Please check and let me know if it can be merged now.

Thanks.

herom:
Awesome, thanks a lot :+1:

Himanshu4:
@herom : In which version, would this change be start reflecting in the npm module.

herom:
@Himanshu4 I'm planning on releasing a new version of `node-soap` hopefully by Wednesday/Thursday this week :+1:

PR #586, opened 2015-03-04
ClientSSLSecurity should be able to take a Buffer as `key` and `cert` parameter

Mik13:
Additionally the certificates are checked whether they are correct or not (starting with `-----BEGIN`).

Added test, and corrected another one, which is passing invalid parameters.

Regards,
Mik13


herom:
Thanks a lot @Mik13 :+1:

Mik13:
You're welcome!

Issue #585, opened 2015-03-04
Object.keys called on non-object - error


ramotswa:
Hi,
I know this will probably have to do with my WSDL, because just replacing mine with one from the examples works, but I wonder if you might be able to point me in the direction of what might be wrong when I get the error;

/node_modules/soap/lib/wsdl.js:90
  Object.keys(obj).forEach(function(key) {
         ^
TypeError: Object.keys called on non-object
    at Function.keys (native)
    at extend (/node_modules/soap/lib/wsdl.js:90:10)
    at ExtensionElement.description (/node_modules/soap/lib/wsdl.js:798:9)
    at SimpleContentElement.description (/node_modules/soap/lib/wsdl.js:838:20)
    at ComplexTypeElement.description (/node_modules/soap/lib/wsdl.js:818:20)
    at ElementElement.description (/node_modules/soap/lib/wsdl.js:874:46)
    at AllElement.description.SequenceElement.description (/node_modules/soap/lib/wsdl.js:918:29)
    at ComplexTypeElement.description (/node_modules/soap/lib/wsdl.js:818:20)
    at ElementElement.description (/node_modules/soap/lib/wsdl.js:903:31)
    at MessageElement.postProcess (/node_modules/soap/lib/wsdl.js:543:33)

All I am doing is;
var soap = require('soap');
soap.createClient('http://localhost:2022/tracer.wsdl', function(err, client) {
  if (err) throw err;
  console.log(client.describe());
});

Thanks for any help,
Ramo

herom:
Thanks for raising this issue @ramotswa - it seems that either the parsed base or the whole element is not an `object` as expected by the `extend()` function (which raises the error).

A quick fix would be to check the given `base` and `obj` params for their type and only proceed to copy values if both are of type `object` and `!== null` in the `extend()` function before further proceeding is done.

Could you open a PR for this so we can merge it? Would be awesome :+1:

herom:
Closing as the provided fix at #589 was merged :+1:

PR #587, opened 2015-03-04
Updated extend() for null and typeof checks

ramotswa:
The extend function was throwing an error (https://github.com/vpulim/node-soap/issues/585). This null and typeof check will handle elements that are not objects.


ramotswa:
Will add tests... just read the contrib file...

herom:
Thanks @ramotswa :+1:

PR #588, opened 2015-03-04
ClientSSLSecurity now accepts a `ca`-certificate.

Mik13:
This is necessary if the server uses a self-signed certificate, which we want to trust, by setting it as the CA-certificate of our client.

+Test


herom:
Thanks a ton @Mik13 :+1:

PR #590, opened 2015-03-06
Add support for One-Way Operations

wimve:
I wanted to use this module for a webservice with some one-way soap operations (http://www.ws-i.org/profiles/basicprofile-1.1.html#One-Way_Operations) but it did not seem to support this. 
So I ended up adding some code and tests. It would be nice if this would be included in the project.

The commit passed the Travis build: https://travis-ci.org/wimve/node-soap/builds/53335821


herom:
Thanks @wimve - seems like a really valuable addion to `node-soap` :+1:

PR #595, opened 2015-03-11
Server 'request' event

mdhooge:
The server emits a "request" event for every received message.

(I needed it to easily handle the case where a client doesn't send a mandatory header…)


herom:
Great! Could you also please provide a proper description of this new event in our README? (please open a new PR for this).

PR #607, opened 2015-03-19
sub namespace should be correct regardless of order of enumeration i.e. ...

jiangyang:
...should not be overriden by other prop's namespace


test/request-response-samples/update__complex_extension_with_array_attributes/request.json
```diff
@@ -8,14 +8,14 @@
         "xmlns": "urn:relationships_2013_2.lists.webservices.netsuite.com"
       }
     },
+    "firstName": "Allister",
```
herom: Why did you prefer modifying an existing test in favour of creating a new one? Was the "old" test wrong in any sense?
I'm just asking, because "normally" rewriting tests in order to get a code running is very bad practice - unless the test was obviously wrong or not testing things at all. :smiley:

jiangyang: In this case, the old test was not wrong but rather covered a case too specific. If I'd add a new test it would be almost identical to this existing one only with the order of the properties changed. So I thought...might as well...we could make the existing test cover this aspect, too. If you'd prefer that, I can add a new one for sure.

Well, it is not exactly "rewriting a test to get a code running". The code isn't totally "correct" in the sense it "leaks" a local variable value supposedly only meant for the current iteration to affect other iterations and hence breaks the sub namespace.

herom: Ah, thanks - that makes sense :smiley:

PR #608, opened 2015-03-19
Bug fixes for import requests and imports that have a schema root

OuranosSkia:
Added a fix that allows the extra request headers and options to keep passing through for all imports/includes. This is necessary for servers which require basic authentication or other special parameters for every request.

Added a fix to properly handle import/include files that start with a schema element.

Bumped the version for you to allow quick publishing.


OuranosSkia:
Also fixed a large number of unit tests which were broken before, but went unnoticed due to schemas not importing correctly.

lib/wsdl.js
```diff
@@ -79,12 +79,12 @@ function trim(text) {
 }
 
 /**
- * What we want is to copy properties from one object to another one and avoid 
- * properties overriding. This way we ensure the 'inheritance' of 
+ * What we want is to copy properties from one object to another one and avoid
+ * properties overriding. This way we ensure the 'inheritance' of
```
herom: Could you please revert all of the changes which are not related to code but did change whitespaces and eol? we ask for this so we can keep our history clean and really just related to code changes where they belong.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.8.0",
+  "version": "0.8.1",
```
herom: Please revert your version change as we can't guarantee that our next release will be `0.8.1` or `0.9.0` or something the like. Please just change the `package.json` if it's necessary because of `dependencies`, `devDependencies` or `scripts`.

test/platform/lists.relationshipTypes.xsd
```diff
@@ -0,0 +1,203 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
herom: In general, could you please shorten your `xsd`, ... files so that the library isn't bloated with overloaded files just for testing purposes. I guess just a few child nodes per node will be fine - a whole list of countries or values which are basically all the same is not necessary for testing imho.

test/request-response-samples/addList__complex_extension_namespace_for_arrays/soap.platform.core.xsd
```diff
@@ -18,7 +18,7 @@
         <attribute name="isSuccess" type="xsd:boolean"/>
     </complexType>
     <element name="status" type="platformCore:Status"/>
-
```
herom: Also here, please revert changes which are just formatting/whitespace related.

test/request-response-samples/addList__complex_extension_namespace_for_arrays/soap.platform.messages.xsd
```diff
@@ -10,21 +10,21 @@
         </sequence>
     </complexType>
     <element name="writeResponse" type="platformMsgs:WriteResponse"/>
-
```
herom: Same here :smiley:

herom:
Hey @OuranosSkia,
thanks a lot for this PR - we appreciate it a lot and it seems that a lot of effort was made in order to make all these changes.

I left a few comments in the code which are related mostly to unneccesary changes (mostly formatting/whitespaces) which I kindly ask you to revert, as we want to keep our history clean.

Other than that, I'll also have to ask you for a few things (which are explained in our [Guidelines for Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request))

- Please, add accompanying tests for your code changes too, so that it's guaranteed we don't override any of your work accidentialy in the future.
- Please, once you've done the reverts/changes squash all your commits into a single one (links to an excellent guide on how to do this (by GitHub) are givein in our [Guidelines](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request))
- Please, please, please:  **do not close a pull request due to a request to rebase** 

Thanks a ton!

test/platform/lists.relationshipTypes.xsd
OuranosSkia: These are the actual files that were *supposed* to have been included before. I had to find these online to make sure they are the right ones being included.

OuranosSkia:
I addressed the comments and squashed commits. As far as tests, the existing tests should be sufficient to make sure my changes do not get reverted. In fact, making these changes brought about the issue that the existing tests were in fact broken. 

The long xsd files are xsd's that are referenced in the existing tests, but *did not* exist in the codebase at all. Due to the number of tests which used these xsd files and the fact that I did not know the explicit purpose of each test, I thought it prudent to leave the xsd's in their full form. I also thought this prudent since I put them in a common location instead of duplicating them for each test which used them.

Let me know if you have any more questions or comments!

herom:
Thanks a lot for your effort :+1: 

Regarding the "new" `xsd` files I guess we'll just integrate 'em and see how far we'll get :smiley:

OuranosSkia:
Thanks!

Issue #545, opened 2015-01-26
Logging Behaviours: Event Driven Logging vs. Client.lastRequest


evanshortiss:
How would you guys feel about a PR that allows the client object to emit events? I'd like to add it soon.

My primary reason for wanting this behaviour is for logging reasons. Using _Client.lastRequest_ wouldn't be my approach as you need to monitor it for changes and log accordingly. Using something like the following would be preferable:

```javascript
Client.on('request', function (xml) {
 // Do logging here
});
```

The same approach could be taken for a server.

herom:
I think this is a really good idea and it would be a valuable feature for this library. I'd love to see a PR for this :+1:

jsdevel:
+1

evanshortiss:
Awesome, while I'm working on my project I'll tackle this an submit a PR. Hopefully within 1-2 weeks.

evanshortiss:
Sending a PR now.

Two events are exposed on Client instances, "message" and "request".

"message" contains just the Soap Body.
"request" contains just the entire Soap request contents.

Having both is useful if you do not want headers in the XML you get from the event.

evanshortiss:
@herom let me know when you guys plan on merging the PR (#547) for this.

herom:
Sorry for the late response @evanshortiss and thanks a lot for your contribution, we appreciate it a lot :+1: 

In order to get this merged, I have to ask you to follow our [Guidelines for Submitting a  Pull  Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request), which - in your case - means, that we ask you to squash your commits into a single one. Your best bet would be with `git rebase -i HEAD~2` and a force push onto your remote repository, I guess. If you don't know what `git rebase` is, I'd recommend to head over to the (excellent) GitHub Help pages [here](https://help.github.com/articles/using-git-rebase/) and [here](https://help.github.com/articles/about-git-rebase/).

evanshortiss:
Thanks @herom I'm pretty sure I've successfully rebased and squashed into a single commit following those instructions

herom:
thanks a lot :dancers:

briandunnington:
This works great for us, but it would be even more awesomer if it also emitted a corresponding response event. Then all logging of requests and responses could be done in one spot and I wouldnt have to remember to do it in each callback.

PR #610, opened 2015-03-23
Client 'response' event

briandunnington:
Added a 'response' event to the client that is emitted when the SOAP response is received. Used in conjunction with the existing 'request' and 'message' events, it makes logging in one centralized place much easier. #545


herom:
Thanks a lot for this valueable addition :+1:

PR #619, opened 2015-04-03
Handle HTML answer from non-SOAP server

mdhooge:
I just hit the case of the attached test: a client with local WSDL and URL to an Apache server.

Because of the local WSDL, the typo in the URL isn't detected when fetching the wsdl from the remote server. And, on the first POST, it isn't a 400 error but valid HTML (saying something like "it works").
And then, the wdsl instance of the client returns the parsed html as json (instead of an XML body).


herom:
Hey @mdhooge - thanks a lot for your contribution & sorry for the extraordinary long delay :disappointed:

Issue #533, opened 2015-01-03
0.6.1 "Object #<Object> has no method 'indexOf'"


yicone:
In my case, there are an array arg in soap request, I received this error.

After review node-soap source code(0.6.1), I'm not sure there are a bug (in wsdl.js:1431):
```
     if(parentNamespace) 
```
I debug it for a long time, found the array element's "parentNamespace" is ''(empty), so that 'namespace' is continue to be a Object, that why I received the error of "Object #<Object> has no method 'indexOf'"(throw by line 1463).

So I guess there should be:
```
    if(parentNamespace || parentNamespace === '')
```

Relatedly， when I change it as above, I received another error: "Unmarshalling Error: Undeclared namespace prefix 'xs'".

Then I found the line 1469(still in wsdl.js):
```
    correctOuterNamespace = parentNamespace || ns; //using the parent namespace if given
```
Because parentNamespace may make sense when it equals '' , so I changed the line:
```
    correctOuterNamespace = (parentNamespace === '' ? '' : (parentNamespace || ns));
```

That works for me, but I'm not sure if I do a correct fix! So I hope that I can get some help.

herom:
@yicone thanks for reaching out and debugging the code down to the problem, this is much appreciated :+1: 

Given your description, this seems like a legit and valuable fix to the problem, so I'm happy if you can provide your code as a Pull Request :smiley:

Please, don't forget to add at least one test (request-response-sample preferred) and run `npm test` at your local machine before you reach in the PR. I'd also suggest you read our [Guideline for submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) as this would make the process much easier and faster :+1:

MageCoder:
We have the same issue, we cannot pass an array in the soap request args input. The 2 amendments described here fix this for us. It would be much appreciated to have this included in the next release version.
 (Fix described in #536 does not work for us.)

PR #630, opened 2015-04-14
Get current namespace when parent namespace is empty string

ibash:
This fixes issue: https://github.com/vpulim/node-soap/issues/533
Please let me know if the test should be elsewhere or should be changed in anyway. When the test fails an error is thrown.


herom:
Hello @ibash - thanks a lot for your contribution and sorry for the extraordinary long delay :disappointed: 

The test seems good to me, but would it be possible to shorten the `wsdl` file? In order to test the behavior of your fix I think it's not necessary to have such a big wsdl file and it would help us to keep the filesize low :+1:

ibash:
Hi @herom - I don't have access to the originally repo so I made the change and created a new pull request #661 . Closing this one.

PR #636, opened 2015-04-22
[Proposal] Generic XML string transformation before request

endorama:
Hello! First of all thanks for this library.

I had a problem using it due to a "strange" constrain in the SOAP API I have to use. Strange because I wasn't able to find information about a `ent` namespace in SOAP requests...
I'm a SOAP noob, so please forgive incorrectness or silly errors :)

For some reason the API instead of using `tns` ( I guess this is a xml namespace ) uses `ent`. This is in conflict with the generated request XML. I tried to dig in `node-soap` source code to find a way to change this but wasn't able to find one. My solution is a generic callback function to edit the XML before is sent ( simple RegExp search&replace ). It seems that `xmlns:tns` is taken from the `WSDL` definition ( so it is actually "expected" but the request returns an error without `ent` ).

I thought that maybe this could be useful, so here is this PR. Is more a proposal than a feature request.  
If there is already a way to perform this that I missed, please let me know.

Thank you.

PS: I wasn't able to create tests for this feature, but if you like the proposal and explain me how to create them, I will.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -40,6 +40,9 @@ The `options` argument allows you to customize the client with the following pro
 - endpoint: to override the SOAP service's host specified in the `.wsdl` file.
 - request: to override the [request](https://github.com/request/request) module.
 - httpClient: to provide your own http client that implements `request(rurl, data, callback, exheaders, exoptions)`.
+- xmlTransform: to perform custom editing of the request XML **before** is sent to the server. You can specify a function that implements`function (xml)`. This function will be executed just before the request XML is sent to the server and let you perform arbitrary modification to the XML string ( using string modification functions like RegExp or similar ).  
+`xml` argument is the XML request _string_.  
+The xmlTransform function specified **must** returns the transformed XML string.
 
 ### soap.listen(*server*, *path*, *services*, *wsdl*) - create a new SOAP server that listens on *path* and provides *services*.
 *wsdl* is an xml string that defines the service.
--- a/lib/client.js
+++ b/lib/client.js
@@ -208,6 +208,10 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
     "</soap:Body>" +
     "</soap:Envelope>";
 
+  if (typeof this.options.xmlTransform === 'function') {
+    xml = this.options.xmlTransform(xml);
+  }
+
   self.lastMessage = message;
   self.lastRequest = xml;
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -208,8 +208,19 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
     "</soap:Body>" +
     "</soap:Envelope>";
 
+  // if xmlTransform is defined as a function
   if (typeof this.options.xmlTransform === 'function') {
-    xml = this.options.xmlTransform(xml);
+    // run it and store return value
+    var tmpXml = this.options.xmlTransform(xml);
+    // check for undefined, which means that the function didn't have explicit return
+    if (typeof tmpXml === 'undefined') {
+      return callback(new Error('xmlTransform returned undefined, please return the transformed XML'));
+    } else if (typeof tmpXml !== 'string') {
+      return callback(new Error('xmlTransform returned a non-string value, please return the transformed XML'));
+    } else {
+      // if everything was ok overwrite xml
+      xml = tmpXml;
+    }
   }
 
   self.lastMessage = message;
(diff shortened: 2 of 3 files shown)


herom:
Hello @endorama - thanks a lot for your time & effort :+1: 

Is the purpose of your PR to ensure that the `ent` namespace is taken into account and served as well or to be ignored like the `tns` namespace? From your tests, it's not clear how the issue of the (missing?) `ent` namespace is related to it as it's only testing the presence of the `options` object on the client instance. I'll rather go for tests which ensure that your `xmlTransform` function on the `options` object is executed, etc.

Also, if we introduce this feature it would be necessary to mention it in our README. It would be awesome if you could provide another PR for adding some documentation on the `options` object and the newly introduced `xmlTransform` "hook".

In order to get your PR merged, please take a look at our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) and go through the steps provided over there. Namely, the following steps are necessary in order to merge your PR:

- Provide accompanying tests (it would be awesome if you could also provide tests for the `xmlTransform` "hook" and a "Request Response Sample~ which shows what you're trying to achieve (it would be best - I guess - if you show it with your `ent` namespace example)
- Rebase/squash your commits into a single one and rebase them on latest master so it could be merged automatically through GitHub (for more information see the Guidelines)

Thanks a lot in advance :smile_cat:

endorama:
Thanks for your feedback! As I wasn't sure about this PR, I didn't squashed commits to better clarify code edits, I'll do it when ready for merging :)

To clarify my proposal: I have this code for a SOAP request
```javascript
var url = 'https://example.com/endpoint?wsdl';
var args = {
  token: {
    tokenId: '...',
    locale: 'IT',
    authenticationUser: {
      username: '',
      password: ''
    }
  },
  .
  .
  .
};

soap.createClient(url, function(err, client) {
  if (err) { callback(err, null); }
  client.getChargingUnitsList(args, function(err, result) {
    if (err) { callback(err, null); }
    else { callback(null, result); }
  });
});
```

The XML sent for this request is:
```xml
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:tns="http://example.com/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soap:Body>
    <getChargingUnitsList>
      <token>
        <tokenId>...</tokenId>
        <locale>IT</locale>
        <authenticationUser>
          <username />
          <password />
        </authenticationUser>
      </token>
      .
      .
      .
    </getChargingUnitsList>
  </soap:Body>
</soap:Envelope>
```

The problem with this XML is that the endpoint is expecting this:
```xml
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ent="http://example.com/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soap:Body>
    <ent:getChargingUnitsList>
      <token>
        <tokenId>...</tokenId>
        <locale>IT</locale>
        <authenticationUser>
          <username />
          <password />
        </authenticationUser>
      </token>
      .
      .
      .
    </ent:getChargingUnitsList>
  </soap:Body>
</soap:Envelope>
```
Please notice the `ent` in the second line and before `getChargingUnitsList`. The xmlTranform that I defined to achieve this is:
```javascript
function _xmlTransformationCallback (xml) {
  xml = xml
          .replace(/xmlns:tns/, 'xmlns:ent')
          .replace(/getChargingUnitsList/g, 'ent:getChargingUnitsList');
  return xml;
}
```

The code is really simple, was just a quick hack to achieve my needs. I thought that could be a useful addition, and I know it still needs some work ( for example if you do not return something from the callback the xml var will became `undefined` which is obviously something you would like to check ).

I'll try to add some checks and the missing tests! :) Thank you

endorama:
I added Unit Tests for the xmlTransform feature, but unfortunately wasn't able to add a "Request Response Sample"; I didn't understand how to add my test case.

I have also updated the Readme with the relevant informations.

jsdevel:
@endorama you should be able to change the namespace now by prefixing like `ent:elementName`.  Also, [object-mapper](https://www.npmjs.com/package/object-mapper) is a really nice module for transforming soap requests.

jsdevel:
Closing for lack of activity.

endorama:
I can rebase and squash it if you are still interested in the PR :)

jsdevel:
Is this still needed?

endorama:
You're right, the same thing can be achieved using the new prefix overrides! Sorry

jsdevel:
no worries!  Thanks for your efforts!

Issue #606, opened 2015-03-18
Extended objects incorrectly parsed


kevinohara80:
I'm currently working with the Salesforce.com Metadata API. I'm calling their `readMetadata` method that return an array of Metadata objects. The Metadata complexType is a base for many other metadata types.

I expected that the js object returned would have all of the elements casted to their correct types but they are mostly being returned as strings. All other SOAP methods that do not return extended complexTypes are parsing just fine.

For example, in the output below the `enableActivities` is an element in the CustomObject metadata type and is supposed to be a boolean. 

``` js
[ { '$attributes': { 'xsi:type': 'CustomObject' },
    fullName: 'Test_Object__c',
    actionOverrides:
     [ [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object],
       [Object] ],
    compactLayoutAssignment: 'SYSTEM',
    deploymentStatus: 'Deployed',
    enableActivities: 'false',
    enableBulkApi: 'true',
    enableFeeds: 'false',
    enableHistory: 'false',
    enableReports: 'false',
    enableSharing: 'true',
    enableStreamingApi: 'true',
    fields: [ [Object], [Object], [Object], [Object] ],
    label: 'Test Object',
    listViews:
     { fullName: 'All',
       columns: [Object],
       filterScope: 'Everything',
       label: 'All' },
    nameField:
     { displayFormat: '{0}',
       label: 'Test Object Name',
       type: 'AutoNumber' },
    pluralLabel: 'Test Objects',
    searchLayouts: {},
    sharingModel: 'ReadWrite' } ]
```

Here are the Metadata and CustomObject complex types for reference from the WSDL:

``` xml
<xsd:complexType name="Metadata">
  <xsd:sequence>
    <xsd:element name="fullName" minOccurs="0" type="xsd:string"/>
  </xsd:sequence>
</xsd:complexType>
```

``` xml
<xsd:complexType name="CustomObject">
  <xsd:complexContent>
    <xsd:extension base="tns:Metadata">
      <xsd:sequence>
        <xsd:element name="actionOverrides" minOccurs="0" maxOccurs="unbounded" type="tns:ActionOverride"/>
        <xsd:element name="articleTypeChannelDisplay" minOccurs="0" type="tns:ArticleTypeChannelDisplay"/>
[...]

PR #638, opened 2015-04-25
pick up schema definition from xsi:type attribute

pentode:
This addresses https://github.com/vpulim/node-soap/issues/606

* add support for picking up schema from xsi:type attribute.
* add tests for feature.


pentode:
Closing due to multiple commits.  Will re-open after I clean this up.

Issue #613, opened 2015-03-26
Async Fault not working


ghughal:
I updated server-test.js to throw Fault exception from an async function but it doesn't seem to be working. Is this a bug or I should be handling this some other way?

``` javascript
    if (args.tickerSymbol === 'trigger error') {
        throw new Error('triggered server error');
    } else if (args.tickerSymbol === 'SOAP Fault') {
        console.log("before timeout");
        setTimeout(function() {
            console.log("throwing Fault exception after timeout");
            throw {
                Fault: {
                    Code: {
                        Value: "soap:Sender",
                        Subcode: { value: "rpc:BadArguments" }
                    },
                    Reason: { Text: "Processing Error" }
                }
            }
        }, 0);
    } else {
        return { price: 19.56 };
    }
```

Output:

```
before timeout
throwing Fault exception after timeout

/tmp/soap/test.js:30
                        throw {
                        ^
[object Object]
```

ghughal:
One more thing I noticed was when soap fault is thrown the HTTP status code is 200 but as per [SOAP spec](http://www.w3.org/TR/soap12-part2/#tabresstatereccodes) it should be either 500/400 and there is no way to specify that.

I've been playing with server.js and I think both of these issues can be fixed by extending callback arguments like this:

```javascript
        setTimeout(function() {
            cb(null, fault, 500);
        }, 0);
```

I'll try to submit a pull request for this.

brodin:
Think this is solved by https://github.com/vpulim/node-soap/pull/641

PR #641, opened 2015-04-29
Support async server errors

brodin:
This will add the possibility to send async errors back as a response from the server handler while staying backwards compatible with the current callback style. To get the test working I fixed a current bug where the tests in `request-response-samples-test.js` exposed its before/after/beforeEach hooks in a way which applied them to every other test in the suite.

### DIFF
--- a/test/request-response-samples-test.js
+++ b/test/request-response-samples-test.js
@@ -36,31 +36,6 @@ var requestContext = {
   }
 };
 
-var origRandom = Math.random;
-module.exports = {
-  before:function(done){
-    timekeeper.freeze(Date.parse('2014-10-12T01:02:03Z'));
-    Math.random = function() { return 1; };
-    server = http.createServer(requestContext.requestHandler);
-    server.listen(0, function(e){
-      if(e)return done(e);
-      port = server.address().port;
-      done();
-    });
-  },
-  beforeEach:function(){
-    requestContext.expectedRequest = null;
-    requestContext.responseToSend = null;
-    requestContext.doneHandler = null;
-  },
-  after:function(){
-    timekeeper.reset();
-    Math.random = origRandom;
-    server.close();
-  },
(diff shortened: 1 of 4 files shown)


brodin:
@herom any idea why these unrelated tests fails? Works locally with the different versions of node.

simonjosefsson:
The first two errors are mocha timeouts, probably caused by peak demand on the build system
(FYI, one can increase mocha's defaults two second timeout with --timeout <millisecond>)

The third one could be 0.12 specific, or a fluke... hard to say

brodin:
@herom changed this to one commit rather than two now.

test/request-response-samples-test.js
```diff
@@ -36,31 +36,6 @@ var requestContext = {
   }
 };
 
-var origRandom = Math.random;
-module.exports = {
```
herom: Why was this removed?

brodin: Tried to explain the purpose in my commit message:
> Changed how tests are generated to avoid global before/after hooks. By using the format with exporting before, beforeEach and after hooks aswell as the test suite one actually register the hooks globally - applying them to every test run by npm test. This wasn't really a problem until you try to do setTimeouts, which will never resolve since timekeeper.freeze is used (https://github.com/vesln/timekeeper#featuresproblems)"

Before it was attached to the global before hook, which meant that every test in the entire suite would get a fake random - if they wanted it or not. Now I limited the fake random to just be for the "request-response-sample" tests. The biggest problem was with timekeeper, blowing up any attempt to use setTimeouts in a test.

herom: Thanks a lot for the explanation - just wanted to be sure :+1:

gangstead:
As mentioned in #613 after this fix a http status code of 200 is still returned even when there is a soap fault.

PR #654, opened 2015-05-19
Ignore the `.idea` folder when pushing to npm

arthurschreiber:
The `node-soap@0.8.0` package on npm contains an `.idea` folder with a bunch of files that are not needed to use the `soap` module.


herom:
Thanks a lot - this was obviously my fault :smile:

PR #660, opened 2015-05-22
Add ability to customize http client/request lib

raymondfeng:
This PR refactors the http client into a class and add options to customize the `httpClient` or `request`. It makes the module extensible.  In my case, I need this change for two purposes:

- Be able to mock up the soap req/res over http
- Be able to register async hooks to the http calls to allow custom handling of http related security, dynamic headers and other mediations.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -11,7 +11,7 @@ function findKey(obj, val) {
       return n;
 }
 
-var http = require('./http'),
+var HttpClient = require('./http'),
   assert = require('assert'),
   events = require('events'),
   util = require('util'),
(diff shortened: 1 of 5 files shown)


herom:
Thanks a lot for time & effort @raymondfeng - we appreciate it a lot :+1: 

Could you please open another Pull Request and provide a guideline/some more context on this newly introduced (and in my opinion super valuable) feature in our README once this got merged? This would be awesome!

raymondfeng:
@herom Thank you for the quick response. I'll update the README.

Issue #470, opened 2014-10-07
Why is this xml node in my body request?


brunnakato:
### XML Request expected:
```xml
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:ws="http://ws.misp.com/">
   <soapenv:Header/>
   <soapenv:Body>
      <ws:getCountTicketsOnDraws>
         <srvId>5</srvId>
         <grp>1</grp>
      </ws:getCountTicketsOnDraws>
   </soapenv:Body>
</soapenv:Envelope>
```

### XML Request created with node-soap:
```xml
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"  xmlns:tns="http://ws.misp.com/">
  <soap:Header></soap:Header>
  <soap:Body>
    <getCountTicketsOnDraws xmlns:tns="http://ws.misp.com/" xmlns="http://ws.misp.com/">
      <tns:getCountTicketsOnDraws>
        <srvId>5</srvId>
        <grp>1</grp>
        <partnerId>0</partnerId>
      </tns:getCountTicketsOnDraws>
    </getCountTicketsOnDraws>
  </soap:Body>
</soap:Envelope>
```
I'm not undestanding how this node `<getCountTicketsOnDraws xmlns:tns="http://ws.misp.com/" xmlns="http://ws.misp.com/">` appeared there.

### Request I made:
```javascript
  var soap = require('soap');
  var url = "http://int.whitelabel.com.br/misp?wsdl";
  var args = {"tns:getCountTicketsOnDraws":{"srvId": "5", "grp": "1", "partnerId": "0"}};
  soap.createClient(url, function(err, client){
      console.log(client.describe());
      client.countTicketsOnDraws(args, function(err, result){
        console.log(err, result);
        console.log(client.lastRequest);
      });
  }, url);
```
### Error I got:
```
[Error: Invalid response: 500
[...]

PR #662, opened 2015-05-24
Issue 537: Take namespaces configured to be ignored into account

tobias-neubert:
Please take a thorough look into the code because I had to fix 3 tests after I changed the wsdl.js. Those three tests are at the same time the tests to proof my commit works as expected. I do not understand the purpose of the shouldIgnoreNamespace method that I deleted in this commit. The reason I deleted it is described below. 

Although one can provide an array of namespaces that shall be ignored to createClient(), they are not used in objectToXml(). This commit fixes that.

Moreover it removes the shouldIgnoreNamespace() method because it makes no sense to explicitly configure namespaces to be ignored and than let some weird logic calculate if a given namespace is going to be ignored.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1063,7 +1063,6 @@ var WSDL = function(definition, uri, options) {
 };
 
 WSDL.prototype.ignoredNamespaces = ['tns', 'targetNamespace', 'typedNamespace'];
-WSDL.prototype._ignoredSchemaNamespaces = ['tns', 'xs', 'xsd'];
 
 WSDL.prototype.valueKey = '$value';
 WSDL.prototype.xmlKey = '$xml';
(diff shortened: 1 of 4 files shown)


herom:
@tobias-neubert thanks a lot for your effort :+1: 

The `shouldIgnoreNamespace()` method was introduced in #506 as a Bugfix for the issue described in #470. 

As this introduced another array of ignored namespaces (which is only taken into consideration on the schema) - `WSDL.prototype._ignoredSchemaNamespaces` I think this property should also be removed if there is no need for the method itself anymore. As the tests are not failing when the method is removed (as in your PR) the method is either not necessary anymore, or the tests were wrong... hmm...

tobias-neubert:
Well, the tns namespace was exactly the one I had problems with too. As I said already, I do not really understand what the shouldIgnoreNamespace method tries to find out but, again, I believe that it makes no sense to have any kind of automation that overrides an explicit configuration. And as you stated: The tests are green.

Hence, I deleted the _ignoredNamespaces property.

herom:
Thanks a bunch :+1:

I guess, due to the reason of "green tests", we're save to delete/revert this previous PR and move firward savely. If problems because of this change come up, we should address them first with failing tests imho.

PR #661, opened 2015-05-23
Get current namespace when parent namespace is empty string

ibash:
This fixes issue: #533
Originally pull request #630

PR #651, opened 2015-05-14
Add last endpoint

rfink:
Addresses #650

PR #649, opened 2015-05-11
Need to obtain URL path from incoming soap request

fyatzeck:
As a solution to issue #643, I am proposing an update to server.js to invoke soap service function with
additional argument to provide URL path from incoming request.

I have tested soap requests with new "pathName" field in updated function callback and successfully received value as expected by my service function.

PR #664, opened 2015-05-26
Update readme with docs for client options

raymondfeng:
@herom This is a follow-up patch to https://github.com/vpulim/node-soap/pull/660

Can we get a release after this?


herom:
Thanks a lot :+1: 

For the release I want to wait until the end of this week so that we'll maybe able to merge #651, #649 and #661 which I think would make a great pack of features for a `0.10.0` release :smiley_cat:

PR #670, opened 2015-06-03
added test for vSphere Web Services API to fix #668

### LINKED ISSUE #668: soap 0.9.x seems to break node-vsphere-soap
Versions 0.9.x seem to break my reedog117/node-vsphere-soap#2 module.  All my tests seem to pass when reverting back to 0.8.0.

It seems that when I try to use the RetrieveServiceContent call on an ESXi or vCenter server, instead of getting the proper response I get something like:

``` xml
<soapenv:Envelope xmlns:soapenc="http://schemas.xmlsoap.org/soap/encoding/"
 xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/"
 xmlns:xsd="http://www.w3.org/2001/XMLSchema"
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
<soapenv:Body>
<soapenv:Fault><faultcode>ServerFaultCode</faultcode><faultstring>Unable to resolve WSDL method name RetrieveServiceContentRequestMsg for namespace name

while parsing SOAP body
at line 1, column 205

while parsing SOAP envelope
at line 1, column 0

while parsing HTTP request before method was determined
at line 1, column 0</faultstring><detail><InvalidRequestFault xmlns="urn:vim25" xsi:type="InvalidRequest"></InvalidRequestFault></detail></soapenv:Fault>
</soapenv:Body>
</soapenv:Envelope>
```

Doing a `Client.describe()` shows the correct method 'RetrieveServiceContent', so I'm not sure what's going on.

The original wsdl URL looks like this:

``` xml
<?xml version="1.0" encoding="UTF-8" ?>
<!--
   Copyright 2005-2015 VMware, Inc.  All rights reserved.
-->
<definitions targetNamespace="urn:vim25Service"
   xmlns="http://schemas.xmlsoap.org/wsdl/"
   xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/"
   xmlns:interface="urn:vim25"
>
   <import location="vim.wsdl" namespace="urn:vim25" />
   <service name="VimService">
      <port binding="interface:VimBinding" name="VimPort">
         <soap:address location="https://localhost/sdk/vimService" />
      </port>
   </service>
</definitions>
```

and the vim.wsdl it links to is pretty big so I'll link to another project that also has it: https://github.com/shoumikhin/ESXi-PoC-Client/blob/master/vim25/vim.wsdl

reedog117:
This adds a test that works in 0.8.0 but fails in 0.9.x, addressing issue #668.

Note that the vSphere Web Services WSDL is pretty complex and required the mocha timeout to be increased from the default 2000ms.


herom:
Would it be possible to shorten the WSDL so that we don't have to have the full file in our tests? I'm not happy with adding those full WSDL files for testing this particular issue only. If we'd do that, the size of the library would explode... 

I know that this would perhaps take some time, but otherwise I'm afraid I can't merge the PR as the file sizes are so big (when I compare this to other rejected PRs in the past which where done by the main contributor @jsdevel) :(

reedog117:
WSDL shortened, and I was able to switch the test timeout back to defaults.

herom:
Thanks a lot for your effort :+1:

jsdevel:
closing for lack of activity

herom:
I'd vote to reopen this one @jsdevel , as this PR was intended to introduce a failing test for this library, so that another community-member is able to take the challenge of providing a fix for it. I thouth it would be a good idea to have some "failing" request-response tests, so that we're aware of the bug/feature-request.

Maybe we can add the "needs pull-request" flag (which is kind of ironic, to flag a PR with)? I think it would make sense to keep this (and maybe we'll have some more of this kind) when I think of closing the "issues" section all together in the near future :smiley:

reedog117:
I concur -- I created this pull request, as something between 0.8.0 and 0.9.x caused my module [node-vsphere-soap](https://www.github.com/reedog117/node-vsphere-soap) to break.

That plus issue #668 is still open, and this pull request was created so that a fix could be made so that issue could be closed.

jsdevel:
@reedog117 do you plan on fixing the issue?

reedog117:
I'm not totally familiar with how node-soap parses the WSDL, so I'm not sure where to begin.  Any pointers as to where to look first so I'm not spinning my wheels as much?

jsdevel:
I'd just dig in and trace it down to where the problem is.

jsdevel:
closing for lack of activity

PR #674, opened 2015-06-15
Use different namespace styles for SOAP Fault 1.1 and 1.2

brodin:
Updated the fault handler in the server component to support both the fault style for SOAP 1.1 and SOAP 1.2. The main difference is that for soap 1.1 only the root element should have a namespace, but for 1.2 every element should have the same namespace.

## SOAP 1.1
```xml
<soap:Fault>
  <faultcode>soap:Client.BadArguments</faultcode>
</soap:Fault>
```
http://www.w3.org/TR/2000/NOTE-SOAP-20000508/#_Toc478383507

## SOAP 1.2
```xml
<soap:Fault>
  <soap:Code>
    <soap:Value>soap:Sender</soap:Value>
    <soap:Subcode>
       <soap:Value>pc:BadArguments</soap:Value>
    </soap:Subcode>
</soap:Fault>
```

http://www.w3.org/TR/soap12-part1/#soapfault

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -30,7 +30,7 @@ var url = require('url'),
 
 try {
   compress = require("compress");
-} catch (e) {
+} catch (error) {
 }
 
 var Server = function(server, path, services, wsdl, options) {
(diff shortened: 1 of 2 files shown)


herom:
Thanks a bunch :+1:

PR #680, opened 2015-07-04
Cleaning up and refactor code.

Neppord:
I have taken the time to extract common functions and Element constructors into a number of smaller files.

I did not collapse the commits into one, since i don't think there is a need since all of them could be there own pull-request and it would only hide documentation, on why things where changed. Please tell me if its still needs to be done.


herom:
Thanks a lot for your time and effort @Neppord - clean up and refactoring is much needed within this project (as far as I can tell). As I'm not the owner of the project nor a "lead maintainer" or something the like I'll ping @jsdevel and @vpulim so that they have the chance to sum up their thoughts on this rather huge PR and if we're going to make an exception from our contributing guidelines.

If we don't hear from them within a week I'll go further and merge your PR as is, as to me this seems very valuable :+1:

jsdevel:
Sorry @Neppord, but this is too dangerous a refactor.  It's also hard to follow given the number of changes.  Please issue smaller, more focused PRs in the future.

PR #686, opened 2015-07-20
Fix a typo in WSDL#findChildParameterObject

arthurschreiber:
I'm not quite sure what impact this typo had, but it's obviously wrong.

I stumbled over this when reviewing the code changes between 0.6.0 and 0.6.1.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1673,7 +1673,7 @@ WSDL.prototype.findChildParameterObject = function(parameterTypeObj, childName) 
       i = 0,
       child;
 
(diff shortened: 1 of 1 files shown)


herom:
Thanks a lot @arthurschreiber - this seems like it was really borked... it seems that we didn't really have a test for this scenario, as it's obviously wrong :fearful: 

Because, if you try it in your console, you can see that:
```
var x = [1, 2, 3];
var isBorked =Array.isArray(x && x.length); // returns `false` because `x && x.length === 3` (!)

console.log('Check for `isArray` on `x`: ' + isBorked; // returns false
```

PR #688, opened 2015-07-23
Fix for null in array

allancarvalho:
Hello,

I found a problem when the module receives null in array.

https://gist.github.com/allancarvalho/6a3e0cb0efecc883bf19

I need to generate the xml like soap.xml (gist).
In line 21 has a empty tag and the module remove this like soap_generated.xml (gist).

The part of my javascript object that generated the problem is:
var records = [['allan_tk@msn.com', null, 'I', '0'], ['fulano@test.com', 'FULANO', 'I', '0']];

In line 36 the module generated the correct tag when it receives null. 
The problem occurred when I send null in array.

This pull request resolve this.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1503,11 +1503,12 @@ WSDL.prototype.objectToXML = function(obj, name, namespace, xmlns, first, xmlnsA
   }
 
   if (Array.isArray(obj)) {
-    for (var i = 0, item; item = obj[i]; i++) {
+    for (var i = 0, item; i < obj.length; i++) {
+      item = obj[i];
       var arrayAttr = self.processAttributes(item),
           correctOuterNamespace = parentNamespace || ns; //using the parent namespace if given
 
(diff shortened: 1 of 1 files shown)


jsdevel:
@allancarvalho can you add a request response sample?

allancarvalho:
The gist that I posted has the generated XML with error.

https://gist.github.com/allancarvalho/30964474598b50d249b0#file-errorsoap-xml

The service that I am using receives an FieldNames and an Array of records (users).
But some users doesn't have some fields, like in the gist.
The first user in array of records doesn't have the field APELIDO and with the XML generated has a conflict because this think that the field APELIDO has a Value "I", but the value "I" is for field EMAIL_PERMISSION_STATUS_.

The service expects receive NULL in this case.

Thanks

jsdevel:
Please update the PR with tests.

jsdevel:
closing for lack of activity.

PR #690, opened 2015-07-28
added sample result

SunboX:


SunboX:
ups, this happens only if HTTP Code is 500

PR #694, opened 2015-08-11
Allow WSDL to be loaded from HTTPS sites

jimmyjones2:


jsdevel:
@jimmyjones2 Can you think of an easy way to test this?  I'm fine merging as is, but wanted to know.

jimmyjones2:
Did you just want an example? [Paypal SOAP API](https://developer.paypal.com/docs/classic/api/PayPalSOAPAPIArchitecture/) uses https for WSDL, eg. https://www.sandbox.paypal.com/wsdl/PayPalSvc.wsdl

jsdevel:
If that's a constant source that will not change, then yes, please add a test for it.  I don't want your hard work to go to waste by getting broken in future changes.

jsdevel:
Thanks for the PR @jimmyjones2 !  It would be really hard to maintain the tests for this change, so I merged it.  If you can add tests later, pleast submit a PR.

PR #703, opened 2015-08-25
Include correct namespace alias for operations without parameters

tdecaluwe:
Add the correct namespace alias for operations without parameters by simply removing the special case where input.parts is empty. If special logic is wanted for this case, it should be contained in objectToRpcXML in any case.


tdecaluwe:
Since `typeof(null)` will never be `'string'` this commit does not change the behaviour of the conditional.

jsdevel:
THis PR needs tests.

tdecaluwe:
I added a request sample to test operations without params.

jsdevel:
Thanks!

PR #714, opened 2015-09-10
Fix for wsdl retrieval with specified options.httpClient

kevmegforest:
When using soap.createClient with special options.httpClient, the specified client is not used when fetching the wsdl file. This fix will force the wsdl to use the specified httpClient.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1893,7 +1893,7 @@ function open_wsdl(uri, options, callback) {
   }
   else {
     debug('Reading url: %s', uri);
-    var httpClient = new HttpClient(options);
+    var httpClient = options.httpClient || new HttpClient(options);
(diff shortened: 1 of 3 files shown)


jsdevel:
please add a test.

kevmegforest:
I don't know how to make the test I wrote work for nodejs 0.8 or 0.10. 
This is really the use case I need, passing the http requests through the stream instead of absolutely passing by the net socket.

Do you have some idea.

Thank you

kevmegforest:
I've made all the tests. It works with all node version
I'm waiting for comments.

Thank you

kevmegforest:
@jsdevel , Is there something I would need to add or modify for you to merge this pull request?

jsdevel:
Looks good.  Thanks @kevmegforest !

PR #716, opened 2015-09-14
New Ignorebasenamespaces option

lauort:
Hi all,

I added a new option because in some cases you need to remove the base name of xml request nodes.

With the new option you will be able to set or not the basenamespaces.

Regards,

And thanks a lot to  @jsdevel with his help.


jsdevel:
Thanks @lauort !

Issue #720, opened 2015-09-22
Need to override the default namespace for "xmlns:soap".


chaitanyagpt21:
By default, the namespace "xmlns:soap=\"http://schemas.xmlsoap.org/soap/envelope/\" " is added to the SOAP envelope. But, I need to add a different namespace for the same in the SOAP request which is xmlns:soap="http://www.w3.org/2003/05/soap-envelope". I am passing the latter in client.wsdl.xmlnsInEnvelope but since the former one is present by default in request, I dont get any response from server. When I remove the former namespace and keep only the latter in request, I get the desired response.

Himanshu4:
+1

garvitlnmiit:
+1

jsdevel:
Please open a PR.

shashank-14612:
+1

Issue #719, opened 2015-09-22
Content-Type in header as 'application/soap+xml' instead of 'text/xml'


chaitanyagpt21:
I am doing API integration with a third party for my node application. Their APIs are written in .NET probably using SOAP 1.2 and they require content-type as 'application/soap+xml' instead of 'text/xml'. I also tested testcases for this module after changing the content-type. There are no such issues. Please look into it or if possible, we can pass parameter for the same.

Himanshu4:
+1

garvitlnmiit:
+1

jsdevel:
Please open a PR with tests.

PR #721, opened 2015-09-22
Override default namespace for xmlns:soap and set Content-Type as application/soap+xml (Issue #719, #720)

chaitanyagpt21:
I have compared for namespace 'xmlns:soap'. If it is already there in 'this.wsdl.xmlnsInEnvelope', then that will be used in place of default one. Also, since SOAP 1.2 applications require Content-Type as 'application/soap+xml', I have changed the content type. Please let me know if there are issues with that.


jsdevel:
You need tests and only 1 commit in your history.

herom:
ping @chaitanyagpt21 - it would be awesome if you you could add tests and squash your commits in order to get you PR merged :+1:

jsdevel:
closing for lack of activity

PR #731, opened 2015-09-28
Adding gitter badge in preparation of disabling issues.

jsdevel:
Here are the main reasons we're doing this:

* We get lots of issues from users that are of the 'Hey debug this for me' type.
* The amount of issues we get prevents us from adequately addressing PRs.
* If we solely focused our efforts on PRs, we'd likely get more PRs in and advance the library faster.
* Given the volume of issues we receive and the number of open issues we have open, users are less inclined to use  in the fear that is is overly buggy.
* We have so many open issues it's almost impossible to know which ones are still valid.
* Github issues aren't the best mechanism for community support.  Something like Google Groups or Gitter would best serve the interests of most issues we receive.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,4 +1,5 @@
-# Soap [![NPM version][npm-image]][npm-url] [![Downloads][downloads-image]][npm-url] [![Build Status][travis-image]][travis-url]
+# Soap [![NPM version][npm-image]][npm-url] [![Downloads][downloads-image]][npm-url] [![Build Status][travis-image]][travis-url] [![Gitter chat][gitter-image]][gitter-url]
+
 > A SOAP client and server for node.js.
 
 This module lets you connect to web services using SOAP.  It also provides a server that allows you to run your own SOAP services.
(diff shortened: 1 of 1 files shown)


herom:
I'm all :+1: for this one :smiley:

Name-less:
I understand your reasons, but It is still possible to see old issues to compare them with our one without borrowing you ?

zenparsing:
Lame

prusanoff:
Hey @jsdevel 

      I can understand why you disabling the issues section because of 'Hey debug this for me'. Your time is valuable commodity.  However for some users like my self XML, WSDL's are migraines and your ReadMe is not very descriptive. Not every method has a description of how to use it and some of them are vary confusing. 

Can you please update the read me especially for [ignored NameSpaces](https://www.npmjs.com/package/soap#handling-ignored-namespaces). I do not understand this at all! I am using a WSDL with two schema's and I have a tns and a ns1 which is being ignored here. I am not sure how to fix this problem since there is no documentation for it. thanks for your time. Peter

jsdevel:
@prusanoff, us maintainers do not have the time to fix every issue or add every improvement that is requested of us.  We do have time to review PRs.

PR #733, opened 2015-09-28
issue707: returning null instead of empty json object 

anvk:
for empty XML nodes

I'm not sure how to write tests for this pull request so I would appreciate any help or suggestions on how to do so.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1325,7 +1325,8 @@ WSDL.prototype.xmlToObject = function(xml) {
       topObject[name].push(obj);
     }
     else {
-      topObject[name] = obj;
+      topObject[name] = typeof obj !== 'object' || Object.keys(obj).length ?
+        obj : null;
     }
 
     if (cur.id) {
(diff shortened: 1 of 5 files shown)


jsdevel:
See `request-response-samples`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.9.3",
+  "version": "0.9.4",
```
jsdevel: we'll handle version bumping.  Please remove this.

anvk:
I removed version bump per request and changed some of the tests to accommodate that change. Although I have problems with tests of such format:

```
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
   <soap:Header />
   <soap:Body>
      <Response xmlns="http://www.example.com/v1" />
   </soap:Body>
</soap:Envelope>
```

What would be the result of this? I assume maybe `{}` would make more sense than `null` but again not entirely sure....

jsdevel:
see the other response samples under `test/`

jsdevel:
@anvk can you update the tests?

jsdevel:
Taking this up in https://github.com/vpulim/node-soap/pull/784.

PR #735, opened 2015-09-29
Timestamp Optional 

jayharper:
Added the option to generate a timestamp when creating a soap client.  Also added a new test in  request response sample to test this option.


herom:
Thanks a lot for your time and effort @jayharper :+1: 

In order to get your PR merged, please squash your commits like pointed out in our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

jayharper:
My apologies for not squashing to a single commit originally.

jsdevel:
Looks good @jayharper !  :+1: from me @herom

lib/security/WSSecurity.js
```diff
@@ -3,10 +3,11 @@
 var crypto = require('crypto')
   , passwordDigest = require('../utils').passwordDigest;
 
-function WSSecurity(username, password, passwordType) {
+function WSSecurity(username, password, options) {
```
herom: As the `passwordType` String param was not only replaced by an `options` Object parameter,  we should do something like this when setting `this._passwordType`:
```javascript
options = options || {};

this._passwordType = typeof options === 'string' ? options : options.passwordType;
this._hasTimeStamp = options.hasTimeStamp;
```

Also, I think we should update the README in order to let the people know, that an `options` Object is expected and what it could contain.

lib/security/WSSecurity.js
```diff
@@ -3,10 +3,11 @@
...
   this._username = username;
   this._password = password;
-  this._passwordType = passwordType || 'PasswordText';
+  this._passwordType = (options && options.passwordType) || 'PasswordText';
```
herom: If we "reinit" the `options` param at the very beginning, I think it could save us from checking for the `options` to be properly introduced furthermore.

```javascript
function WSSecurity(username, password, options) {
  options = options || {}; // <-- this
```

herom:
I just addressed some concerns I had regarding the newly introduced `options` Object. That's just my 2 cents but I think this would be not also a "nice to have" but also a "must have" if we "break" the constructor like this. :smile_cat:

lib/security/WSSecurity.js
jayharper: @herom take a look at my new updates.  My goal was not to break implementations utilizing the `passwordType` string parameter.  The `options` object is not required but rather optional.  The `passwordType` property default is `PasswordText` and the `hasTimeStamp` property will default to true.

lib/security/WSSecurity.js
```diff
@@ -3,10 +3,21 @@
...
+  options = options || {};
   this._username = username;
   this._password = password;
-  this._passwordType = passwordType || 'PasswordText';
+  //must account for backward compatibility for passwordType String param as well as object options defaults: passwordType = 'PasswordText', hasTimeStamp = true   
+  if (typeof options === 'string') {
+    this._passwordType = options;
+    this._passwordType = this._passwordType === 'PasswordDigest' ? 'PasswordDigest' : 'PasswordText';
+  } else {
+    this._passwordType = typeof options.passwordType !== 'undefined' ? options.passwordType : 'PasswordText';
```
herom: thanks a lot for your patience @jayharper :+1: 

I think it's completely viable to set `this._passwordType` and `this._hasTimeStamp` properties once without setting them and resetting them within the next line. So if you please -  as a last step - could change the "double set"-lines with "one-liners" like:

```javascript
if (typeof options === 'string') {
  this._passwordType = options ? options : 'PasswordText';
} else {
  this._passwordType = options.passwordType ? options.passwordType : 'PasswordText';
}

this._hasTimeStamp = !!options.hasTimeStamp; // implicit Boolean conversion!
```

This would a) shorten the code and b) make it more readable. Thanks a lot for your understanding :smiley:

jayharper: @herom I understand your desire to shorten the code and make it more readable, unfortunately your suggested code does not pass the three new tests I have implemented.  Could I ask you to check out my commit and run the tests yourself to help me find a solution?

herom:
Somehow I was not able to fork your repo and create a PR there, so here's the snippet for the `WSSecurity` constructor, I hope this helps :smiley:

```javascript
"use strict";

var crypto = require('crypto');
var passwordDigest = require('../utils').passwordDigest;
var validPasswordTypes = ['PasswordDigest', 'PasswordText']; // <-- this is new!

function WSSecurity(username, password, options) {
  options = options || {};
  this._username = username;
  this._password = password;

  //must account for backward compatibility for passwordType String param as well as object options defaults: passwordType = 'PasswordText', hasTimeStamp = true
  if (typeof options === 'string') {
    this._passwordType = options ? options : 'PasswordText';
  } else {
    this._passwordType = options.passwordType ? options.passwordType : 'PasswordText';
  }

  if (validPasswordTypes.indexOf(this._passwordType) === -1) {
    this._passwordType = 'PasswordText';
  }

  this._hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true;
}
```

jayharper:
@herom thanks for cleaning this up :thumbsup:.  I have incorporated your changes and we should now be good to go.

herom:
Thanks a ton :+1: :dancers:

jsdevel:
:dancers:

PR #745, opened 2015-10-13
Allow circular XSD files to be loaded

### LINKED ISSUE #743: XSD Circular/Cyclic Reference
Hi,

I'm dealing with WSDL with circular/cyclic reference, e.g. like below

    file.wsdl
        xs:import namespace="A" schemaLocation: A.xsd
    A.xsd
        xs:import namespace="B" schemaLocation: B.xsd
    B.xsd
        xs:import namespace="A" schemaLocation: A.xsd

I noticed the lib/wsdl.js will perform infinite loop when hitting circular reference.
Any suggestion or thought?

fabiancook:
Fixes #743

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1130,7 +1130,7 @@ WSDL.prototype._processNextInclude = function(includes, callback) {
     includePath = url.resolve(self.uri, include.location);
   }
 
-  open_wsdl(includePath, this.options, function(err, wsdl) {
+  open_wsdl_recursive(includePath, this.options, function(err, wsdl) {
     if (err) {
       return callback(err);
     }
@@ -1875,7 +1875,47 @@ WSDL.prototype._xmlnsMap = function() {
   return str;
 };
 
+var WSDL_CACHE = { };
+
+/* 
+  Have another function to load previous WSDLs as we
+  don't want this to be invoked externally (expect for tests)
+  This will attempt to fix circular dependencies with XSD files, 
+  Given
+
+    file.wsdl
+        xs:import namespace="A" schemaLocation: A.xsd
+    A.xsd
+        xs:import namespace="B" schemaLocation: B.xsd
+    B.xsd
+        xs:import namespace="A" schemaLocation: A.xsd
+
+  file.wsdl will start loading, import A, then A will import B, which will then import A
+  Because A has already started to load previously it will be returned right away and 
+  have an internal circular reference
+
+  B would then complete loading, then A, then file.wsdl
+
+  By the time file A starts processing its includes its definitions will be already loaded, 
+  this is the only thing that B will depend on when "opening" A
+*/
+function open_wsdl_recursive(uri, options, callback) {
+  var fromCache;
+
+  if (typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
+
+  if (fromCache = WSDL_CACHE[ uri ]) {
+    return callback.call(fromCache, null, fromCache);
+  }
+
+  return open_wsdl(uri, options, callback);
+}
+
 function open_wsdl(uri, options, callback) {
+
   if (typeof options === 'function') {
     callback = options;
     options = {};
@@ -1893,6 +1933,7 @@ function open_wsdl(uri, options, callback) {
       }
       else {
         wsdl = new WSDL(definition, uri, options);
+        WSDL_CACHE[ uri ] = wsdl;
         wsdl.onReady(callback);
       }
     });
@@ -1905,6 +1946,7 @@ function open_wsdl(uri, options, callback) {
         callback(err);
       } else if (response && response.statusCode === 200) {
         wsdl = new WSDL(definition, uri, options);
+        WSDL_CACHE[ uri ] = wsdl;
         wsdl.onReady(callback);
       } else {
         callback(new Error('Invalid WSDL URL: ' + uri + "\n\n\r Code: " + response.statusCode + "\n\n\r Response Body: " + response.body));
--- a/test/wsdl-parse-test.js
+++ b/test/wsdl-parse-test.js
@@ -14,4 +14,11 @@ describe(__filename, function () {
       done();
     });
   });
+
+  it('should parse recursive wsdls', function (done) {
+    open_wsdl(path.resolve(__dirname, 'wsdl/recursive/file.wsdl'), function (err, def) {
+      // If we get here then we succeeded 
+      done( err );
+    });
+  });
 });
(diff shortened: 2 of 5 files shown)


fabiancook:
May want to limit the caching of external ( http/s ) files, may be reloading because of change? Or maybe have an option to disable this

lib/wsdl.js
```diff
@@ -1130,7 +1130,7 @@ WSDL.prototype._processNextInclude = function(includes, callback) {
     includePath = url.resolve(self.uri, include.location);
   }
 
-  open_wsdl(includePath, this.options, function(err, wsdl) {
```
jsdevel: why wouldn't we just fix `open_wsdl`?  I also find `open_wsdl_private` to be very ambiguous.

fabiancook: Should we add the same code to open_wsdl, reasoning for moving it out was to limit it to internal invocation

jsdevel: My point is that `open_wsdl` is only ever used internally.

fabiancook: It is exposed using exports.open_wsdl, will add it to this function

fabiancook: open_wsdl is used in lib/soap, tests started to fail as I modified the original function, thoughts on current changes?

jsdevel: > It is exposed using exports

This is correct.  I was wrong.

jsdevel: > open_wsdl is used in lib/soap, tests started to fail as I modified the original function, thoughts on current changes?

I generally do not prefer exposing internal functions for test purposes. It seems to me we should fix `open_wsdl` instead of creating a new function.  Does `open_wsdl` even make sense as a name if it can't handle recursive wsdls?  Seems we should rename it to `open_non_recursive_wsdls`.

test/wsdl-parse-test.js
```diff
@@ -14,4 +15,13 @@ describe(__filename, function () {
       done();
     });
   });
+
+  it('should only parse a WSDL once', function(done) {
```
jsdevel: From what I can tell, this test does _not_ test `require('soap').createClient`.

lib/wsdl.js
fabiancook: That would make sense, then should the other be called `open_recursive_wsdls`?

test/wsdl-parse-test.js
fabiancook: Sorry, I don't understand why it would test `require('soap').createClient`, this is related to the WSDL, not `createClient` ?

jsdevel: `createClient` is only using `open_wsdl`.  Seems we ultimately care that `require('soap').createClient` will not fail on a recursive wsdl.

lib/wsdl.js
jsdevel: I would call it `open_wsdl_recursively`.

test/wsdl-parse-test.js
fabiancook: From `require('soap').createClient` point of view, it calls `open_wsdl`, when the WSDL is doing it's imports that's when the recursive WSDLs come into play, proving that a WSDL is only parsed once ( by returning the exact same WSDL object ) we can prove that a WSDL isn't going to set of an infinite loop

I can add a test with recursive WSDLs using `open_wsdl`

jsdevel: > I will add a test with recursive WSDLs using open_wsdl

Thank you!  I believe that would be the best approach.  If you do that, then I believe you can avoid exporting the new function used internally by `WSDL`.

fabiancook: There we go, should be good. Removed the export

lib/wsdl.js
```diff
@@ -1875,7 +1875,47 @@ WSDL.prototype._xmlnsMap = function() {
   return str;
 };
 
+var loading = { };
```
jsdevel: can you rename this `WSDL_CACHE`?  I think we're good after that.

test/wsdl/recursive/B.xsd
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xs:schema targetNamespace="http://www.Dummy.com/Name/Types"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="http://www.Dummy.com/Name/Types">
+  <xs:import namespace="http://www.Dummy.com/Common/Types" schemaLocation="A.xsd" />
```
jsdevel: should this be `./A.xsd`?

test/wsdl/recursive/A.xsd
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xs:schema targetNamespace="http://www.Dummy.com/Common/Types"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="http://www.Dummy.com/Common/Types"
+  >
+  <xs:import namespace="http://www.Dummy.com/Name/Types" schemaLocation="B.xsd" />
```
jsdevel: should this be `./B.xsd`?

jsdevel:
looking really good @fabiancook !  Couple minor comments yet, but once those are finished it'll be exciting to merge this!  Nice work.

test/wsdl/recursive/A.xsd
fabiancook: This is correct `file.wsdl` imports `A.xsd` and `B.xsd`, `A.xsd` imports `B.xsd` and `B.xsd` imports `A.xsd`

With the code currently published `file.wsdl` would import `A.xsd` which would then import `B.xsd` and then `A.xsd` and so on

With the change `file.xsdl` would import `A.xsd` which would then import `B.xsd`, which would then use the reference to the previously imported `B.xsd`

If `A.xsd` referenced `A.xsd` and `B.xsd` referenced `B.xsd` it would work the exact same.

lib/wsdl.js
fabiancook: Should this be a time limited thing as well, if the WSDL was imported in the last few minutes then use it, else get a new WSDL, could imagine this being a problem for someone out there, or even a way to "invalidate" the WSDL, maybe a future feature, food for thought

test/wsdl/recursive/A.xsd
fabiancook: This should mimic 


    file.wsdl
        xs:import namespace="A" schemaLocation: A.xsd
    A.xsd
        xs:import namespace="B" schemaLocation: B.xsd
    B.xsd
        xs:import namespace="A" schemaLocation: A.xsd

fabiancook:
Should be good to go now.

jsdevel:
Sure is.  Thank you!

fabiancook:
Awesome, thanks. Will hope to do a few more pull requests eventually :).

jsdevel:
Looking forward to it :dancers:

jsdevel:
published `v0.9.5` with this change.

PR #749, opened 2015-10-16
correct merging multiple XSDs from the same url namespace

jbarrus:
This resolves an issue I'm having with XSDs that import other XSDs with the same namespace url.  Using _.merge results in some values getting overwritten and the final wsdl definitions missing types.  Switching to the merge module's merge.recursive resolved the issue for me.  Sorry I don't have any test cases to demonstrate the problem/resolution.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -21,7 +21,7 @@ var selectn = require('selectn');
 var merge = require('merge');
 
 //shim defaultsDeep (it's not in lodash 2.x but is in lodash 3.x)
-var defaultsDeep = _.partialRight(_.merge, _.defaults);
+var defaultsDeep = _.partialRight(_.merge, _.default);
 
 var mergeMethod = merge.recursive;
 function set_merge_method(method) {
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -20,6 +20,16 @@ var _ = require('lodash');
 var selectn = require('selectn');
 var merge = require('merge');
 
+var mergeMethod = merge.recursive;
+function set_merge_method(method) {
+  if (method === 'lodash') {
+    mergeMethod = _.merge;
+  } else if (method === 'merge') {
+    mergeMethod = merge.recursive;
+  }
+}
+
+
 var Primitives = {
   string: 1,
   boolean: 1,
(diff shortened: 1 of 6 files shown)


herom:
Thanks a lot for your contribution @jbarrus - we appreciate it a lot :+1: 

But, wouldn't it be sufficient to use `_.defaults` instead of `_.merge` in order to get the same result? (See lodash's API documentation on [_.defaults](https://lodash.com/docs#defaults)). If so, I'd prefer to use `lodash` as I'm not a fan of introducing new dependencies for new functions/methods - especially if they are a "one trick pony" :smiley:

jsdevel:
> If so, I'd prefer to use lodash as I'm not a fan of introducing new dependencies for new functions/methods - especially if they are a "one trick pony" :smiley:

Best comment I've read in a while @herom :smile:

herom:
I know it's against the whole "micro-service" thing, but I'm a fan of "if I already have a thing that is able to do it - use it" :smile_cat:

jbarrus:
I don't blame you, I wouldn't have accepted it either but I thought I'd give it a try :)  I love lodash so I would prefer to use that as well.  I still don't really know how merge.recursive differs from lodash merge, but this proves the discrepancy.

jbarrus:
Side note, I added a clear_cache method to wsdl.js and put it in a beforeEach for the tests.  The other test stuff should be removed after you can verify it, but clearing cache between tests is probably a good idea.

lib/wsdl.js
```diff
@@ -1957,5 +1975,13 @@ function open_wsdl(uri, options, callback) {
   return wsdl;
 }
 
+function clear_cache() {
+  WSDL_CACHE = {};
```
jsdevel: instead of assigning a new object, I feel you should delete all the properties using `Object.keys(WSDL_CACHE).forEach(key) {delete WSDL_CACHE[key];});` or something similar to avoid stale references from lingering around.

test/wsdl-parse-test.js
```diff
@@ -21,4 +27,69 @@ describe(__filename, function () {
       done( err );
     });
   });
+
+  it('should parse multiple xsd files with the same namespace using lodash merge', function (done) {
+    wsdl.set_merge_method('lodash.default');
```
jsdevel: This is extremely weird to me.  What is the purpose of this method?

jbarrus: It's test code that needs to be removed after being reviewed.  I'm only trying to demonstrate that merging the definitions from two XSDs does not work with lodash merge or lodash defaults.

lib/wsdl.js
jbarrus: I don't mind changing that, I don't think I've seen that problem before.  Is there some basis for this other than a feeling?

jsdevel: It's generally better to clear object references to avoid bugs down the road.  I've been bitten by this many times, so it's more about experience rather than a feeling.

jbarrus: Okay, I guess I can see where it could make a difference.  If another reference had been created to WSDL_CACHE, it would be left with the old values if WSDL_CACHE was just reassigned.

jbarrus:
Does this make sense yet?  Can I remove the test cases that show the problems with lodash merge and defaults?

Issue #748, opened 2015-10-15
Possible Array serialization issue


theanimal666:
Hallo,

First of all thanks for this great library.

I'm working on an API and having troubles because of the way arrays get serialized into the request body.

My input object looks like:

```
{
options:
  [{
   option: {
      optionName: 'optname1',
      optionValue: 'value1'
    }
   },{
    option: {
      optionName: 'optname2',
      optionValue: 'value2'
    }
 }]
}
```
it gets serialized to:
```
<options>
   <option>
       <optionName>optname1</optionName>
       <optionValue>value1</optionValue>
   </option>
</options>
<options>
   <option>
       <optionName>optname2</optionName>
       <optionValue>value2</optionValue>
   </option>
</options>
```

while the server expects it to be formatted as:
```
<options>
   <option>
       <optionName>optname1</optionName>
       <optionValue>value1</optionValue>
   </option>
   <option>
       <optionName>optname2</optionName>
       <optionValue>value2</optionValue>
   </option>
</options>
```

I found the responsible code (llib/wsdl.js l: 1528..) and implemented a fix in my local fork but this causes several tests to fail.
I'm not really deep into this SOAP/XML stuff an therefore i'm not sure if that is a issue of the library or just bug in the servers SOAP server implementation.

thanks in advance for you help,
kind regards

jsdevel:
please open a PR

PR #750, opened 2015-10-16
Fix Array serialization

theanimal666:
This PR is not to be merged (test are failing!) it's just to demonstrate the issue originally described in https://github.com/vpulim/node-soap/issues/748, 

original issue text below:

Hallo,

First of all thanks for this great library.

I'm working on an API and having troubles because of the way arrays get serialized into the request body.

My input object looks like:

```
{
options:
  [{
   option: {
      optionName: 'optname1',
      optionValue: 'value1'
    }
   },{
    option: {
      optionName: 'optname2',
      optionValue: 'value2'
    }
 }]
}
```
it gets serialized to:
```
<options>
   <option>
       <optionName>optname1</optionName>
       <optionValue>value1</optionValue>
   </option>
</options>
<options>
   <option>
       <optionName>optname2</optionName>
       <optionValue>value2</optionValue>
   </option>
</options>
```

while the server expects it to be formatted as:
```
<options>
   <option>
       <optionName>optname1</optionName>
       <optionValue>value1</optionValue>
   </option>
   <option>
       <optionName>optname2</optionName>
       <optionValue>value2</optionValue>
   </option>
</options>
```

I found the responsible code (llib/wsdl.js l: 1528..) and implemented a fix in my local fork but this causes several tests to fail.
I'm not really deep into this SOAP/XML stuff an therefore i'm not sure if that is a issue of the library or just bug in the servers SOAP server implementation.

thanks in advance for you help,
kind regards

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1525,14 +1525,15 @@ WSDL.prototype.objectToXML = function(obj, name, namespace, xmlns, isFirst, xmln
   }
 
   // start building out XML string.
+  var correctOuterNamespace;
   if (Array.isArray(obj)) {
     for (var i = 0, item; item = obj[i]; i++) {
-      var arrayAttr = self.processAttributes(item),
-          correctOuterNamespace = parentNamespace || ns; //using the parent namespace if given
+      var arrayAttr = self.processAttributes(item);
+      correctOuterNamespace = parentNamespace || ns; //using the parent namespace if given
 
-      parts.push(['<', correctOuterNamespace, name, arrayAttr, xmlnsAttrib, '>'].join(''));
+      if( i === 0 ) parts.push(['<', correctOuterNamespace, name, arrayAttr, xmlnsAttrib, '>'].join(''));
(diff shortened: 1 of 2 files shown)


theanimal666:
Hi, 

As written im the PR i DID NOT WANT this to be merged - I KNOW that there are tests missing. 
My Point was that i wanted to know if this is really a bug in the library - or the expected bahaviour...

kind regards

jsdevel:
> i wanted to know if this is really a bug in the library - or the expected bahaviour

I think if it doesn't follow the standards then its a bug.

theanimal666:
i fixed the issue in a fork as i need it for a project. As i'm not into SOAP internals I HAVE NO IDEA what the standard here is... (already wrote that in the initial PR) 
So before i get my hand dirty implementing tests for the fix i simply wanted to know if it's worth the work as i'm not into the code and it would require quite some time for me to provide tests.

kind regards

LeandroFavero:
I have the same problem. Delphi use this "standard" and a TOTVS (https://www.totvs.com/) too.

My idea is has an option in module, to setup each kind of serialization use.

PR #753, opened 2015-10-16
improve array handling

jbarrus:
This isn't really complete and I was hoping for some feedback.  I have found that the array handling is only working correctly for me if the result has more than one element by the same name.  It looks like it fails to read the schema correctly and is only formatting arrays correctly by implying, rather than using the schema.  I think it's trying to add '[]' to the end of names hint this, but only under a condition I didn't really understand and am not hitting in the schemas I am working with (something to do with ref being used I think).

Thanks for your work on this library, it's great!

The example assumes items1 and item2 have a schema something like this:
```xsd
<xsd:sequence>
  <xsd:element ref="itemElement" minOccurs="0" maxOccurs="unbounded"/>
</xsd:sequence>
```

before change:

```xml
<items1>
  <item>
    <name>test1</name>
  </item>
</items1>
<items2>
  <item>
    <name>test1</name>
    <name>test2</name>
  </item>
</items2>
```

```javascript
{
  items1: {item: {name: 'test1'}},
  items2: [
    {name: 'test1'},
    {name: 'test2'}
  ]
}
```

after change:

```xml
<items1>
  <item>
    <name>test1</name>
  </item>
</items1>
<items2>
  <item>
    <name>test1</name>
    <name>test2</name>
  </item>
</items2>
```

```javascript
{
  items1: [
    {name: 'test1'}
  ],
  items2: [
    {name: 'test1'},
    {name: 'test2'}
  ]
}
```


jsdevel:
Looks good to me.

lib/wsdl.js
```diff
@@ -883,6 +885,9 @@ ElementElement.prototype.description = function(definitions, xmlns) {
 
         if (this.$ref) {
           element = elem;
+          if (arrayHint) {
```
herom: As `arrayHint` is guaranteed to be set correctly after its initialization, I would say it's unneccessary to check for `arrayHint` to be truthy before setting it. Just do `element.arrayHint = arrayHint;`  which should be save as it's either `false` or `true` :smiley:

lib/wsdl.js
```diff
@@ -898,6 +903,10 @@ ElementElement.prototype.description = function(definitions, xmlns) {
       else {
         if (this.$ref) {
           element = definitions.descriptions.types[typeName];
+
+          if (arrayHint) {
+            element.arrayHint = true;
```
herom: Same as above :+1:

jbarrus:
I tried this out on a few different requests and found that this causes issues with other elements that are not arrays.  It looks like adding [] to the name affects the usage of an element, not the element definition itself.  Adding the arrayHint property affects the element, so all usages of the element become arrays.

This can be overcome by adding ```_.cloneDeep()``` where typeElement.description(...) is called so that the original is not affected.  I'm not sure what other negative impact this might have though.

jsdevel:
Thank you for closing.

PR #754, opened 2015-10-19
Add missing type attribute for PasswordText in WSSecurity

wmcmurray:
I just encountered a SOAP API from a big client that required the password type to be set in any cases, even with the default "PasswordText" type so... voilà !

Thank you :)


hstonge:
+1

jsdevel:
@wmcmurray is there any way this can be a breaking change for existing uses using `BasicAuth`?

wmcmurray:
@jsdevel according to [this document](http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-username-token-profile-1.0.pdf) (around line 169) the #PasswordText type is the default, so if the type attribute is not specified, it'll be considered of this type anyway so I don't think it's a breaking change... but I am not a SOAP expert ;)

jsdevel:
Awesome @wmcmurray !  Verified in the `.xsd` file that `Type` is indeed required, so I believe this is safe.  Can you squash down to a single commit?

wmcmurray:
Cool ! Commits are now squashed :)

jsdevel:
Thanks @wmcmurray !

PR #756, opened 2015-10-20
Fix xml namespace/element/type handling

raymondfeng:
/to @jsdevel 

The PR improves the XML namespace handling for JSON to XML transformation. 

1. Introduce a `NamespaceContext` class in `lib/nscontext.js` to keep track of namespace prefix/uri mappings and declarations in the hierarchy. 

2. Rename the variables so that they are consistent with XML terminology to avoid confusions.

3. With 1, if the same prefix/uri mapping is declared at ancestor elements, we don't repeat such `xmlns:prefix=uri`

4. Fix some test cases that reference the wrong element names

5. Fix/enhance the resolution to various schema objects, such as local element, global element, element ref, type ref. Better handing the local element qualification form.

6. Fix the allowed child elements based on XML schema spec, for example, `sequence` can be a child element of `sequence`.

Please review.


jsdevel:
The changes look very promising.  I have a couple of questions:

1. Why rename `DummyFilterString` to `DummyStringFilter`?
2. Why were request response samples (not related to `DummyFilterString`) changed?

raymondfeng:
> 1. Why rename DummyFilterString to DummyStringFilter?

`DummyStringFilter` is the element name defined in the XSD.

> 2. Why were request response samples (not related to DummyFilterString) changed?

It's related to item 3 to dedup the prefix/uri declarations at child levels.

test/wsdl-test.js
```diff
@@ -89,6 +89,18 @@ wsdlStrictTests['should handle element ref'] = function(done) {
   });
 };
 
+wsdlStrictTests['should handle type ref'] = function(done) {
+  var expectedMsg = require('./wsdl/typeref/request.xml.js');
+  var req_json = require('./wsdl/typeref/request.json');
```
jsdevel: camelCase variable name

test/wsdl/elementref/bar1.xsd
```diff
@@ -5,34 +5,30 @@
     <xsd:element name="paymentRq">
         <xsd:complexType>
             <xsd:sequence>
-                <xsd:sequence>
```
jsdevel: are these changes needed herein?  or is it a cleanup to avoid redundancy?

test/request-response-samples/addList__complex_extension_namespace_for_arrays/request.xml
```diff
@@ -1 +1 @@
...
...ite.com" xmlns:platformMsgs="urn:messages_2013_2.platform.webservices.netsuite.com" xmlns:platformFaults="urn:faults_2013_2.platform.webservices.netsuite.com"><soap:Body><platformMsgs:addList xmlns:platformMsgs="urn:messages_2013_2.platform.webservices.netsuite.com" xmlns="urn:messages_2013_2.platform.webservices.netsuite.com"><platformMsgs:record xsi:type="ns1:Contact" xmlns:ns1="urn:relationships_2013_2.lists.webservices.netsuite.com"><ns1:firstName>First-1400867779067</ns1:firstName><ns1:lastName>Sullivan</ns1:lastName><ns1:phone>800-555-2819</ns1:phone><ns1:subsidiary internalId="1" xsi:type="platformCore:RecordRef" xmlns:platformCore="urn:core_2013_2.platform.webservices.netsuite.com"></ns1:subsidiary></platformMsgs:record><platformMsgs:record xsi:type="ns1:Contact" xmlns:ns1="urn:relationships_2013_2.lists.webservices.netsuite.com"><ns1:firstName>aLFirst-1400867779067</ns1:firstName><ns1:lastName>Sullivan</ns1:lastName><ns1:phone>800-555-2819</ns1:phone><ns1:subsidiary internalId="1" xsi:type="platformCore:RecordRef" xmlns:platformCore="urn:core_2013_2.platform.webservices.netsuite.com"></ns1:subsidiary></platformMsgs:record></platformMsgs:addList></soap:Body></soap:Envelope>
```
jsdevel: Isn't it now an error that `platformCore` isn't declared?

jsdevel: Nvm.  I see now that it is declared on the element that needs it.

jsdevel:
@raymondfeng have you provided tests for item 6?

jsdevel:
> Fix some test cases that reference the wrong element names

Why weren't these failing before?

test/wsdl-test.js
raymondfeng: sure

test/wsdl/elementref/bar1.xsd
raymondfeng: I just cleaned up to avoid redundancy. With the proposed changes, the nesting sequence is supported too. I can add it back.

raymondfeng:
If the code fails to look up a schema object by name, it assumes that element is under the parent namespace. And it works by coincidence.

raymondfeng:
I fixed the camelCase and nesting sequence.

lib/wsdl.js
```diff
@@ -1877,28 +1982,24 @@ WSDL.prototype._xmlnsMap = function() {
 
 var WSDL_CACHE = { };
 
-/* 
-  Have another function to load previous WSDLs as we
-  don't want this to be invoked externally (expect for tests)
-  This will attempt to fix circular dependencies with XSD files, 
-  Given
-
-    file.wsdl
```
jsdevel: I like this formatting, but see now that each line should start with `*`.  Can you add that?

jsdevel:
Overall it looks good to me.  I'd just go through it and ensure that each change was intended (no auto indending etc.).  I'd also like @herom to review this.

lib/wsdl.js
raymondfeng: Sure, that's messed up from the merge.

raymondfeng:
@jsdevel Thanks for the review. I'm still not happy about the JSON to XML object mapping as-is. The `description` phase already builds up the tree of elements as the template for XML structure. We should reuse that knowledge to greatly simplify the mapping code and improve performance.

lib/client.js
```diff
@@ -6,9 +6,11 @@
 "use strict";
 
 function findKey(obj, val) {
```
jsdevel: can we extract this out into a utility file?

jsdevel:
>  I'm still not happy about the JSON to XML object mapping as-is.

I hear you.  `NamespaceContext` is going to be a great step in moving us more towards a robust soap library though, so I'm glad you've done this work!

> The description phase already builds up the tree of elements as the template for XML structure. We should reuse that knowledge to greatly simplify the mapping code and improve performance.

Couldn't agree more.

lib/client.js
raymondfeng: Sure

jsdevel:
Another great PR @raymondfeng !  Thank you!

Issue #459, opened 2014-09-14
Request to a service with self signed SSL certificate


franza:
I'm trying to create a client to SOAP service that runs over HTTPS using self-signed SSL certificate:

``` javascript
var soap = require('soap');
var url = 'https://example.com/wsdl/Service.wsdl';

soap.createClient(url, function (err, client) {
  if (err) {
    return console.log(err.message);
  }
  console.log(client.describe());
});
```

However I get error: `[Error: DEPTH_ZERO_SELF_SIGNED_CERT]`. Is there some kind of option to avoid this behavior?

tesfel:
Add the cert to your trusted list at you computer or add

process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"

to your code.

brmarick:
I realize this is a bit of an old post, however I recently had this problem. While @tesfel suggestion is nice, it opens __vulnerabilities__. I was able to solve this problem via adding the certificates or certificate chain used on the soap server. In particular, I had to add the certificates in two locations, creating the client and the setting the security. 

### createClient
It took some digging, but I found the process to validate a server certificate or use a self-signed certificate on a server. To do this simply make use  of the 'options' parameter for createClient. To set the certificate(s) used by the server set the 'wsdl_options.ca' element. This element takes in an array of Base 64 certificates. Here is an example of how to complete this: 

```javascript
var cert = ['-----BEGIN CERTIFICATE-----\n....--END CERTIFICATE-----'];
var list_cert = [ '-----BEGIN CERTIFICATE-----\n....--END CERTIFICATE-----', ..., '-----BEGIN CERTIFICATE-----\n....--END CERTIFICATE-----'];

var options = {
    wsdl_options: {
        ca: cert,
    },
};
soap.createClient(soap_wsdl, options, function(error, client) {
    if(error) {
        console.log(error)
        return
    }
    client.MyFunction(args, function(err, result) {
        console.log(result);
    });
});
```

### setSecurity
Additionally, you can also run into the self-signed certificate error if you are using an authentication format. Again the solution is to set the ca property in the options parameter of the different security objects. Here is an example for BasicAuthSecurity.

```javascript
securitySettings = { ca: ['-----BEGIN CERTIFICATE-----\n....--END CERTIFICATE-----'] };
client.setSecurity(new soap.BasicAuthSecurity(username, password, securitySettings));
```

I will try to work up a documentation pull request to help document this process.

jsdevel:
[...]

PR #761, opened 2015-10-21
Add Client Certificate Documentation

brmarick:
#459 
Specifically, shows how to solve self-signed certificate errors,
[SELF_SIGNED_CERT_IN_CHAIN], and validate a server's certificate.
Includes a use case example.

Additonally, adjusted existing documentation for 'Client.*method*' to
contain the 'options' parameter that is used in the client certificate
validation.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -258,7 +258,7 @@ as default request options to the constructor:
   client.setSecurity(new soap.BearerSecurity('token'));
 ```
 
-### Client.*method*(args, callback) - call *method* on the SOAP service.
+### Client.*method*(args, callback[, options]) - call *method* on the SOAP service.
 
 ``` javascript
   client.MyFunction({name: 'value'}, function(err, result, raw, soapHeader) {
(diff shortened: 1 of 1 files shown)


jsdevel:
Is this accurate across node versions?

brmarick:
This actually relies on the `request` package. While, I have not confirmed on older versions. I would assume that it would be the same.

jsdevel:
@brmarick I believe the documentation is incorrect.  `client.setSecurity(new soap.security.ClientSSLSecurity(pathToKey, pathToCert, pathToCaFile));` should be used instead if I'm not mistaken.

brmarick:
The documentation is correct, but you can also set it in that location. You can actually set the path to the CA files in all security __types__ by passing it into the [objects parameter](https://github.com/vpulim/node-soap/blob/master/lib/client.js#L176) .
The reason I did not include the documentation for  this functionality was due to the length of other sections. My section already is one of the longest sections. I did not want to completely over haul the documentation. Would you like me to add this documentation for this option?

jsdevel:
The documentation now implies that you should always use the api you've described when others are available I.E. `client.setSecurity(...)`.  Perhaps we need to have a separate `Security.md` file to describe all things security.  Can you handle that in this PR and migrate your current proposal therein?

Thank you very much for doing this!

brmarick:
Yes, I can take the AR. However I do have a couple question based around this new task. 
* Should a directory be created to contain documentation? 
* If so, how do we want to split up the documentation. I realize this expands past our current topic, but feel like this is a valid topic to raise. 
* If not, where should I locate this document?

jsdevel:
> Should a directory be created to contain documentation?

I vote yes.

> If so, how do we want to split up the documentation. I realize this expands past our current topic, but feel like this is a valid topic to raise.

I completely agree.  I envision something like this:

```
documentation/
  getting-started.md
  security.md
  soap-options.md
  working-with-namespaces.md
```
Then links in the README to the appropriate doc file (like a table of contents).

brmarick:
Sound good to me. I will begin working on this item as soon as possible. Do we want to pull in any of the other Collaborators on this topic?

jsdevel:
> Do we want to pull in any of the other Collaborators on this topic?

They're watching now :smile:

jsdevel:
@brmarick any update on this?

jsdevel:
closing due to lack of response.

PR #767, opened 2015-10-28
Add signing of outgoing envelope

fabiancook:


fabiancook:
This is the start of a few commits, starting the discussion.

fabiancook:
@jsdevel have any opinions for naming? Should 'signing' be something else, signing is only related if you are wanting to sign the request, but they could just be interceptors in general

Readme.md
```diff
@@ -206,6 +206,23 @@ terminated.
   };
 ```
 
+### Server.*addSigner*(signer) - use the specified function to sign the response
+
+With a signer you can intercept the response envelope before it is sent to the client.
```
jsdevel: should this say `request envelope` instead?

lib/client.js
```diff
@@ -21,10 +23,50 @@ var Client = function(wsdl, endpoint, options) {
...
+  this._signers.push(signer);
+};
+
+/**
+* Sign an envelope
+* @param {String} envelope The envelope to sign
+* @param {ClientEnvelopeSignerOptions} options Options to pass to the signer
+* @param {NodeCallback} callback Function to be called once the envelope is signed
+*/
+Client.prototype._sign = function(envelope, options, callback) {
```
jsdevel: Is it possible to avoid this method altogether and just call `sign` directly?

lib/server.js
```diff
@@ -72,73 +75,151 @@ Server.prototype._initializeOptions = function(options) {
   this.wsdl.options.attributesKey = options.attributesKey || 'attributes';
 };
 
+/**
+* @callback NodeCallback
+* @param {?*} error
+* @param {String} envelope
+*/
+
+/**
```
jsdevel: I feel the jsdocs here are overkill.

fabiancook: Fair enough, was following suit

lib/client.js
fabiancook: Yeah I guess, that wouldn't be a problem. Only thought was that you could have multiple. Will change to single

fabiancook:
Think I am having a derp moment and buggered my fork. Lets try again

PR #771, opened 2015-11-02
Fixed issue when an error was undefined: undefined

tomgco:
This occurs on responses that from what I can tell do not have an
xsi:type of string, Thus these objects would not have an internal $value
and would return undefined.

This commit keeps backwards compatibility and falls back to assuming the
error was a string.


herom:
Looks good to merge :+1: - thanks a lot :smile:

jguepin:
Do you plan to release a new version on npm with the fix soon? Thanks!

bnetter:
:+1:

Manyuuz:
:+1:

jsdevel:
done @jguepin sorry for the delay.

PR #773, opened 2015-11-04
Fixing issues with ignoredNamespaces on children

tomgco:
Previously it would ignore user defined ignores.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -621,8 +621,10 @@ MessageElement.prototype._createLookupTypeObject = function (nsString, xmlns) {
  * @private
  */
 MessageElement.prototype._getNestedLookupTypeString = function (element) {
+  var ignoredNamespaces = this.options ? this.options.ignoredNamespaces
+      : this.ignoredNamespaces;
   var resolvedType = '^',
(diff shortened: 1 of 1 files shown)


jsdevel:
Can you add to `test/request-response-samples`?

lib/wsdl.js
```diff
@@ -621,8 +621,10 @@ MessageElement.prototype._createLookupTypeObject = function (nsString, xmlns) {
  * @private
  */
 MessageElement.prototype._getNestedLookupTypeString = function (element) {
+  var ignoredNamespaces = this.options ? this.options.ignoredNamespaces
+      : this.ignoredNamespaces;
   var resolvedType = '^',
```
herom: Just an overall thought: wouldn't it be more sufficient if we merge/concat `this.options.ignoredNamespaces` with `this.ignoredNamespaces` on creation-time so we can call `this.ignoredNamespaces` in every case without having to re-assign and double-check it?

jsdevel: Seems you are right @herom.

jsdevel:
Closing from lack of activity.

PR #775, opened 2015-11-11
Remove SOAPAction http header in SOAP 1.2, extra header was causing s…

mszlapa:
SOAP 1.2 request should not contain SOAPAction http header. 

Sending this header causes error on some servers (observed on .net based server)


jsdevel:
Thanks!

PR #781, opened 2015-12-09
Wrong namespace on elements when complexType has same name

Mik13:
When a complexType and an element has the same name, and we set the element to a value, the element gets the wrong namespace (a new one, e.g. ns1), because  the `findChildSchemaObject`-function returns the complexType, not the element.

With this fix, it first looks for elements with the name, and if not found, returns the complexType.


jsdevel:
Thanks @Mik13 !

PR #783, opened 2015-12-22
WSDL-defined method namespaces are supported (http://www.w3.org/TR/ws…

poroh:
WSDL-defined method namespaces are supported (http://www.w3.org/TR/wsdl#_soap:body namespace attribute)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -197,7 +197,7 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
 
   if (input.parts || args === null) {
     assert.ok(!style || style === 'rpc', 'invalid message definition for document style binding');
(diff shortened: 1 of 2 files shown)


poroh:
Note that test is failed because it is not correct. If in binding section body defines use="encoding" for input/output then input must be in namespace defined by "namespace" attribute. In the test this statment is true for output expectation (response.xml) but it is not true for input expectation (request.xml).

jsdevel:
@poroh couple of things:

* Please update the tests to make the build pass.
* Please squash your commits.

poroh:
I've closed this pull request because have got that RPC/encoded style bit more complex than I thought before. Please take a look on article http://www.ibm.com/developerworks/library/ws-whichwsdl/

PR #785, opened 2016-01-11
added in addressing ns

jnyryan:
I was having trouble connecting to a .Net WCF WsHTTP endpoint until i added the following namespace to the xmlns Envelope

```
" xmlns:a=\"http://www.w3.org/2005/08/addressing\"";
```

I cannot get the code correct or find a solution online so i'm using this as a means of publishing a feature request. 

Apologies if i am breaking etiquette.


jsdevel:
@jnyryan this change broke a lot of tests.  If you're able to fix them and re-submit with added tests to verify your change, we'd love to re-consider.  Thanks!

PR #786, opened 2016-01-16
Update http.js for allowing multipart requests

volkmarbuehringer:
Hi,

this is a very small change. Almost all work is done  by pushing the multipart body into the
request-options.
The only problem is placing the xml into the first mulitpart body. This cannot be done from
outside.

With multipart support it is possible to use soap for uploading files to soap-based services.

The other possibility would be providing a callback function which allows to set the xml into
the body.

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -101,6 +101,10 @@ HttpClient.prototype.request = function(rurl, data, callback, exheaders, exoptio
   var self = this;
   var options = self.buildRequest(rurl, data, exheaders, exoptions);
   var headers = options.headers;
+  
+  if ( "multipart" in options ){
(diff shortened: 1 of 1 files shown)


jsdevel:
Hmm, not sure about including this without tests.  Can you add some?

volkmarbuehringer:
Hi,
I resolved the problem by setting a request module in options.
It works now with connection=keep-alive.
After setting the options I call request myself.
For streaming base64 data through request, I need to overwrite httpClient.request, because of
the stream.end() there.
A better solution would be to remove the stream.end() and letter the user control the
streaming, if the request function is set.

PR #788, opened 2016-01-25
Fix schema <xsd:include> support

aleung:
node-soap has support to <xsd:import> which imports schema definitions of _other_ target namespace. But it has issue in support to <xsd:include> which refer to definitions in the _same_ target namespace. Definitions (elements, types, complexTypes) from included file will overwrite existing definitions in the same target namespace.

In below example, after include file `types.xsd` is processed, elements `DummyRequest` and `DummyResponse` will be gone.

        <xs:schema targetNamespace="http://www.dummy.com/Types"
                   xmlns:xs="http://www.w3.org/2001/XMLSchema"
                   xmlns:dummy="http://www.dummy.com/Types"
                   elementFormDefault="qualified" attributeFormDefault="unqualified">
            <xs:include schemaLocation="types.xsd"/>
            <xs:element name="DummyRequest">
                <xs:complexType>
                    <xs:sequence>
        				<xs:element name="ID" type="dummy:IdType"/>
        				<xs:element name="Name" type="dummy:NameType"/>
                    </xs:sequence>
                </xs:complexType>
            </xs:element>
            <xs:element name="DummyResponse">
                <xs:complexType>
                    <xs:sequence>
                        <xs:element name="Result" type="dummy:DummyList"/>
                    </xs:sequence>
                </xs:complexType>
            </xs:element>
        </xs:schema>

The issue is because that lodash `merge()` can only merge plain object. `SchemaElement` is not plain object so after merge it's overwritten.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -344,6 +344,18 @@ DefinitionsElement.prototype.init = function() {
 DocumentationElement.prototype.init = function() {
 };
 
+SchemaElement.prototype.merge = function(source) {
+  assert(source instanceof SchemaElement);
+  if (this.$targetNamespace === source.$targetNamespace) {
+    _.merge(this.complexTypes, source.complexTypes);
+    _.merge(this.types, source.types);
+    _.merge(this.elements, source.elements);
+    _.merge(this.xmlns, source.xmlns);
+  }
+  return this;
+};
+
+
 SchemaElement.prototype.addChild = function(child) {
   if (child.$name in Primitives)
     return;
@@ -373,12 +385,6 @@ TypesElement.prototype.addChild = function (child) {
   assert(child instanceof SchemaElement);
 
   var targetNamespace = child.$targetNamespace;
-
-  if(!targetNamespace) {
-    if(child.includes && (child.includes instanceof Array) && child.includes.length > 0) {
-      targetNamespace = child.includes[0].namespace;
-    }
-  }
 
   if(!this.schemas.hasOwnProperty(targetNamespace)) {
     this.schemas[targetNamespace] = child;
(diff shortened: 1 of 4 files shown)


herom:
Thanks a lot for your time & effort @aleung - this looks really good! Please could you squash your commits to follow our [Guideline on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request), so that we have only a single commit per PR - this would be awesome!

aleung:
@herom I has squashed the commits.

lib/wsdl.js
```diff
@@ -344,6 +344,19 @@ DefinitionsElement.prototype.init = function() {
...
 };
 
+SchemaElement.prototype.merge = function(source) {
+  assert(source instanceof SchemaElement);
+  if (this.$targetNamespace === source.$targetNamespace) {
+    _.merge(this.complexTypes, source.complexTypes);
+    _.merge(this.types, source.types);
+    _.merge(this.elements, source.elements);
+    _.merge(this.xmlns, source.xmlns);
+    this.includes.concat(source.includes);
```
herom: Are you shure you didn't want to write `this.includes = this.includes.concat(source.includes);`? As `Array.prototype.concat` only returns a new concatenated Array but leaves the original Arrays untouched, `this.includes` would not change and therefore not contain the contents of `source.includes`.

aleung:
Line 354 can be removed. When a file is merged into its parent, its includes already have been processed.
I'll submit a new commit.

herom:
:+1:

aleung:
@herom Please review the new commit.

jsdevel:
lgtm

herom:
Thanks a lot @aleung - that's great :+1:

PR #789, opened 2016-01-28
Children may be empty resulting in a NPE

tobias-neubert:
I think the title says it all :-)

Tobi


jsdevel:
Can you add to `test/request-response-samples`?  We'll need a way to re-create this.

tobias-neubert:
Ok, added a test. Hope it helps.

jsdevel:
It does!  Thank you!  Can you squash down to 1 commit?

tobias-neubert:
What do I have to do to repeat the tests? It seems something out of my scope went wrong.

jsdevel:
@tobias-neubert there should be a rerun build button in travis, but it looks like coveralls is failing.See this issue: https://github.com/lemurheavy/coveralls-public/issues/713

jsdevel:
Thanks for the contribution @tobias-neubert !

PR #795, opened 2016-02-04
give complete control to overwritten request module

volkmarbuehringer:
by removing the streaming to request in the code (req.end() )
the request can be completly controlled by overwriting the request module

This is important for supporting file  up- and downloading through soap

options.body can be replaced by multipart-body,
the xml in options.body can be split and streaming of base64 data included
without req.end() in the code.

so req.write(), piping to and from request and request.end() can be done
in the overwritten function and then the request module be called which will
give control back to soap through the callback

In the moment this works only if I set "keep-alive" in the header.

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -102,19 +102,14 @@ HttpClient.prototype.request = function(rurl, data, callback, exheaders, exoptio
   var options = self.buildRequest(rurl, data, exheaders, exoptions);
   var headers = options.headers;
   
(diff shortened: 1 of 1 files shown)


herom:
sorry for the late reply @volkmarbuehringer - I completely forgot that I haven't responded :disappointed: 

could you please rebase/squash your commits into a single one in order to get your PR merged? thanks a ton :+1:

PR #796, opened 2016-02-07
consider previous ignoredNamespaces option when processing includes

bjrmatos:


herom:
Thanks a lot for your time & effort @bjrmatos :+1: 

It'd be much appreciated if you could provide also a test for your commit so that it's guaranteed that it doesn't get overrideen.

bjrmatos:
@herom sure, i'll work on that

bjrmatos:
for testability purposes i needed to keep a reference to includes wsdl objects inside a recursive wsdl

bjrmatos:
my last commit "WSDL_CACHE from global -> per instance" was needed because the WSDL_CACHE was global, when it saves a wsdl object it will save it with the first options that it finds, that its a problem because when it process a wsdl object for the first time it always be re-used and it will not take into account new options for next cases.

For example you can run the tests without the last commit, if you run the `"should parse recursive wsdls keeping provided options'"` test as standalone using `it.only()` you can see that the test pass, but if you run the whole suite of tests you can see the test fails because the global WSDL_CACHE.

a local WSDL_CACHE (per `new WSDL()` instance shared between its includes) makes more sense and it will not have these problems.

herom:
Could you please rebase and squash your commits into a single commit so that we can merge your PR? (see our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request))

bjrmatos:
@herom sure! sorry for not reading it

herom:
no problem :dancer:

bjrmatos:
done! :smiley:

herom:
Thanks a ton @bjrmatos :+1:

bjrmatos:
thnks! :heart:

PR #799, opened 2016-02-12
Fixing wrong nsContext.

Mik13:
The nsContext did not always get popped after getting pushed.


herom:
Thanks a lot for your time & effort @Mik13 - we appreciate it a lot :+1:

PR #800, opened 2016-02-22
Remove assertions around location of BEGIN CERTIFICATE and END CERTIFICATE

JosephJNK:
Currently node-soap throws an error if a certificate file does not have "------ BEGIN CERTIFICATE ------" as the first characters in the file. However, https://tools.ietf.org/html/rfc7468 provides a spec for encodings of X.509 certificates, and says:
```
"Textual encoding begins with a line comprising "-----BEGIN ", a label, and "-----", and ends with a line comprising "-----END ", a label, and "-----".  Between these lines, or "encapsulation boundaries", are base64-encoded data according to Section 4 of [RFC4648].  (PEM [RFC1421] referred to this data as the "encapsulated text portion".)  Data before the encapsulation boundaries are permitted, and parsers MUST NOT malfunction when processing such data.
```
In this PR I removed the check for BEGIN and END entirely. It may be possible to provide a more thorough verification while remaining true to the spec, by accepting a wider range of labels (such as "TRUSTED CERTIFICATE" in addition to "CERTIFICATE") and accepting them at any location in the file. I don't think this is worth it, though, as complete verification of the spec would be non-trivial, and since it's ultimately the receiving server's responsibility to validate the certificate.


herom:
I totally agree with you @JosephJNK, that the `soap` library should not verify certificates per se if it's used as client, what about you @jsdevel ?

jsdevel:
Yea, I feel this change is acceptable.  My only concern is on backward compatibility.  @JosephJNK can you see any issues with that?

JosephJNK:
@jsdevel I wouldn't expect any backwards compatibility issues, since this is strictly increasing the accepted set of inputs. The only way this would result in an error is if a user was depending on this error being thrown, like if they were trying to load multiple files as certs until one finally worked. I really doubt anyone is doing something like that, so this shouldn't cause any breakage.

JosephJNK:
Just pushed/squashed another change, I realized that I had saved the wrong cert for the unit test I added, and needed to add the preamble to it for the test to be valid.

jsdevel:
LGTM

herom:
Merged. Thanks a lot @JosephJNK :+1:

JosephJNK:
Sure thing, thanks for accepting the PR so quickly.

PR #801, opened 2016-02-24
Support x509 certificates

ctarrington:
This pull request would require the removal of support for node 0.8 due to its dependency on ursa which supports old and new node versions, but only if you install with engine-strict set to true. Unfortunately, even without this pull, node-soap will not work with node 0.8 and engine checking. strip-bom is the first library to fail to support 0.8 but is probably not alone...

### DIFF
--- a/lib/security/WSSecurityCert.js
+++ b/lib/security/WSSecurityCert.js
@@ -1,0 +1,78 @@
+"use strict";
+
+var ursa = require('ursa');
+var fs = require('fs');
+var path = require('path');
+var ejs = require('ejs');
+var SignedXml = require('xml-crypto').SignedXml;
+var uuid = require('node-uuid');
+var wsseSecurityHeaderTemplate = ejs.compile(fs.readFileSync(path.join(__dirname, 'templates', 'wsse-security-header.ejs')).toString());
+var wsseSecurityTokenTemplate = ejs.compile(fs.readFileSync(path.join(__dirname, 'templates', 'wsse-security-token.ejs')).toString());
+
+function addMinutes(date, minutes) {
+  return new Date(date.getTime() + minutes * 60000);
+}
+
+function dateStringForSOAP(date) {
+  return date.getUTCFullYear() + '-' + ('0' + (date.getUTCMonth() + 1)).slice(-2) + '-' +
+    ('0' + date.getUTCDate()).slice(-2) + 'T' + ('0' + date.getUTCHours()).slice(-2) + ":" +
+    ('0' + date.getUTCMinutes()).slice(-2) + ":" + ('0' + date.getUTCSeconds()).slice(-2) + "Z";
+}
+
+function generateCreated() {
+  return dateStringForSOAP(new Date());
+}
+
+function generateExpires() {
+  return dateStringForSOAP(addMinutes(new Date(), 10));
+}
+
+function insertStr(src, dst, pos) {
+  return [dst.slice(0, pos), src, dst.slice(pos)].join('');
+}
+
+function generateId() {
+  return uuid.v4().replace(/-/gm, '');
+}
+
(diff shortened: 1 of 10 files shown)


jsdevel:
@ctarrington looks like something is wrong with the commit history.  Can you take a look?

ctarrington:
Sorry about that - I accidentally added them when rebasing to squash my commits down to one.
Looks better to me.

jsdevel:
Looks OK to me.  Thoughts @herom ?

lib/security/WSSecurityCert.js
```diff
@@ -0,0 +1,81 @@
...
+function insertStr(src, dst, pos) {
+  return [dst.slice(0, pos), src, dst.slice(pos)].join('');
+}
+
+function generateId() {
+  return uuid.v4().replace(/-/gm, '');
+}
+
+function WSSecurityCert(privatePEM, publicP12PEM, password, encoding) {
+  if(publicP12PEM.toString().lastIndexOf('-----BEGIN CERTIFICATE-----', 0) !== 0) {
```
herom: I don't think we should "re-implement" the checks for `BEGIN` and `END` as we just removed the checks from the `ClientSSLSecurity.js` (see PR #800).

jsdevel: +1

lib/security/WSSecurityCert.js
```diff
@@ -0,0 +1,81 @@
...
+
+  var _this = this;
+  this.signer.keyInfoProvider = {};
+  this.signer.keyInfoProvider.getKeyInfo = function (key) {
+    return wsseSecurityTokenTemplate({ x509Id: _this.x509Id });
+  };
+}
+
+WSSecurityCert.prototype.postProcess = function (xml) {
+  this.Created = generateCreated();
```
herom: For consitency and sanity of the code, I'd prefer if `this.Created` and `this.Expires` would be renamed to `this.created` and `this.expired`, as only "classes" which are instantiated with `new` should start with an uppercase letter.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.13.0",
   "description": "A minimal node SOAP client",
   "engines": {
-    "node": ">=0.8.0"
+    "node": ">=0.10.0"
```
herom: I think we should it mark clearly within the next release notes, that we've dropped the support for node 0.8 with this commit.

ctarrington:
I removed the BEGIN check and fixed the case of the fields.

Thanks for the quick responses

herom:
Thanks a lot @ctarrington :+1:

ctarrington:
Props to @darful for doing the real lift on the certificates!
And many thanks to @herom and @jsdevel for making the process quick and painless!

package.json
jsdevel: :+1:

CageFox:
Please consider to unmerge this commit. It brings heavy dependancy upon ursa and really x509 support not needed, as anyone can easily convert x509 to key/crt
Ursa installing on windows is really pain and requires to install binary packages, what is unacceptable.

herom:
On which machine (OS, node.js, npm ... versions) are you running your code @CageFox ?

CageFox:
Windows 10 x64, Node 5, npm 3.7.3

batjko:
I'm very new to certificates and all that, but if I have received a certificate file (with a password) from the vendor to whose service I am trying to connect, how do I use this package then? 
The `.WSSecurityCert` function seems to ask for private and public keys, and I'm not sure if I should just call it like this, i.e. setting the private key to null: 
`new soap.WSSecurityCert(null, publicKey, password, 'utf8');`

(I am currently trying to make `ursa` work, due to the usual node-gyp issues on Windows, so I haven't been able to try the above yet.)

PR #802, opened 2016-02-26
Note about keep-alive workaround

kevincoleman:
In newer versions of node and request, `connection: close` in SOAP headers can result in truncated response XML due to a connection termination before all data chunks have arrived. Changing `connection` to `keep-alive` solves this problem. This is untested and may cause issues in some environments, so I thought I’d add this as a documentation note. I’d test and PR the change, but I’ll be the first to admit that it’s beyond my skill level. This note is just to (possibly) save others the trouble I dealt with. You can toggle node versions [here](https://tonicdev.com/56be371d273d6b0c00c70599/56c60f5afc5fed0d00f34c2b) to see the error take place—this is even without involving node-soap, but the effect is the same.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,17 @@
+0.14.0 / 2016-04-12
+=================
+* [ENHANCEMENT] Allow to call methods with `callback` as last param in order to align with node.js `callback last` pattern (#814)
+* [ENHANCEMENT] Re-enabled `ignoreBaseNameSpaces` option (#809)
+* [FIX] Avoid overwriting request headers with options in client method invocation (#813)
+* [ENHANCEMENT] Accept `time` value in in `extraHeaders` options in order to retrieve the `lastElapsedTime` for the response (#811)
+* [ENHANCEMENT] Allow to set a custom envelope key for the SOAP request (#812)
+* [FIX] Removed double declaration of `WSDL` variable in `lib/soap.js` (#810)
+* [DOC] Added documentation for `wsdl_options` and `wsdl_headers` options in `createClient()` method (#806)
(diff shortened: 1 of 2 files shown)


jsdevel:
Would it be safe to just always set `connection: 'keep-alive'`?

kevincoleman:
@jsdevel that would be my first inclination, but I really don’t have high confidence in my ability to run the PR/testing gauntlet properly.

herom:
@jsdevel do we have a solution for it yet? if not, I think we should at least update our README as suggested by @kevincoleman

jsdevel:
yea, that sounds good.

herom:
sorry for the long delay @kevincoleman - but could you please rebase and squash your commits into a single one so that I'm able to merge your PR? this would be awesome :+1:

kevincoleman:
Just learned how to squash commits. #TIL

Hope that’s what you were looking for! Is it normal to submit another PR? In any case, I did, and you can reject the first one. Thanks for baring with me as I figure this out...!

herom:
thanks a ton @kevincoleman  :+1:  - there's only just one thing left: could you please rebase your commit against `master`? this would be awesome!

kevincoleman:
Ok—I think I’ve got it figured out. Tried rebasing but kept having to deal with merge conflicts, more commits, more squashing... &c. Turns out the note is better moved down in the doc anyway, so I just reverted, re-updated (this time on a feature branch instead of right on master), rebased, merged, and pushed.

Learning new things in droves today— @herom thanks for your patience for such a simple little update...!

PR #803, opened 2016-03-02
XSD import in WSDL files and relative path (server creation)

srius:
I recently faced an issue creating a new SOAP service using a WSDL file in which XSD in included with relative path.
XSD loading is done from NodeJS working dir, instead of WSDL directory.

With this patch and setting uri option in "listen" function arguments everything is OK.


herom:
Can you please add tests @srius ?

jsdevel:
closing due to inactivity

PR #804, opened 2016-03-03
Make request comparisons in request-response-samples-test.js easier to work with

ctarrington:
Specifically, ignore whitespace only differences and make differences easier to spot.
In addition to the expected vs actual output as generated by the assert, add a colorized difference line adjacent to the test case in the list.


jsdevel:
This is great!  Thank you so much!

PR #805, opened 2016-03-04
Control the namespace definition of the root element via wsdlOptions

ctarrington:
In rare cases, control the namespace definition of the root element through wsdlOptions.


v4l3r10:
+1

herom:
Thanks a lot for your time & effort @ctarrington - I also do appreciate the extension of our README :+1:

PR #810, opened 2016-03-21
Update soap.js

upcu:
WSDL was declared and assigned twice


herom:
Thanks a lot @upcu :+1:

PR #812, opened 2016-03-30
Allow to set a custom envelop

tiste:


herom:
Thanks a lot @tiste & sorry for the delay :+1:

PR #813, opened 2016-03-31
Avoids overwriting request headers with options in method invocation

gabmontes:
When specifying `options` during a method invocation, those options will be added to the set of options already defined to invoke `request` and call an endpoint as specified in the [docs](https://github.com/vpulim/node-soap#options-optional).

But if a `headers` option is specified, it will overwrite all the already-set request headers including, for instance, any auth header previously set:

```javascript
client.MyService.MyPort.MyFunction({}, function(err, result) {
  // the HTTP request will **only** have 'my-header' header!!
}, {headers: {'my-header': 'header value'})
```

This PR takes into consideration `headers` is a special case and, instead of overwriting the `headers` property, it will merge the properties into the existing request headers:

```javascript
client.MyService.MyPort.MyFunction({}, function(err, result) {
  // the HTTP request will have 'my-header' in addition to any other existing header
}, {headers: {'my-header': 'header value'})
```

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -47,6 +47,8 @@ HttpClient.prototype.buildRequest = function(rurl, data, exheaders, exoptions) {
     'Host': host + (isNaN(port) ? '' : ':' + port)
   };
   var attr;
+  var header;
+  var mergeOptions = ['headers'];
 
   if (typeof data === 'string') {
     headers['Content-Length'] = Buffer.byteLength(data, 'utf8');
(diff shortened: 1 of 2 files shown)


gabmontes:
@jsdevel, @herom, `coveralls` failed for 2 out of 4 builds. Possibly an issue with TravisCI. Can the build be run again or I should close this PR and create a new one?

jsdevel:
@gabmontes  I believe travis has a button to re-build.

gabmontes:
@jsdevel thanks but it looks like only the ones with write access to the repo can trigger a rebuild. I will try resending the PR if you agree.

jsdevel:
restarted the jobs that failed.  coveralls has been having lots of issues lately.

jsdevel:
Looks good to me.  Thanks!

gabmontes:
@jsdevel thanks!!!

PR #814, opened 2016-04-11
Allows calling methods with callback as last param

gabmontes:
The main motivation behind this PR is to allow using a promise-based interface on top of this module.

The most common tools to promisify the async calls expect the callback to be the last argument but the SOAP methods are built with it in the second position, allowing a third and fourth optional params (`options`, and `extraHeaders`).

The proposed change allows more flexibility in method calls in order to support the following signatures:

```javascript
// standard and documented method signatures
client.MyFunction(args, callback)
client.MyFunction(args, callback, options)
client.MyFunction(args, callback, options, extraHeaders)

// non-documented but supported signature
client.MyFunction(callback)

// new supported signatures with callback as last param
client.MyFunction(args, options, callback)
client.MyFunction(args, options, extraHeaders, callback)
```

Documentation has been updated and tests added.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -348,28 +348,12 @@ as default request options to the constructor:
       // soapHeader is the response soap header as a javascript object
   })
 ```
-### Client.*service*.*port*.*method*(args, callback[, options]) - call a *method* using a specific *service* and *port*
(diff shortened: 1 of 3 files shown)


herom:
Thanks a lot @gabmontes - this seems like a really valuable addition to our library :+1:

gabmontes:
@herom great to hear that! Thanks!!!

PR #815, opened 2016-04-13
Duplicated ns

Hamatek:
Prevent from adding many times the same ns, by comparing the new ns with previous one

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -2011,6 +2011,7 @@ WSDL.prototype._fromServices = function(services) {
 WSDL.prototype._xmlnsMap = function() {
   var xmlns = this.definitions.xmlns;
   var str = '';
+  var oldNs = '';
   for (var alias in xmlns) {
     if (alias === '' || alias === TNS_PREFIX)
       continue;
@@ -2030,7 +2031,10 @@ WSDL.prototype._xmlnsMap = function() {
       continue;
     if (~ns.indexOf('http://xml.apache.org/'))
       continue;
-    str += ' xmlns:' + alias + '="' + ns + '"';
+      if(ns !== oldNs){
+      str += ' xmlns:' + alias + '="' + ns + '"';
+      oldNs = ns
+    }
   }
   return str;
 };


lib/wsdl.js
```diff
@@ -2011,6 +2011,7 @@ WSDL.prototype._fromServices = function(services) {
 WSDL.prototype._xmlnsMap = function() {
   var xmlns = this.definitions.xmlns;
   var str = '';
+  var oldNs = '';
```
jsdevel: Why not use an object or `_.uniq`?

herom:
I see 3 failing tests over at Travis CI, could you please take a look at it @Hamatek ?

Hamatek:
@herom I think the test is failing because the test is expecting a duplicated ns to be added: 
* expected : `xmlns:tns="urn:RpcExample" xmlns:RpcExample="urn:RpcExample"`

But the actual adds it only once
* actual : `xmlns:tns="urn:RpcExample"`

@jsdevel  You are right sounds like a good solution it will also do a better job since it  will make them uniq. The current solution only compare with the previous one

herom:
@Hamatek but this would not be wrong, would it? they are both declared under a different xmlns namespace so I guess it would break things if we omit one of them? am I wrong @jsdevel?

Hamatek:
@herom Im not sure either but I don't see why you would need 2 differents ns declaration to point to the same thing.
 I use this code in production and it works well on my hand, yet I don't know if it can be apply for all.

herom:
@Hamatek I see your point I also agree that it doesn't seem logical to have two (or more) different namespace declarations pointing to the same namespace, yet I'm a bit hesitant over adjusting the failing tests and merge the PR as it could probably break production apps... hmm....

Hamatek:
@herom  && @jsdevel any chance to merge this pull request It would be nice since I can't use it without this fix :(

jsdevel:
> ...et I'm a bit hesitant over adjusting the failing tests and merge the PR as it could probably break production apps...

I feel the same way @herom

jsdevel:
closing due to inactivity

kushbhandari:
I am also facing this issue, different name space declarations pointing to same namespace. I also see around 60 declarations pointing to same namespace. Any idea, What is the reason of generating diff namespace declarations pointing to same namespace ?

PR #822, opened 2016-04-25
Fix elementFormDefault handling.

aleung:
This PR has two changes.

The first is modification to the test cases which test handling of schema `elementFormDefault` setting.  `elementFormDefault` doesn't make sense when WSDL is in `rpc/encoded` style. I change the test cases to use `document/literal` style WSDL.

Below is a [reference](http://stackoverflow.com/a/28322548/94148) from SO:

> When using a literal use model, the body contents should conform to a user-defined XML-schema(XSD) structure. The advantage is two-fold. For one, you can validate the message body with the user-defined XML-schema, moreover, you can also transform the message using a transformation language like XSLT.
> 
> With a (SOAP) encoded use model, the message has to use XSD datatypes, but the structure of the message need not conform to any user-defined XML schema. This makes it difficult to validate the message body or use XSLT based transformations on the message body.

The second modification is to fix the issue identified by the new test case. The default value of `elementFormDefault` is 'unqualified'. When this attribute is missing in schema, the behaviour should be like 'unqualified'.


aleung:
Has anyone reviewed this PR?

jsdevel:
Can you keep the old files in place?  It's our only way of guaranteeing that this change isn't going to break existing functionality.

aleung:
@jsdevel I removed the old test cases because I believe they are incorrect. They were testing and expecting invalid behaviour.

jsdevel:
It's hard to see exactly what's changed being that the files were renamed.  Can you avoid renaming them in this PR?  Maybe have a follow up PR to correct the naming?

aleung:
I removed three test cases:
- execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_missing
- execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_qualified
- execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_unqualified

They verify how node-soap generates SOAP request in various value of `elementFormDefault` property in XML schema: property omitted, set to `qualified` and set to `unqualified`.

The problem is, in these test cases the WSDL are using **rpc/encoded** binding, it means that the SOAP body does not follow a schema. 

When WSDL is **document/literal**, the schema included in WSDL controls how the SOAP body is generated. The test case should use document/literal style WSDL so that the `elementFormDefault` in schema makes sense.

I created new test cases to replace the deleted ones. The WSDL in new test cases are using document/literal binding.
- Dummy__element_form_default_missing
- Dummy__element_form_default_qualified

There is already an existing test case which covers the 'unqualified' scenario, so I didn't add a duplicated one.
- Dummy__ref_element_should_has_correct_namespace (#450)

The deleted test cases where added in PR #377 'Fixup support for `elementFormDefault`' and no modification afterward. My newly added test cases have covered the same feature (actually it found out new defect) so I believe it doesn't reduce test coverage.

aleung:
@jsdevel I just see your new comment.

The test case folder naming is in pattern '<method>_<expectation>'. The 'execFile' is the method name in RPC style and looks not in convention of document style. Some other document style WSDL test cases uses 'Dummy' so I just copy it.

I can keep the old name of the test cases if it's necessary.

jsdevel:
Please keep the old names if you can, and limit changes to the test case files so we can see _exactly_ what is changing.

aleung:
The new commit keeps old names.

To make it easy to view, I post the formatted `request.xml` here.

execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_qualified:

```xml
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope
      xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xmlns:tns="urn:FileExecutor"
      xmlns:FileExecutor="urn:FileExecutor">
      <soap:Body>
            <FileExecutor:execFileParams
                  xmlns:FileExecutor="urn:FileExecutor"
                  xmlns="urn:FileExecutor">
                  <FileExecutor:fileName>PSTN</FileExecutor:fileName>
                  <FileExecutor:arguments>0</FileExecutor:arguments>
                  <FileExecutor:base64EncodedStdIn>null</FileExecutor:base64EncodedStdIn>
                  <FileExecutor:startDetached>true</FileExecutor:startDetached>
            </FileExecutor:execFileParams>
      </soap:Body>
</soap:Envelope>
```

execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_unqualified:

```xml
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope
      xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xmlns:tns="urn:FileExecutor"
      xmlns:FileExecutor="urn:FileExecutor">
      <soap:Body>
            <FileExecutor:execFileParams
                  xmlns:FileExecutor="urn:FileExecutor">
                  <fileName>PSTN</fileName>
                  <arguments>0</arguments>
                  <base64EncodedStdIn>null</base64EncodedStdIn>
                  <startDetached>true</startDetached>
            </FileExecutor:execFileParams>
      </soap:Body>
</soap:Envelope>
```

execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_missing:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope
      xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      xmlns:tns="urn:FileExecutor"
      xmlns:FileExecutor="urn:FileExecutor">
      <soap:Body>
            <FileExecutor:execFileParams
                  xmlns:FileExecutor="urn:FileExecutor">
                  <fileName>PSTN</fileName>
                  <arguments>0</arguments>
                  <base64EncodedStdIn>null</base64EncodedStdIn>
                  <startDetached>true</startDetached>
            </FileExecutor:execFileParams>
      </soap:Body>
</soap:Envelope>
```

herom:
Thanks a lot for your time & effort @aleung - that's amazing :clap: 

LGTM, @jsdevel?

test/request-response-samples/execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_missing/request.json
```diff
@@ -1,8 +1,6 @@
 {
-  "params" : {
     "fileName": "PSTN",
```
jsdevel: Isn't this a breaking change?

test/request-response-samples/execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_qualified/soap.wsdl
```diff
@@ -11,39 +23,40 @@
...
 			</input>
 			<output>
-				<SOAP:body use="encoded" namespace="urn:RpcExample" encodingStyle="http://www.w3.org/2003/05/soap-encoding"/>
+				<SOAP:body use="literal"/>
 			</output>
 		</operation>
 	</binding>
-	<service name="RpcExample">
-		<port name="RpcExample" binding="tns:RpcExample">
+	<service name="FileExecutorService">
```
jsdevel: I still don't see the reason why all these names where changed.

test/request-response-samples/execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_missing/request.json
aleung: No. It's by node-soap design. For rpc style SOAP web service, node-soap requires the request parameters being wrapped in "params" property in the input object. For document style it isn't wrapped. Since I change the WSDL from rpc style to document style, the input json has to be changed.

test/request-response-samples/execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_qualified/soap.wsdl
aleung: If I don't rename it, the name "RpcExample" will confuse other maintainers -- the WSDL isn't in RPC style any more. Another reason to rename is that the service name, binding name and port name need to be different, at least in some WSDL validator they're required to be different, e.g. https://www.wsdl-analyzer.com/. Before I renamed, they're the same "RpcExample".

jsdevel:
> If I don't rename it, the name "RpcExample" will confuse other maintainers

I understand this sentiment, but it needs to be in a separate PR.  As a maintainer, I find it very difficult to see what the ramifications of a change are when they're mixed in with a lot of refactoring.  Please revert to the old name, and follow up with a PR to correct the invalid name.

test/request-response-samples/execFile__valid_xml_generated_on_request_with_complex_type_element_form_default_missing/request.json
jsdevel: Are there tests that cover RPC style web services?  I feel the approach we've been taking with additional features / fixes is to add new tests with the goal of leaving old ones in tact.  That way we ensure no breaking changes for existing users.

aleung: No. They don't cover RPC style. Because `elementFormDefault` doesn't make sense when WSDL is in rpc/encoded style. I explained in the first comment of this PR.

jsdevel: Gotcha.

jsdevel:
@aleung thank you for your persistence and flexibility!  I hope you understand that our goal is to make `node-soap` the best library for everyone.  Your efforts to maintain backwards compatibility are most definitely appreciated!

@herom I'm good if you are.  👍

herom:
just a last commit squashing session so that this PR has only one single commit and I'm all up for merging this grand piece of work @aleung @jsdevel :+1:

aleung:
@herom Recently Github has provided the feature to squash commits on merging a PR (https://github.com/blog/2141-squash-your-commits). Would you like to try it?

herom:
I did it :dancer: Thanks a ton @aleung :+1:

PR #824, opened 2016-04-29
added options object to cache hash

tylerhou:
Currently different options (like different authentication details) are being cached as long as they have the same URL. This is bad. This is a fix to make sure that the options object is also taken into consideration when caching objects.


tylerhou:
oops I'm bad i'll try again later

PR #825, opened 2016-05-05
`elementFormDefault ` test case refactoring

aleung:
Followup of #822 

What's changed:
- Rename 'RpcExample' to 'ExecFile', since the WSDL isn't in RPC style.
- Some naming change to follow convention
- Remove useless namespace definition in WSDL
- Break long lines so that easy to trace diff


jsdevel:
Thanks @aleung !

PR #787, opened 2016-01-20
Proposal for CDATA handling.

dskatz:
This offers a potential solution for submitting CDATA text nodes in a SOAP Request.  

When integrating with many Enterprise level third party APIs CDATA text nodes are often used.

PR #828, opened 2016-05-06
Fix issue from #787

aleung:


aleung:
Fix build failing after merging #787.

PR #830, opened 2016-05-09
Fix incorrect WSDL in CDATA test case.

aleung:
The WSDL of CDATA test case was invalid. It couldn't pass validation by SoapUI and https://www.wsdl-analyzer.com/. Major issues in the WSDL includes:

- Undefined namespace `s`
- Undefined `message` element in request message

Another mistake was in `request.json`. The object should not contain top level element name.


aleung:
@jsdevel @herom I'd suggest to add into contribution guideline: Any WSDL checked in should be passing validation by WSDL validator.

PR #831, opened 2016-05-09
Fix issue of reference element in another namespace.

aleung:
This PR adds a test case to verify the case that an element in another namespace is referenced in the schema (line 18 and 25 in soap.wsdl).

Explanation to why I modified an existing test case:

After my modification applied, the test case `addPets__force_namespaces` failed. Examined the expected [request.xml](https://github.com/vpulim/node-soap/blob/master/test/request-response-samples/addPets__force_namespaces/request.xml): elements `petList`, `pet` and `name` haven't namespace prefix, so schema `attributedFormDefault` should be `unqualified` rather than `qualified`. It seems to be an issue which happen to be fixed right now.


jsdevel:
Love seeing the new test cases!

PR #832, opened 2016-05-09
Make ursa an optional dependency.

aleung:
'ursa' is a binary module which is required when `WSSecurityCert` is used. For users who don't make use of X509 certificate it isn't required.

Introducing binary module brings extra requirement to the build environment. Some users report that unable to installed 'ursa'.

This PR makes 'ursa' an optional dependency. Fail to install it won't break other functionality of node-soap.

Update and restructure the README as well.


herom:
Your contribution is really amazing @aleung - thanks a ton man :+1: 

I guess we can live with the `optional` solution, but I do also think that we should be aware that this does not solve the problem of binary modules when working with `ursa`... so, I'm just referencing the matching `ursa` issue report https://github.com/quartzjer/ursa/issues/133 so we can probably be updated whenever they come up with a solution for Windows/`node-gyp`

PR #833, opened 2016-05-10
Optionally add Created to wssecurity header

forivall:
This PR introduces an option to disable inclusion of the "wsu:Created" element from the WSSecurity auth scheme.

The api that I'm accessing breaks when this is included, for some dumb reason.

If this isn't suitable to be merged, I'll be fine with just using a customized WSSecurity protocol.


herom:
Thanks a lot for your time and effort @forivall  - we appreciate it a lot :+1: 

Would you mind to also mention the newly introduced option in our README, so that it's clearly documented for all future users too? This would be awesome :dancers:

forivall:
Updated :wave:

herom:
Sorry for the late response - I was "out of internet" the last couple of days. Thanks a lot!

PR #834, opened 2016-05-11
Fix custom http client not being used when fetching related resources

forivall:
I found this issue when I was first trying to fix strictSSL issues when fetching the WSDL, but initially worked around it with `wsdl_options`. However, since I will need to rewrite some urls on the fly, I need the full custom client to work when fetching the wsdl and all related resources.

`_socketStream.js` can also be used to clean up `client-customHttp-test`, on which I based my test (which is why I put it in a separate file).

In summary: a 1 line fix, with 208 additional lines of test code. :beers:

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1124,6 +1124,9 @@ WSDL.prototype._initializeOptions = function (options) {
   // Allow any request headers to keep passing through
   this.options.wsdl_headers = options.wsdl_headers;
   this.options.wsdl_options = options.wsdl_options;
+  if (options.httpClient) {
+    this.options.httpClient = options.httpClient;
+  }
 
   var ignoreBaseNameSpaces = options ? options.ignoreBaseNameSpaces : null;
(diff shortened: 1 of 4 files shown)


lib/wsdl.js
```diff
@@ -1124,6 +1124,7 @@ WSDL.prototype._initializeOptions = function (options) {
   // Allow any request headers to keep passing through
   this.options.wsdl_headers = options.wsdl_headers;
   this.options.wsdl_options = options.wsdl_options;
+  if (options.httpClient) this.options.httpClient = options.httpClient;
```
herom: Could you please either use the "curly brace" form of the `if` condition? Personally I think it makes the code more readable and I hope that I get a `jscs` rule set in `soap` as soon as possible :+1:

forivall:
@herom updated! My personal preference is braces when the body of the statement is on the next line, but I'm fine with following your style. JSCS / eslint would be great in communicating your preferences :+1:

herom:
Thanks a lot for your time & effort @forivall - we appreciate it a lot :+1:

PR #836, opened 2016-05-12
Bumping ursa to 0.9.4

mwain:
ursa 0.9.4 has fixes for node 6 deprecation warnings and efforts to fix the issues with installing it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "xml-crypto": "~0.8.0"
   },
   "optionalDependencies": {
-    "ursa": "0.8.5 || >=0.9.3"
+    "ursa": "0.8.5 || >=0.9.4"
   },
   "repository": {
     "type": "git",


.travis.yml
```diff
@@ -6,6 +6,7 @@ node_js:
   - 0.10
   - 0.12
   - 4.0
+  - 6.0
```
herom: Please revert this change as it's not obvious why this would change with a version bump of a dependency -  feel free to open up a new PR in order to get tests for node.js 6.x :+1:

test/_socketStream.js
```diff
@@ -1,3 +1,4 @@
+'use strict';
```
herom: The same as above, please revert this change as it's not obvious why this would be changed in the same commit as a dependency version bump.

mwain:
@herom I have updated this PR to remove the other changes.

herom:
Thanks a lot :+1:

PR #837, opened 2016-05-13
Streams

forivall:
I need to write tests for this and document it, but from a quick run, it worked, and resulted in a 30-50% speed increase!

It's also opt-in, since otherwise it would be a breaking change: just pass `{stream: true}` to `createClient()` options. It's breaking in that it emits events differently and requires a `createStream` method on the `HttpClient`.

The diff is much easier to read with the `-w` option: https://github.com/vpulim/node-soap/pull/837/files?w=1


herom:
Awesome! :+1:  Can't wait to merge this :dancers:

jsdevel:
Is there a way to just make it always work?  Perhaps with some shims?  I feel if we can make internal modifications to improve performance without changing the actual API of the library then we should.

forivall:
@jsdevel Ideally, yes. I'll see once I start setting up tests. If I can improve things to make it not a breaking change, then it can be on by default. Otherwise, it'll have to wait for a major release. semver!

I'd also like to investigate if i can completely remove duplication everywhere, by using streams instead of buffers everywhere, and do some fancy stuff like keeping a max size buffer in case there is a parsing error and we do need to make the body available.

Right now, this is proof of concept, and the concept worked. It'll take more time to polish it.

jsdevel:
sounds _very_ promising!

jsdevel:
@forivall what's the status on this one?  Can we merge?

forivall:
@jsdevel Status hasn't changed since my last comment in May. Thanks for the reminder though, I'll see if I can get to this sometime within a week.

jsdevel:
That would be awesome if you could @forivall !  We'd really like to get this merged.

ZepAviator:
@forivall Have you tried streaming an extremely large SOAP response? I have to pull in about 50 megs of data that I'd like to just pipe directly into a file and parse it from there.

forivall:
Tests added. I just made it so that the client tests also run with the stream option. Note that when I ran code coverage, the MultiRef support wasn't covered, but it isn't tested on master either.

jsdevel:
Thanks @forivall !

levithomason:
@ZepAviator I'm striving for a solution to streaming large responses as well.  Did you find a solution for this?  From what I tell, the `{ stream: true }` option doesn't return the stream so there is no way to pipe it.

PR #841, opened 2016-05-25
Support array of certs for ClientSSLSecurity ca.

awong-dev:
In server configurations where there are problems with the server side
certificate chain, it is sometimes required that the client add multiple
certs into the trust repository. The lower-level https library supports
this transparently in the agentOptions but ClientSSLSecurity's type
detection of Buffer and String made it impossible to access directly.

This commit adds support for passing an array of certificates.

### DIFF
--- a/lib/security/ClientSSLSecurity.js
+++ b/lib/security/ClientSSLSecurity.js
@@ -10,7 +10,7 @@ var fs = require('fs')
  * @module ClientSSLSecurity
  * @param {Buffer|String}   key
  * @param {Buffer|String}   cert
- * @param {Buffer|String}   [ca]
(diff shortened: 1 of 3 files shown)


jsdevel:
Thanks!

bluefangs:
@awong-dev  - I know this is a bit old, I'd like to know how you pass multiple ca certs in the array format. Do I pass the path to the location of the ca certs in an array format? Or is it something else?

Eg> 
```
client.setSecurity(new soap.ClientSSLSecurity(
    '/path/to/key'
    , '/path/to/cert'
, ['/path/to/cert-dir/cert1.key', 'path/to/cert-dir/cert2.key']

    , {/*default request options*/}
  ));
```

PR #842, opened 2016-05-25
gracefully handle errors while parsing xml in xmlToObject and resume the parser with p.resume()

deepaknverma:
Trying to handle sax parser error but can’t find sax parser handling error event in wsdl which result in breaking of app. We can  gracefully handle errors and resume the parser with p.resume(), assign a listener to the error event in the catch block


jsdevel:
Thanks!

PR #846, opened 2016-05-26
XSD import in WSDL files and relative path (server creation) - resubmit

mbayfield:
As per pull request - #803
Added test cases + adding option to escapeXML as the object response.


jsdevel:
Thanks!

PR #851, opened 2016-06-10
Add nonce and soap:actor support for WSSecurity

raymondfeng:


raymondfeng:
@vpulim @jsdevel Any chance to review and merge the PR? We would love to have a release with this fix soon.

lib/security/WSSecurity.js
```diff
@@ -20,7 +21,17 @@ function WSSecurity(username, password, options) {
   }
 
   this._hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true;
+  /*jshint eqnull:true */
+  if (options.hasNonce != null) {
```
herom: wouldn't it be better to check for `hasOwnProperty`?

jsdevel:
LGTM.  Thoughts @herom ?

lib/security/WSSecurity.js
herom: why do we have to check for `!= null` for the `options` properties instead of using `hasOwnProperty()`?

raymondfeng: I use `options.hasNonce == null` to check if `hasNonce` is `undefined` or `null`. If we use `options.hasOwnProperty('hasNonce')`, it will return `true` when `hasNonce` is explicitly set to `undefined` or `null`. Either way is fine but it doesn't really matter for this case, IMO.

jsdevel: Sounds like it's good then.  Thanks @raymondfeng !

raymondfeng:
@jsdevel Thanks! I look forward to a new release in npm soon.

raymondfeng:
@herom @jsdevel Would you please publish a new version to npm? We're blocked.

jsdevel:
I'll try to get to this today.

jsdevel:
@raymondfeng sorry for the delay.  This was published with `v0.16.0`.

PR #852, opened 2016-06-13
fixes an issue that causes the module to break if no req 

abrarahmedbcg:
when using soap client against exchange 2010 , req variable is not set  causing the soap client to fail
check for not null or undefined before accessing the variable


jsdevel:
Thanks!

PR #854, opened 2016-06-23
Remove special handling of methods only taking a string paramter

andreaspsson:
Methods where the only argument is a string will get caught here
and because of that wrong xml message will be generated and posted
to the server.

The service I am trying to connect to as well as this site: https://www.wsdl-analyzer.com/
says that the expected payload now is generated, before it was not possible to pass a string argument to the client.[METHOD] function unless that was a full xml-formatted body. Calling the function with an object as argument does not generate the expected body, since the data then get's embedded into one more layers of xml-tags.

Please correct me if I'm wrong, maybe I use the library wrong, but could not get this to work in any other way.

### Changes
**lib/client.js** - Removed line 224 and 225 and that is my actual patch

**test/client-test.js** - Added unit test for the patch

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -221,8 +221,6 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
       assert.ok(!style || style === 'rpc', 'invalid message definition for document style binding');
       message = self.wsdl.objectToRpcXML(name, args, alias, ns,(input.name!=="element" ));
       (method.inputSoap === 'encoded') && (encoding = 'soap:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/" ');
-    } else if (typeof (args) === 'string') {
-    message = args;
   } else {
     assert.ok(!style || style === 'document', 'invalid message definition for rpc style binding');
     // pass `input.$lookupType` if `input.$type` could not be found
--- a/test/client-test.js
+++ b/test/client-test.js
@@ -695,4 +695,35 @@ describe('SOAP Client', function() {
 
   });
 
+  describe('Method invocation', function() {
+
+    it('shall generate correct payload for methods with string parameter', function(done) {
+      // Mock the http post function in order to easy be able to validate the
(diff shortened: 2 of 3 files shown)


andreaspsson:
Can someone give any hints to why the travis build fails?
The failing command does not work locally for me, neither on upstream/master or  my own fork

herom:
I just restarted the tests and now it's all green :smiley:

lib/client.js
```diff
@@ -217,12 +217,10 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
   if (self.security && self.security.addOptions)
     self.security.addOptions(options);
 
-    if ((style === 'rpc')&& ( ( input.parts || input.name==="element" ) || args === null) ) {
-      assert.ok(!style || style === 'rpc', 'invalid message definition for document style binding');
-      message = self.wsdl.objectToRpcXML(name, args, alias, ns,(input.name!=="element" ));
-      (method.inputSoap === 'encoded') && (encoding = 'soap:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/" ');
-    } else if (typeof (args) === 'string') {
-    message = args;
+  if ((style === 'rpc')&& ( ( input.parts || input.name==="element" ) || args === null) ) {
```
herom: could you please revert all formatting/intendation only changes in the files as otherwise it's really hard to tell what exactly was changed with your PR. thanks.

test/client-test.js
```diff
@@ -695,4 +695,35 @@ describe('SOAP Client', function() {
 
   });
 
+  describe('Method innovacion', function() {
```
herom: do you mean "invocation"?

andreaspsson:
@herom 
Removed the changes that were just formatting, now I can no longer successfully run unit tests locally anymore, but I hope they pass on travis.

jsdevel:
Thanks @andreaspsson !

PR #856, opened 2016-06-27
added tns prefix in SOAP's method name, the server requires it

MichaelBuen:
SOAP server doesn’t accept the SOAP call if tns colon prefix is not
present in SOAP’s method name

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -224,7 +224,7 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
   } else {
     assert.ok(!style || style === 'document', 'invalid message definition for rpc style binding');
     // pass `input.$lookupType` if `input.$type` could not be found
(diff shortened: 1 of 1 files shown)


phanikiranwtc:
i hope this.
This can be achieved with  below json object in soap.createClient function. As ignoredNamespaces has tns as the default value which will be ignored.
{
 ignoredNamespaces: {
 namespaces: ['x'],
 override: true
 }
}

i have been using this config to consume my soap services.

MichaelBuen:
Thanks @phanikiranwtc that works, I remember I did that too, but maybe I forgot to put `override: true.`

namespaces doesn't need to have an element in it, empty namespaces works too:
```

var options = {
  ignoredNamespaces: {
    namespaces: [],
    override: true
  }
};

```

PR #857, opened 2016-06-30
add missing # char to uri reference in wss security token

jg-linetco:
URI references to local elements via id should prefix the id with a #-sign


herom:
this issue was already solved with #850 - thanks :smiley:

PR #860, opened 2016-07-14
Fix method name discovery for RPC calls where the first XML element has some attributes.

rodrigovr:
The same logic was already in place for "Document" type request. So I just applied it to both cases (RPC and Document).


jsdevel:
Please add request response samples.

jsdevel:
closing due to inactivity

PR #862, opened 2016-07-18
Always submit valid XML from the client.

Wilfred:
Previously, if the client made a request that contained ampersands:

    {
      "DummyField1": "foo & bar",
    }

then we would submit invalid XML:

    <n:DummyField1>foo & bar</n:DummyField1>

The WSDL server I'm using (a Service-Now instance) then complains:

    <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/">
      <SOAP-ENV:Header/>
      <SOAP-ENV:Body>
        <SOAP-ENV:Fault>
          <faultcode>SOAP-ENV:Server</faultcode>
          <faultstring>Unable to parse SOAP document</faultstring>
          <detail>Error completing SOAP request</detail>
        </SOAP-ENV:Fault>
      </SOAP-ENV:Body>
    </SOAP-ENV:Envelope>

We now default to escaping XML, so a stray & can't break usage of node-soap. I've added a test for this.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1121,7 +1121,11 @@ WSDL.prototype._initializeOptions = function (options) {
 
   this.options.valueKey = options.valueKey || this.valueKey;
   this.options.xmlKey = options.xmlKey || this.xmlKey;
-  this.options.escapeXML = options.escapeXML || false;
+  if (options.escapeXML !== undefined) {
+    this.options.escapeXML = options.escapeXML;
+  } else {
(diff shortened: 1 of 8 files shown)


jsdevel:
Thanks!

Wilfred:
@jsdevel I'd like to depend on this code -- would you mind doing a release?

jsdevel:
Done @Wilfred.  Sorry for the wait!

PR #863, opened 2016-07-22
Fix broken options variable in client SOAP requests

justathoughtor2:
The options variable was not registering in client SOAP requests, regardless of its contents. I modified the client.js file to include some items from wsdl.js, and it automagically worked the way it was supposed to. I suspect there might be a broken link to the WSDL class in the Client class that caused this, but I cannot tell for certain. Perhaps someone with more experience with this library could investigate further?

Thanks!
Jesse

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -16,7 +16,6 @@ var HttpClient = require('./http'),
 
 var Client = function(wsdl, endpoint, options) {
   events.EventEmitter.call(this);
-
   options = options || {};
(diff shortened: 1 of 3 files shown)


justathoughtor2:
Odd. Not sure why this didn't build. It runs perfectly on my local machine... Hrmm...

jsdevel:
please add tests

justathoughtor2:
Added tests. All tests pass except for the first two. Odd.

PR #868, opened 2016-08-10
handle SOAP message without Envelope and Body

prasunsultania:
If the incoming SOAP message is without Envelope and body it throws a weird error:

TypeError: Cannot convert undefined or null to object
    at Function.keys (native)
    at Server._process (/Users/prasunsultania/git/soaptests/node_modules/soap/lib/server.js:246:27)
    at IncomingMessage.
<anonymous> (/Users/prasunsultania/git/soaptests/node_modules/soap/lib/server.js:146:14)
    at emitNone (events.js:67:13)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:921:12)
    at nextTickCallbackWith2Args (node.js:442:9)
    at process._tickCallback (node.js:356:17)

This fix does two tasks (1) tries to throw explicit error than running into type errors (2) still allows to return the response SOAP message even if the Envelope and Body are not specified in the incoming request. Please see the sample request and response added in this pull.


jsdevel:
Can you address the failing build issue @prasunsultania ?

prasunsultania:
Looking into it, allow me some time

prasunsultania:
Fixed these tests, but, there is an another problem:

Without Envelope and Body technically the message is NOT SOAP, as per [W3Org Specs](https://www.w3.org/TR/2007/REC-soap12-part0-20070427/#L1165) which says - 

> The SOAP body is the mandatory element within the SOAP env:Envelope, which implies that this is where the main end-to-end information conveyed in a SOAP message must be carried

.

This causes the HTML tests to fail, as HTML test is passing the error conditions assuming the errors are thrown when Envelope and Body are not defined.

I will open a separate PR for throwing only more explicit error than handling it.

PR #871, opened 2016-08-12
On error, call callback with error instead of message

r24y:
As it stands, the callback for `WSDL` is sometimes called with a string in the error position. This error arises from the line `self.callback(e.message)`. This strips useful information (such as the stack trace) before sending the error back to the consumer, as well as violates JS best practices by throwing a string instead of an `Error`.

This PR corrects the issue by calling the callback with the error.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1022,7 +1022,7 @@ var WSDL = function(definition, uri, options) {
     try {
       fromFunc.call(self, definition);
     } catch (e) {
-      return self.callback(e.message);
+      return self.callback(e);
     }
 
     self.processIncludes(function(err) {


jsdevel:
Hmm.  Not saying I disagree, but this is somewhat of a breaking change wouldn't you say?  Is there a way to add a test to assert an `Error` is returned?

r24y:
If someone's calling `err.substr()` or something like that, then it would be a breaking change. That seems unlikely to me, but of course ultimately it's your call.

Assuming you're okay with the change, I'd be happy to add a test ensuring that an `Error` gets returned. Let me know if I should add it to this PR, or if I ought to make a new one and check what `err` is for all the callbacks.

jsdevel:
closing due to inactivity

PR #872, opened 2016-08-12
express server support

prasunsultania:
supporting express server so, that it is easier to use node-soap in an existing express server without affecting other routes.

It supports middleware for body parsing like body-parser as well.

Please see the tests included in commit.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -41,27 +41,40 @@ var Server = function(server, path, services, wsdl, options) {
   if (path[path.length - 1] !== '/')
     path += '/';
   wsdl.onReady(function(err) {
-    var listeners = server.listeners('request').slice();
-
-    server.removeAllListeners('request');
-    server.addListener('request', function(req, res) {
-      if (typeof self.authorizeConnection === 'function') {
-        if (!self.authorizeConnection(req)) {
-          res.end();
-          return;
+    if(typeof server.route === 'function' && typeof server.use === 'function'){
+      //handle only the required URL path for express server
+      server.route(path).all(function(req, res, next){
+        if (typeof self.authorizeConnection === 'function') {
+          if (!self.authorizeConnection(req)) {
+            res.end();
+            return;
+          }
         }
-      }
-      var reqPath = url.parse(req.url).pathname;
-      if (reqPath[reqPath.length - 1] !== '/')
-        reqPath += '/';
-      if (path === reqPath) {
         self._requestListener(req, res);
-      } else {
-        for (var i = 0, len = listeners.length; i < len; i++) {
-          listeners[i].call(this, req, res);
+      });
+    } else {
+      var listeners = server.listeners('request').slice();
+      server.removeAllListeners('request');
+      server.addListener('request', function(req, res) {
+        if (typeof self.authorizeConnection === 'function') {
+          if (!self.authorizeConnection(req)) {
+            res.end();
+            return;
+          }
         }
-      }
-    });
+        var reqPath = url.parse(req.url).pathname;
+        if (reqPath[reqPath.length - 1] !== '/'){
+          reqPath += '/';
+        }
+        if (path === reqPath) {
+          self._requestListener(req, res);
+        } else {
+          for (var i = 0, len = listeners.length; i < len; i++) {
+            listeners[i].call(this, req, res);
+          }
+        }
+      });
+    }
   });
 
   this._initializeOptions(options);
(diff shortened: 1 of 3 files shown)


jsdevel:
Looks good to me.  I'd like @herom to weigh in as well.

prasunsultania:
@jsdevel / @herom Any update on this ? We want to use this feature in our ongoing project.

lib/server.js
```diff
@@ -41,27 +41,39 @@ var Server = function(server, path, services, wsdl, options) {
...
   wsdl.onReady(function(err) {
-    var listeners = server.listeners('request').slice();
-
-    server.removeAllListeners('request');
-    server.addListener('request', function(req, res) {
-      if (typeof self.authorizeConnection === 'function') {
-        if (!self.authorizeConnection(req)) {
-          res.end();
-          return;
+    if(typeof server.route === 'function' && typeof server.use === 'function'){
```
herom: is there any other possibility to check if the passed in server instance is an `express` app or is this check reliable from at least 4.x upwards?

lib/server.js
```diff
@@ -41,27 +41,39 @@ var Server = function(server, path, services, wsdl, options) {
...
+        if (typeof self.authorizeConnection === 'function') {
+          if (!self.authorizeConnection(req)) {
+            res.end();
+            return;
+          }
         }
-      }
-    });
+        var reqPath = url.parse(req.url).pathname;
+        if (reqPath[reqPath.length - 1] !== '/')
```
herom: please use curly braces (`{}`) for all your `if()` statements

lib/server.js
```diff
@@ -100,6 +112,49 @@ Server.prototype._initializeOptions = function(options) {
...
+        res.statusCode = statusCode;
+      }
+      res.write(result);
+      res.end();
+      if (typeof self.log === 'function') {
+        self.log("replied", result);
+      }
+    });
+  }
+  catch (err) {
```
herom: please line the `catch` along with the closing brace of the `try` statement --> `} catch (err) {`

lib/server.js
```diff
@@ -100,6 +112,49 @@ Server.prototype._initializeOptions = function(options) {
...
+    if (err.Fault !== undefined) {
+      return self._sendError(err.Fault, function(result, statusCode) {
+        if(statusCode) {
+          res.statusCode = statusCode || 500;
```
herom: either we don't need the `if` around the assignment or the `|| 500` as we won't hit it if `statusCode` does not exist.

test/express-server-test.js
```diff
@@ -0,0 +1,166 @@
...
+            }
+          }
+        }
+      }
+    };
+
+    expressServer = express();
+    server = expressServer.listen(51515, function(){
+      var soapServer = soap.listen(expressServer, '/SayHello', service, wsdl);
+      url =
```
herom: please align the assignment into the same line.

test/express-server-test.js
```diff
@@ -0,0 +1,166 @@
...
+        'http://' + server.address().address + ':' + server.address().port;
+
+      if (server.address().address === '0.0.0.0' || server.address().address === '::') {
+        url =
```
herom: same here

lib/server.js
prasunsultania: `instanceof` wont work in this case. This will work 4x onwards as `use` and `route` are most commonly used express methods. 

Current check gives us an another advantage - we have `express` as dev dependency only :)

lib/server.js
prasunsultania: this snippet was taken from previous code, so, I wanted to leave it as it was to keep reviews simpler. But, will correct and squash the commits.

lib/server.js
prasunsultania: this snippet was taken from previous code, so, I wanted to leave it as it was to keep reviews simpler. But, will correct and squash the commits.

lib/server.js
prasunsultania: this snippet was taken from previous code, so, I wanted to leave it as it was to keep reviews simpler. But, will correct and squash the commits.

test/express-server-test.js
prasunsultania: I will correct and squash

test/express-server-test.js
prasunsultania: I will correct and squash

prasunsultania:
@herom have made recommended changes. Please review.

herom:
Thanks a ton @prasunsultania - LGTM :+1: 

One last thing: could you please open a PR and update our `README.md` so that we have it documented, that `soap` is able to handle an `express` server instance as well? :smile_cat:

prasunsultania:
I will open a separate PR after updating docs.

PR #876, opened 2016-09-08
Add option for disabling the WSDL cache

warseph:
Currently the soapClient is caching WSDLs (this seems to be an undocumented/unconfigurable feature).
The proposed change refactors that code a bit to encapsulate that functionality and adds an option to disable it completely.
The main reason this needs to be addressed are tests. If the cache is enabled, tests using a HTTP mocking solutions, such as [node-nock](https://github.com/node-nock/nock) will end up being dependent tests, as only the first case will call the WSDL, and the rest will behave differently.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -87,6 +87,7 @@ If you're in need of support we encourage you to join us and other `node-soap` u
       });
   });
 ```
+This client has a built in WSDL cache. You can use the `disableCache` option to disable it.
 
 #### Options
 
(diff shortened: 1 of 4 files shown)


jsdevel:
Great addition.  Thanks!

warseph:
Hi @jsdevel . Any idea when this will be published to npm?
Thanks!!

jsdevel:
should be able to get a bump out today or tomorrow

ezgcmr:
Ezgi Camur

mbrevda:
Would you happen to have an example of how to mock soap requests?

warseph:
Hi @mbrevda I used https://github.com/node-nock/nock to mock requests.
A coworker made this package to simplify using it, it might help you: https://github.com/comparaonline/nock-utils

PR #877, opened 2016-09-15
Avoid matching <x:Envelope> tags inside comments

ranhalprin:
I have a specific case where I fail to use node-soap because the wsdl service is sending me an xml with a comment that includes an example of an Envelope node:

```
<RealNode>
...
<!--
  Use this format for something: 
    <ins:Envelope> 
       Bla Bla Bla 
    </ins:Envelope>...
-->
</RealNode>

```
The current code picks this piece of the comment and tries to work with ```<ins:Envelope> Bla Bla Bla </ins:Envelope>``` as the body of the request.

This PR solves this by matching on the body after removing comments.


herom:
thanks a lot @ranhalprin - could you please add tests as outlined in our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) so that we can asure, that your code does not get overridden/ripped out accidentially? :+1:

ranhalprin:
Thanks @herom, added a test that fails without my change.

herom:
Awesome! Thanks a ton :dancers:

cedricrey:
Hello,
I have a content with "" inside my soap response, and with the 
`var match =
      body.replace(//, "").match(/(?:<\?[^?]*\?>[\s]*)?<([^:]*):Envelope([\S\s]*)<\/\1:Envelope>/i);
    if (match) {
      body = match[0];`
insctruction, the first comment is deleted, beceause the body is valid for the second regular expression.
Is it normal to remove any first "" in the response even if it doesn't contain <Envelope> code ?
In order to not remove my necessary comment, I surchage the handleResponse, but I'm not sure that's the good way, and I prefer to let you know this behavior.

And the other way, thank you for all your work, soap library is very usefull !!!

herom:
hey @cedricrey,
thanks a lot for your kind words 😺 

it would be awesome if you could show us what you'd expect our library to output if it hits your case with a PR 👍

cedricrey:
Hello @herom, 
I'm sorry I've never post on github, I don't know how to do :-)
I don't know what is a PR.
My case is : I create a soap client, and when I request with it, the response has HTML content in it, something like that : 
`<Envelope><data><content><source><[!CDATA[<!--some necessary comment --><html>....</html>]]></source></content></data></Envelope>`
And when I get the response from the client, the first html comment (`<!--some necessary comment -->)` was deleted from the content. 
After investigation, I found that was comming from the HttpClient.prototype.handleResponse, particulary the `.replace(/<!--[\s\S]*?-->/, "")`. I found that code were added for this pull request.
Is this delete expected ? I may be wrong and have forgot something in the use of the library.

PR #878, opened 2016-09-16
Fixed request-response-samples-test, which didn't run tests with only request.json

pxpeterxu:
While writing a new test that only applies to the requests, I noticed that the tests never actually run to compare the outputs of building a XML document from request.json with request.xml.

**Test case:**

1. Replace test/request-response-samples/Dummy__should_work_with_imported_schemas/request.xml with a file that just says "blah"
2. Run `mocha test\request-response-samples-test.js`

**Expected result**: test fails: `(1) should work with imported schemas`
**Actual result**: test passes: `√ should work with imported schemas`

The fix involves moving the doneHandler() to after the check so that we don't skip this check when we don't have a responseToSend. Since many of the `request.xml` files have not been really tested, I've also updated them (mostly by adding the <?xml> tag at the start).


jsdevel:
Thanks for the improvement!  Always a plus when contributors improve the tests.

PR #879, opened 2016-09-16
Added useEmptyTag option to wsdlOptions to prefer <Tag /> format over <Tag></Tag>

pxpeterxu:
One of the web services I was working with for some reason required that tags without a body be `<Tag />` (empty, self-closing) rather than `<Tag />`. The [XML spec](https://www.w3.org/TR/REC-xml/#NT-content) makes no distinction between the two, but I had to work with the service.

**Change**: added `useEmptyTag` option to wsdlOptions that will prefer `<Tag />` to `<Tag></Tag>` if the tag has no contents.

**Accompanying changes**:
- README.md documentation on extra switch
- Two tests:
  1. should_use_empty_tags_when_directed: checks that when useEmptyTag is set, empty tags are preferred
  2. should_use_empty_tags_only_when_directed: checks that when useEmptyTag is not set, the current behavior does not change


jsdevel:
For whatever reason, one of the builds failed.  I kicked it off again.

Assuming that passes, I think this is a great change.   Thoughts @herom ?

Thanks for adding tests!

pxpeterxu:
Seems to have failed twice (I also tried the same to kick it off again); I'll take a look, but it's with a really old node so might take me a bit to get the environment set up.

jsdevel:
Looks like an issue with `deepEqual`.  `v0.10` and `v0.12` will be deprecated shortly.  Perhaps we can remove them from `.travis.yml` in a separate PR?

herom:
@jsdevel I'm all up for dropping `v0.10` and `v0.12` support in tests as we should move on with the library and don't be too picky about compatibility with officially deprecated versions of the system.

@pxpeterxu thanks a lot for your PR - this really seems like a valuable addition to our lib :+1:

jsdevel:
closing due to inactivity

PR #883, opened 2016-10-05
Update soap stub example

lewisp6:
Clarifying confusing part of the example with the error stub. 

Respond with error should use the createErroringStub method so that the client returns an error rather than using createRespondingStub which returns a non-error response to the callback.


jsdevel:
Thanks @lewisp6 !

PR #888, opened 2016-10-15
Removes wsdl.onReady from the Server constructor

jacargentina:
server.route() must be called sync for the module to work fine when passed server is a instance of express.Router
### Use case

When trying to integrate the module for serving SOAP requests on an already existing express setup, i needed to mount it not on an express app instance, but a Router instance. 
### Why the current code does not work in my case

It seems that the onReady callback makes the `server.route` call too late in this case
### Solution

Dont use onReady at all. Anyway, the wsdl would be full loaded right away.
### Open to make it better

If somebody points me on the need for the onReady call, maybe we can think a better solution than mine.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -274,8 +274,9 @@ Server.prototype._process = function (input, req, callback) {
   }
 
   try {
+    var firstXmlElement = (Object.keys(body)[0] === 'attributes' ? Object.keys(body)[1] : Object.keys(body)[0]);
     if (binding.style === 'rpc') {
-      methodName = Object.keys(body)[0];
+      methodName = firstXmlElement;
 
       self.emit('request', obj, methodName);
       if (headers)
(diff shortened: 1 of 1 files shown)


apkbox:
There is another way to inject soap handling in the right place. I was having trouble with this as usually you want to handle 404, which should be the last route.
Consider an example:

```js
var app = require('express')();
var ws = require('soap');
soap.listen(app, ...blablabla...);
app.use(my404handler);
```

Now, just create a router, insert it in the right location and use router instance instead of server.
The exploit here is that server is in essence a router and happens to have the same interface.
 ```js
var app = require('express')();
var ws = require('soap');
var wsrouter = express.Router();
app.use(wsrouter);  // The router inserted into the right spot (before 404 handler)
soap.listen(wsrouter, ...blablabla...);
app.use(my404handler);
```

jsdevel:
@apkbox will this change be backwards compatible with your proposed workaround?

herom:
@jsdevel as far as I have understood @apkbox 's answer, this works without any changes to the actual code-base, as an `express.Router` instance has the same interface as an `express` instance.

jsdevel:
> Perhaps we should mention this solution in our README.md?

I think we should.  Given that, is this PR even needed?

herom:
> Given that, is this PR even needed?
I don't think so, no.

Would you mind on raising a PR to update our `README.md` with your solution @apkbox ?

PR #889, opened 2016-10-18
Fix body may contain 'attributes' element

liudonghua123:
Fix body may contain 'attributes' element if request body contain extra attributes like <soap:Body soap:encodingStyle=http://schemas.xmlsoap.org/soap/encoding/>


jsdevel:
@liudonghua123 please add tests.

PR #893, opened 2016-10-18
add support for client ssl with wsse security

wellszhane:
Combine Client SSL add options to header and WS security together.


jsdevel:
Please add tests if you can @wellszhane .

leeuwd:
+1

Multiple questions for this on Gitter. Same here.

PR #894, opened 2016-10-25
Ensure that supplied request-object is passed through

ath88:
The request-parameter from createClient is lost for later requests in the WSDL-prototype. This PR fixes that problem.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1114,6 +1114,11 @@ WSDL.prototype._initializeOptions = function (options) {
     this.options.httpClient = options.httpClient;
   }
 
+  // The supplied request-object should be passed through
+  if (options.request) {
+    this.options.request = options.request;
+  }
+
   var ignoreBaseNameSpaces = options ? options.ignoreBaseNameSpaces : null;
   if (ignoreBaseNameSpaces !== null && typeof ignoreBaseNameSpaces !== 'undefined') {
     this.options.ignoreBaseNameSpaces = ignoreBaseNameSpaces;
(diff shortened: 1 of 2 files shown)


jsdevel:
@ath88 is there a way this can be tested?

ath88:
It can be tested, yes. 

I did this by supplying a request-object with a custom header added in [`request.defaults`](https://www.npmjs.com/package/request#requestdefaultsoptions). Using the [`request-debug`-module](https://github.com/request/request-debug), I can inspect the requests being made. On the first request to the actualy WSDL-file, the custom header is included, while the second request (to eg. a schemaLocation) lacks the custom header.

I stumbled across this when trying to supply a request-object with a specific certificate and key. The first request would succeed, while the second failed, since the key and certificate was not used in the second request.

With my changes, the request-object (along with new defaults) is retained.

I could probably write a test for this, but it would be difficult to make it fit the current mocha-scheme.

jsdevel:
can you add a test @ath88 ?

ath88:
I added an extra assert to an existing test. The new assert would fail without to the first commit I made.

ath88:
Incidentally, it wasn't as hard as I had thought. :)

jsdevel:
Thanks!

ath88:
@jsdevel What is your process regarding publishing new versions? :) I ask this because I'd like to know if we should publish this fix internally, or wait for it to be published.

jsdevel:
@ath88 normally @herom or myself will publish a new version.  Lately I've been swamped so haven't had a chance to publish.  I'll try to tonight.

ath88:
Thanks for responding so quickly! *hat tip*

PR #897, opened 2016-11-16
add body parameter

rflament:
Sometimes the code that invokes the webservice needs to access the raw body to handle deserialization itself (for example because the date format is not correct so the deserialization does not work).

When it's the case, it does not seem possible with the current stub to stub the raw soap response.
This pull requests fixes this.


jsdevel:
Thanks @rflament !

PR #840, opened 2016-05-17
Don't generate invalid dates

mredbishop:
Don't generate invalid dates when the schema uses a date time format that is incompatible with the javascript Date object constructor such as '1987-01-01-05:00'

PR #898, opened 2016-11-17
Don't generate invalid dates - same as 840

rflament:
I tried to finish the pull request about the date format error : https://github.com/vpulim/node-soap/pull/840 

It uses moment.js.

Most of the work had been done by mredbishop , I just made small changes to change attributes name and reduce the size of the files used in the test.


package.json
```diff
@@ -13,6 +13,8 @@
     "ejs": "~2.3.4",
     "finalhandler": "^0.5.0",
     "lodash": "^3.10.1",
+    "moment": "~2.13.0",
```
jsdevel: Not sure how crazy I am about adding `moment` as a dependency.  Is there a way to just pass in a string?

jsdevel: Thoughts @herom ?

rflament: Another way would be to be able to pass a function to deserialize specific fields by name or by type. Could be in options.

Then the code would look like :


if (customDeserializerExists(name)) {
value = customDeserializerExists(name)(text);
}
else {
 
// current code
if (name === 'int' || name === 'integer') {
      value = parseInt(text, 10);
    } else if (name === 'bool' || name === 'boolean') {
      value = text.toLowerCase() === 'true' || text === '1';
...
...

}

jsdevel: I think something like that would be better

rflament: I'll open another pull request.

rflament: https://github.com/vpulim/node-soap/pull/901

PR #899, opened 2016-11-29
Ignore VSCode files from git commit

aleung:


jsdevel:
Can you also remove the vscode files?

aleung:
@jsdevel Which files do you refer to? There is no vscode file inside git repo at present. I send this PR because I'm using VSCode and I want to avoid VSCode auto generated files being added into my upcoming PRs. I believe it benefits other contributors who use VSCode as well.

jsdevel:
I see.  I thought some were checked in.  Thanks!

PR #908, opened 2016-12-19
Update the version of ejs dependency to ~2.5.5

countergram:
This is due to the security vulns described here https://snyk.io/blog/fixing-ejs-rce-vuln

I ran the mocha suite and everything seemed OK - there was a nondeterministic result from some timing test but I don't believe it's related, and I couldn't run full `npm test` because of some lint errors in an unrelated file.


jsdevel:
thanks!

PR #910, opened 2017-01-05
Add support for attributes in root elements

jonathanarnault:
Those changes allow to add attributes to root elements of the soap request. For example:
```js
client.foo({
  foo: {
    attributes: { bar: 'true' },
    $value: 'my-value'
  }
}, function(err, result) {

});
```

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1572,7 +1572,14 @@ WSDL.prototype.objectToRpcXML = function(name, params, nsPrefix, nsURI,isParts) 
     if (key !== nsAttrName) {
       var value = params[key];
       var prefixedKey = (isParts ? '' : nsPrefix) + key;
-      parts.push(['<', prefixedKey, '>'].join(''));
+      var attributes = [];
+      if (typeof value === 'object' && value.hasOwnProperty(this.options.attributesKey)) {
+        var attrs = value[this.options.attributesKey];
+        for (var n in attrs) {
+          attributes.push(' ' + n + '=' + '"' + attrs[n] + '"');
+        }
+      }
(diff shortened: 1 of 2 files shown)


jsdevel:
I'm wondering if request/response samples would be a better test for this.  To me the test isn't very clear.  Thoughts @herom ?

herom:
I'd also ask for a request/response sample as the test itself is somewhat unclear. The `bar1:test="attr"` is not an element on it's own but rather added to the `<bar1:paymentRq>` "root" element. This is not the output I'd expect by reading the title of the PR.

mtgto:
Hi, I also have a problem (It will be solved by this PR).
I currently use node-soap to access 3rd-party soap api, and have trouble to create such a request.

```xml
<soap:Envelope>
<soap:Header>
  <Action>ScheduleSearchEvents</Action>
</soap:Header>
<soap:Body>
  <tns:ScheduleSearchEvents>
    <parameters text="Test">
    </parameters>
  </tns:ScheduleSearchEvents>
</soap:Body>
</soap:Envelope>
```
(xml snippet by searching my schedule by text "Test").
But this code don't generate XML with attributes:

```js
client.ScheduleSearchEvents({parameters: {attributes: {text: "Test"}}}, (err, result) => {
  ...
}
```

```xml
<soap:Envelope>
<soap:Header>
  <Action>ScheduleSearchEvents</Action>
</soap:Header>
<soap:Body>
  <tns:ScheduleSearchEvents>
    <parameters>
    </parameters>
  </tns:ScheduleSearchEvents>
</soap:Body>
</soap:Envelope>
```
(contains no text "Test").

And this PR work for me to generate XML with attributes.

Is there any way to add attributes to `parameters` node without this PR solution?
(Or should change `root` to `first children of root` ?)

PR #913, opened 2017-01-11
FIX deprecated node-uuid@1.4.7

kuozumijp:
travis.yml - change nodejs version 6.9.4(LTS) and 7x.
package.json - change coverall command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 notifications:
   email: false
 node_js:
+  - 0.10
+  - 0.12
+  - 4.0
   - 6.9.4
   - 7.4.0
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ language: node_js
 notifications:
   email: false
 node_js:
-  - 0.10
-  - 0.12
-  - 4.0
-  - iojs
+  - 6.9.4
+  - 7.4.0
 env:
   - CXX=g++-4.8
 addons:
--- a/lib/client.js
+++ b/lib/client.js
@@ -14,7 +14,7 @@ var HttpClient = require('./http'),
   findPrefix = require('./utils').findPrefix,
   _ = require('lodash'),
   concatStream = require('concat-stream'),
-  uuid = require('node-uuid');
+  uuid = require('uuid');
 
 var Client = function(wsdl, endpoint, options) {
   events.EventEmitter.call(this);
--- a/lib/security/WSSecurityCert.js
+++ b/lib/security/WSSecurityCert.js
@@ -6,7 +6,7 @@ var fs = require('fs');
 var path = require('path');
 var ejs = require('ejs');
 var SignedXml = require('xml-crypto').SignedXml;
-var uuid = require('node-uuid');
+var uuid = require('uuid');
 var wsseSecurityHeaderTemplate = ejs.compile(fs.readFileSync(path.join(__dirname, 'templates', 'wsse-security-header.ejs')).toString());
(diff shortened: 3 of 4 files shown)


herom:
Regarding the move from `node-uuid` to `uuid`, see the "official" discussion here: https://github.com/kelektiv/node-uuid/issues/142

Regarding the removal of  `0.10`, `0.12`, `4.0` versions from the Travis-CI config: If we'll move forward and "just" support the latest LTS versions I'm good to go if @jsdevel is also OK with it.

Regarding the removal of `io.js` from the Travis-CI config: 👍

jsdevel:
I think we should target April for v0.12 removal personally.  There could be quite a few users out there still using v0.12 that aren't able to upgrade in a timely manner.

bjrmatos:
i'm still using `v0.10.x` 😆 , if supporting other versions of node doesn't change anything in the codebase right now, why dropping support for those versions? 

also, supporting only `6.x.x` or `7.x.x` is like living in the edge.

kuozumijp:
i think don't supported npm at v0.10@travis.
but test command calld "npm run test" what for?

console log:https://travis-ci.org/vpulim/node-soap/jobs/190924124

```
$ node --version
v0.10.48
$ npm --version
2.15.1
$ nvm --version
0.32.0

...

npm WARN npm npm does not support Node.js v0.10.48
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
```

kuozumijp:
test is successful\@travis.
so version returned nodejs v0.10 and more.

herom:
LGTM - what about you @jsdevel?

jsdevel:
Thanks @kuozumijp !

felipegarcia92:
Hello everyone!

Reading this because I'm trying to use https://github.com/ChrisAlvares/node-adwords, where I submitted a new issue, and this soap library is throwing the following error:

```
node_modules/soap/lib/client.js:275
  var eid = options.exchangeId || uuid.v4();
                                       ^
TypeError: uuid.v4 is not a function
```

I tried changing manually the import on the `client.js` file of this library from
`uuid = require('uuid');` to `uuidv4 = require('uuid/v4');`
and the call of the function from
`uuid.v4();` to `uuidv4();`

And it works. Do you see anything wrong with versions listed on `node-adwords`'s package.json, or is this an issue for this library?

Thanks in advance!

herom:
@felipegarcia92 thanks a lot - this is *definitely* a bug! I wonder why we had not a single test for this behaviour which could have saved us from hitting this nasty thing.

It seems that up to version `3.0.0` of the `uuid` package it was really just a drop-in replace for the deprecated `node-uuid` package and the calls to `uuid.v4()` were perfectly valid *but* they decided to [split them into separeate "modules"](https://github.com/kelektiv/node-uuid/commit/681bddd6f4cf3abb8c1af93c30d035b695cbe756) for `3.0.1` (which is - in my opinion - not just a patch relase...) and that's were all hell broke loose 😨 

Would you mind to open a PR and provide a fix for this (I'd also vote for locking the version in our `package.json` to `3.1.0`)?

/ping @jsdevel

felipegarcia92:
@herom Thanks for responding and for me a simple `rm -fr node_modules && npm install` did the job. But yea, `uuid`'s structure has changed.

herom:
@felipegarcia92 I upgraded our dependency version of `uuid` to `3.1.0` and use their new "module system" in our code (see https://github.com/vpulim/node-soap/commit/24180416be9c2b8242a4f472b932e9c560a36e99) 

Thanks for notifying us about this issue 👍

felipegarcia92:
@herom awesome! thanks man! you're welcome

PR #915, opened 2017-01-20
Prevent Maximum callstack exceeded when parsing some recursive wsdl

meriadec:
Hi, thanks for this module.

For some reasons, the recursive search algorithm was infinite looping with the wsdl file that I used. This commit keep a track on already traversed nodes and prevent traverse them again. It also speed up objectToXml *a bit*.

I added a test and the problematic wsdl file. I'm pretty new to SOAP so it may not be perfect, but I didn't achieve to reproduce the problem with the existing `recursive.wsdl`. 😄


meriadec:
Oh, the soap endpoint described in my wsdl file has an IP whitelist.
I'll update it :grin:

PR #918, opened 2017-02-10
Add the ability to submit already-formed XML strings to the SOAP client API

jcald1:
Sometimes a user already has an XML string that they want to submit to a SOAP service. This pull request introduces that functionality.
This pull request also includes some improvements to the Readme.md client documentation for the client.MyFunction "args", both for args being an XML string and a JSON representation of the XML to be generated, the latter of which is existing functionality.  The existing functionality part could use some more documentation.  For example, there was one unit test that uses the "xsi_type" attribute and and that information appears to be split into the xsi:type and xsi:xmlns attributes in the generated XML. 
I also found a small bug where an element that has both a value ($value) and a child element, the child element is not generated even though that should be valid XML.  That bug is not addrsesed here.  I only found one unit test for the existing functionality where args is different from "{}".  I added two for the new functionality.


jcald1:
@jsdevel, I pushed a fix.

lib/wsdl.js~
```diff
@@ -0,0 +1,2198 @@
+/*
```
herom: please remove this file (it seems like this is a backup file from your Editor/IDE?)

jcald1: @herom, done.

Readme.md
```diff
@@ -371,6 +371,31 @@ An instance of `Client` is passed to the `soap.createClient` callback.  It is us
...
+      </ChildElementNoNS>
+  </nsPrefixInXsd:MyMessage>
+```
+The namespaces and namespace prefixes should already be pre-populated from the WSDL/XSDs in the MyMessage and/or SOAP Envelope elements.  If an element in args contains no namespace prefix, the target namespace of MyMessage is used.  A prefix may or may not be added, depending on how the XSD is setup.
+Currently, elements may not contain both child elements and a text value.
+
+##### Example with XML String as an argument
+You may pass in a fully-formed XML string instead the individual elements and attributes that make up the XML.  The XML string should not contain an XML declaration (e.g., `<?xml version="1.0" encoding="UTF-8"?>`) or a document type declaration (e.g., `<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Frameset//EN" "http://www.w3.org/TR/html4/frameset.dtd">`).
+``` javascript
+  {_xml: "<nsPrefixInXsd:MyMessage xmlns:nsPrefixInXsd="http://www.example.com/v1/ns1">
```
herom: Please use simple qoutes(`'`) for defining the `xmlns:nsPrefixInXsd='http://www.example.com/v1/ns1'` or escape the double quotes `"` as otherwise the markup is broken (if you look at this file in the github diff-view, you see that the closing culry brace and everything after is marked "red", which basically means "bad markup")

herom:
@jcald1 would you mind to fix the error (either escaping the double quotes within the documentation or replacing them with single quotes) and rebase?

this is the only thing which holds us back from merging this PR and I think this is a valuable addition to your library so I would appreciate this a lot 👍 😺

jcald1:
@herom, done.  I improved the documentation as well.  We may want to document/test things even more .  For example, the Readme.md doesn't state how to specify which element from the WSDL should be populated right under the SOAP Body, when there are multiple elements in the WSDL and the args is JSON.  It seems to be taking the very first element.

History.md
```diff
@@ -1,3 +1,30 @@
+0.19.1 / 2017-05-30
```
jsdevel: revert this

jcald1: @jsdevel, Done.

jsdevel:
@jcald1 something is wrong with your branch

jcald1:
@jsdevel, I previously reverted commit ad4b85a81c0c46c9f52882940e14e18f8afabd03 (Release v0.19.1).  Is that what you had asked for?   What do you see that's wrong with the branch?

jsdevel:
There shouldn't be any changes to `package.json`.  I didn't want you to revert the Release commit.  I wanted you to revert all of _your_ changes to `package.json`.

jcald1:
@jsdevel, Fixed.

jsdevel:
@jcald1 something is still wrong with your branch.  I should only see commits from you on this branch.  Right now I'm seeing commits from myself and others.

jcald1:
@jsdevel, Fixed (hopefully).

jsdevel:
I think I'm good with this change if you are @herom, other than the need to squash during merge.

herom:
LGTM 👍

jcald1:
@herom, #jsdevel, I did an npm install on the latest version and I see that lib/wsdl.js has 2213 lines but the ones in github has 2217 lines, but both have version 0.19.2 in the package.json files.

herom:
As you can see from a look at the latest release, I'm afraid your PR didn't make it into `0.19.2`

image
image

Sorry for that, but it was me who was too late here 😞

PR #919, opened 2017-02-10
Added support for "RequestMsg" and "ResponseMsg" naming suffixes. Thi…

jarredkeuch:
Added support for "RequestMsg" and "ResponseMsg" naming suffixes. This is a convention for IBM WPM servers.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1256,15 +1256,12 @@ WSDL.prototype.xmlToObject = function(xml, callback) {
         // Determine if this is request or response
         var isInput = false;
         var isOutput = false;
-        if ((/Response$/).test(name)) {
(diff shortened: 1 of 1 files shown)


jsdevel:
@jarredkeuch why did you close this?

jarredkeuch:
Good morning! It's not right. Not yet.

Here's our problem: A complex type is being used for the operation
input.When a complex type is used, the extension name (e.g. "
extension="tns:getMyStuff") should be used instead of the input name.
Unfortunately, node-soap only uses the input name.

I should have an updated pull request today.

jsdevel:
@jarredkeuch please do not close a PR and open a new one simply to make changes to it. You can easily force push to reset the state of your branch in git.

PR #920, opened 2017-02-14
If a complex type is used, use the complex type element name and not …

jarredkeuch:
If a complex type is used, use the complex type element name and not the input name.

Here's an excerpt:
```
    <wsdl:message name="getWidgetSummaryByWidgetNumberRequestMsg">
        <wsdl:part name="getWidgetSummaryByWidgetNumberParameters" element="tns:getWidgetSummaryByWidgetNumber">
        </wsdl:part>
    </wsdl:message>
```
The server is expecting "getWidgetSummaryByWidgetNumber" in the root node, but node-soap uses "getWidgetSummaryByWidgetNumberRequestMsg".

I've tested with SoapUI and it correctly sets the root node to "getWidgetSummaryByWidgetNumber".


jsdevel:
@jarredkeuch please do not close a PR and open a new one simply to make changes to it. You can easily force push to reset the state of your branch in git.

jarredkeuch:
Good to know. Will do (or should I say will not do). Thanks!

PR #921, opened 2017-02-14
If a complex type is used, use the complex type element name and not …

jarredkeuch:
If a complex type is used, use the complex type element name and not the input name.

Here's an excerpt:
```
    <wsdl:message name="getWidgetSummaryByWidgetNumberRequestMsg">
        <wsdl:part name="getWidgetSummaryByWidgetNumberParameters" element="tns:getWidgetSummaryByWidgetNumber">
        </wsdl:part>
    </wsdl:message>
```
The server is expecting "getWidgetSummaryByWidgetNumber" in the root node, but node-soap uses "getWidgetSummaryByWidgetNumberRequestMsg".

I've tested with SoapUI and it correctly sets the root node to "getWidgetSummaryByWidgetNumber".

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -302,7 +302,7 @@ Client.prototype._invoke = function(method, args, location, callback, options, e
     assert.ok(!style || style === 'document', 'invalid message definition for rpc style binding');
 
     // If a complex type is used, use the complex type element name and not the input name.
-    var rootElement = input.name;
+    var rootElement = input.$name;
     if (input.children.length === 1 && input.children[0].$element) rootElement = input.children[0].$element;
 
     // pass `input.$lookupType` if `input.$type` could not be found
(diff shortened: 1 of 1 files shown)


jarredkeuch:
Still not right. Getting "getWidgetSummaryByWidgetNumber is not present in wsdl and cannot be processed correctly." Hmm...

jsdevel:
@jarredkeuch please do not close a PR and open a new one simply to make changes to it.  You can easily force push to reset the state of your branch in git.

mattbrictson:
@jarredkeuch did you ever solve this problem? I am running into the same thing.

mattbrictson:
Also related: https://stackoverflow.com/questions/45636977/how-can-i-configure-the-root-element-of-a-node-soap-request-body

jarredkeuch:
We switched to using strong-soap.
https://www.npmjs.com/package/strong-soap

PR #922, opened 2017-02-20
add array options for namespaces in WSDL to set into Envelop

davidbayo10:
I need to preserve some xml namespaces from schemas.xmlsoap.org and www.w3.org into the envelop from the wsdl. The _xmlnsMap filter always the namespaces without criterial, so I added an option to configure that.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1068,7 +1068,8 @@ var WSDL = function(definition, uri, options) {
       }
 
       // prepare soap envelope xmlns definition string
-      self.xmlnsInEnvelope = self._xmlnsMap();
+      var xmlnsIntoEnvelope = options.xmlnsIntoEnvelope;
+      self.xmlnsInEnvelope = self._xmlnsMap(xmlnsIntoEnvelope);
 
       self.callback(err, self);
     });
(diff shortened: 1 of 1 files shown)


davidbayo10:
if you see correct the changes and if you think is necessary test, then I will add it.

jsdevel:
@davidbayo10 I think this is acceptable.  In addition to tests, can you add the option in the README?

lib/wsdl.js
```diff
@@ -2081,14 +2082,20 @@ WSDL.prototype._fromServices = function(services) {
...
+WSDL.prototype._xmlnsMap = function (xmlnsIntoEnvelope) {
   var xmlns = this.definitions.xmlns;
   var str = '';
   for (var alias in xmlns) {
     if (alias === '' || alias === TNS_PREFIX) {
       continue;
     }
+
     var ns = xmlns[alias];
+    if (ns && xmlnsIntoEnvelope && ~xmlnsIntoEnvelope.indexOf(alias)) {
```
herom: do we really need the (imho too unclear) `~` operator? I'd rather go for a check of `!== -1` or use the `lodash.includes()` method (`_.includes(xmlnsIntoEnvelope)`) - just for maintaining a readable code. In addition we'd be able to drop the check for the existence of `xmlnsIntoEnvelope` while iterating

```javascript
...
if (ns && _.includes(xmlnsIntoEnvelope)) {
...
```

davidbayo10: But, if xmlnsIntoEnvelope does not exists, will _.includes throw an Error? i.e. xmlnsIntoEnvelope not defined. Of course, I think your solution is better than mine.

davidbayo10: Yes, I'll write some tests and I'll update the README

herom: I just wrote a little "proof of concept" that `_.includes()` won't fail if `xmlnsIntoEnvelope` is not given over at the lodash docs page:

```javascript
function test (xmlnsIntoEnvelope) {
  return _.includes(xmlnsIntoEnvelope, 'testNamespace');
}

// call it without params
test(); // the output is "false",

// with a string that contains the given namespace
test('mytestNamespace');  // the output is "true"
```

so, I think we are save to go with the lodash function 😃

davidbayo10: Finally I used lodash include method

iGitScor:
Hi,

any news about this pull request ?
If i well understand it, it allows to add namespace declaration ? (like `xmlns:v2="http://domain.com/GenericEventMsg/v2"` or `xmlns:even="EventServiceCBS.SOAP.MyAction_v2"`

Thank you

PR #923, opened 2017-03-14
Fix missing namespace declaration on array if the namespace is already declared with another prefix.

Mik13:
There is a problem, if the same namespace has another prefix in the schema and the soap definition (in the tests I call them "c" and "common").

I've had to fix some other tests, because there were some inconsistent handling with xmlns-attributes on arrays and non-arrays.


jsdevel:
Looks good to me.  Thoughts @herom ?

herom:
Thanks a lot @Mik13  👍

PR #925, opened 2017-03-22
typo in server example

vladimirbuskin:
tried to run and found, works like a charm after change


herom:
thanks a lot @vladimirbuskin 👍

PR #928, opened 2017-05-13
Remove dependency to ursa

mfred488:
This pull request removes the dependency to ursa.

Since ursa made it quite difficult to build this project on Windows (cf #832), I had a look at how it's used.
It turns out that it's used only to decrypt the private key when it is protected by a passphrase. But node can actually do this on its own (the signature computation being eventually done by Crypto's Sign.sign (cf https://nodejs.org/api/crypto.html#crypto_sign_sign_private_key_output_format).
Giving the private key together with the passphrase to xml-crypto's SignedXml turns out to be enough.

Readme and tests are updated accordingly. One test case is added to make sure that WSSecurityCert can be used with a password-protected private key.

test/sequence-test.js is updated (formatting only) to pass jshint test.


mfred488:
Ew, this does not work on node <= 0.10 :(, but works on 0.11.
By any chance, would you consider dropping support of node <= 0.10

herom:
Thanks a lot for your contribution @mfred488 - we appreciate it a lot 👍 

I'm totally up for dropping support for archaic node versions, since I think we all should move on and embrace the new possibilities with the latest node releases. What about you @jsdevel ?

herom:
@mfred488 we dropped the support for `node.js` < `4.x` now (see #932 ). would you mind rebasing your branch in order to re-run the tests? 👍

mfred488:
Hey @herom ! I just rebased and everything went smoothly; unfortunately some (seemingly unrelated) tests are failing (they are already failing on 94a8b54e6982b707594475874325a8f1ef0446c9 as far as I can see).

I'll have a look at it (after a quick look it seems related to #933) end of this week.

herom:
ah yes, the failing tests were from #933 were merged into this library to have a set of failing tests to work on a lib-breaking regression. I provided a possible fix for this in #935 and waiting for some review until this get's finally merged and have our build-tests up and running again. 👍

herom:
thanks a ton @mfred488 👍

mfred488:
Thank you for the merge @herom ! May I ask a rough estimation of when the next version will be released ?

PR #914, opened 2017-01-16
add fields sequence if it's required (need help)

vfrbgt:
I want sort args on call soap method if sequence required(check parent element sequence). It's my implement for this functional. But I have problem 3 test not passed. Please help me fix it. I really need this functionality in your library. Thank you.

PR #929, opened 2017-05-22
Remove brackets from argsScheme string if they exist

alanmarcell:
We have an argsScheme issue due its string contains brackets. Removing it does not affect existing tests ans solves the problem

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -212,6 +212,7 @@ Client.prototype._setSequenceArgs = function(argsScheme, args) {
     return args;
   }
   for (var partIndex in argsScheme) {
+    partIndex = partIndex.replace('[]','')
     if(typeof args[partIndex] === 'undefined') {
       continue;
     }
--- a/test/sequence-test.js
+++ b/test/sequence-test.js
@@ -1,12 +1,12 @@
 'use strict';
 
 var fs = require('fs'),
-  soap = require('..'),
-  http = require('http'),
-  assert = require('assert'),
-  _ = require('lodash'),
-  sinon = require('sinon'),
-  wsdl = require('../lib/wsdl');
+    soap = require('..'),
+    http = require('http'),
+    assert = require('assert'),
+    _ = require('lodash'),
+    sinon = require('sinon'),
+    wsdl = require('../lib/wsdl');
 
 var sequencedRequest = {
   'url': 'https://github.com',
--- a/test/sequence-test.js
+++ b/test/sequence-test.js
@@ -84,7 +84,7 @@ describe('Method args sequence', function () {
     });
   });
 
-  it('should return args if args[partIndex] === \'undefined\'', function (done) {
+  it('should run setSequence ignoring brackets', function (done) {
     soap.createClient(__dirname + '/wsdl/sequnceexmple.wsdl', {
       suffix: '',
       options: {}
--- a/test/sequence-test.js
+++ b/test/sequence-test.js
@@ -1,12 +1,12 @@
 'use strict';
 
 var fs = require('fs'),
-    soap = require('..'),
-    http = require('http'),
-    assert = require('assert'),
-    _ = require('lodash'),
-    sinon = require('sinon'),
-    wsdl = require('../lib/wsdl');
+  soap = require('..'),
+  http = require('http'),
+  assert = require('assert'),
+  _ = require('lodash'),
+  sinon = require('sinon'),
+  wsdl = require('../lib/wsdl');
 
 var sequencedRequest = {
   'url': 'https://github.com',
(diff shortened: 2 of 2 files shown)


test/sequence-test.js
```diff
@@ -1,12 +1,12 @@
 'use strict';
 
 var fs = require('fs'),
-    soap = require('..'),
-    http = require('http'),
-    assert = require('assert'),
-    _ = require('lodash'),
-    sinon = require('sinon'),
-    wsdl = require('../lib/wsdl');
+  soap = require('..'),
```
jsdevel: please remove any changes that aren't related to your fix.

test/sequence-test.js
```diff
@@ -28,28 +28,71 @@ var notSequencedRequest = {
...
+  });
 
-    it('check sort args on sequence required method', function(done) {
-        soap.createClient(__dirname + '/wsdl/sequnceexmple.wsdl', { suffix: '', options: {} }, function(err, client) {
-          assert.ok(client);
-          var sequencedMethodRequest = client._setSequenceArgs(client._getArgsScheme('pullFile'), notSequencedRequest);
-          assert.ok(JSON.stringify(sequencedMethodRequest) === JSON.stringify(sequencedRequest));
-          done();
-        });
+  it('check sort args on sequence required method', function (done) {
```
jsdevel: `it('should ...`

test/sequence-test.js
```diff
@@ -28,28 +28,71 @@ var notSequencedRequest = {
...
-        assert.ok(client._isSequenceRequired('pullFile') === false);
 
-        soap.createClient(__dirname + '/wsdl/sequnceexmple.wsdl',  { suffix: '', options: {} }, function(err, client) {
-          assert.ok(client);
-          assert.ok(client._isSequenceRequired('pullFile') === true);
-          done();
-        });
+describe('Method args sequence', function () {
+
+  it('check if method required sequence args', function (done) {
```
jsdevel: `it('should ...`

lib/client.js
```diff
@@ -212,6 +212,7 @@ Client.prototype._setSequenceArgs = function(argsScheme, args) {
     return args;
   }
   for (var partIndex in argsScheme) {
+    partIndex = partIndex.replace('[]','')
```
jsdevel: This seems like a pretty big breaking change to me.  Just because there aren't tests, doesn't mean it's good for use.  What are we trying to solve here?

AlanMarcell: Well, I agree it's a very specific issue, then i forked and create my own npm package to solve urgent my issue, but I decided to pull request it anyway.
I think it's a error with the SOAP server, but I have an argScheme with brackets while it should be just a string. If I try to pass the arg with the bracket the application crashes

test/sequence-test.js
AlanMarcell: These test was already there
My test start with should, by the way

lib/client.js
herom: If it's such a specific scenario I guess we should guard it behind some option? I know it won't be an elegant solution, but it would fit both needs if the option will trigger the newly introduced behaviour and proceed "as is" if not provided...?

jsdevel: we could end up breaking things for devs that _depend_ on the brackets being there, so I think an option would be good.

AlanMarcell: I tried to use with the brackets, but I had an error with parse response:

>     Error: Error: Cannot parse response
      at kplClient.CadastrarCliente (src/connectors/applications/kpl/services/kplService.js:35:25)
      at node_modules/soap/lib/client.js:180:7
      at finish (node_modules/soap/lib/client.js:451:14)
      at parseSync (node_modules/soap/lib/client.js:436:12)
      at node_modules/soap/lib/client.js:412:14
      at Request._callback (node_modules/soap/lib/http.js:118:5)
      at Request.self.callback (node_modules/request/request.js:188:22)
      at Request.<anonymous> (node_modules/request/request.js:1171:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1091:12)
      at endReadableNT (_stream_readable.js:975:12)
      at _combinedTickCallback (internal/process/next_tick.js:80:11)
      at process._tickCallback (internal/process/next_tick.js:104:9)

jsdevel: sounds like you have a bug in your code

AlanMarcell: It happens when I try to use the string with the brackets. I don't know if the field can have brackets at all, but in my case it cannot. 
If i use the field without brackets, the field does not come to my args, due its difference to the field from the Scheme, with the brackets.

herom: @AlanMarcell is it possible for you to share the source code of your `kplService.js`? The `client.parseSync()` is "just" a proxy to `wsdl.xmlToObject()` with a `try/catch` block around it. So, if anything, the `xmlToObject()` method would fail.

AlanMarcell: Yes, of course
```javascript
'use strict';

let
  soap = require('soap');

class KplService {

  createConnection(url) {
    return new Promise((resolve, reject) => {
      soap.createClient(url, (err, client) => {
        if (err) {
          return reject(new Error(err));
        }
        return resolve(client);
      });
    });
  }

  createClient(kplClient, args) {
    return new Promise((resolve, reject) => {
      kplClient.CadastrarCliente(args, (err, response) => {
        if (err) {
          return reject(new Error(err));
        }
        if (response.CadastrarClienteResult.ResultadoOperacao.Codigo !== 200002) {
          return reject(new Error(response.CadastrarClienteResult.ResultadoOperacao.ExceptionMessage));
        }
        return resolve(response.CadastrarClienteResult.Rows);
      });
    });
  }

  createOrder(kplClient, args) {
    return new Promise((resolve, reject) => {
      kplClient.InserirPedido(args, (err, response) => {
        if (err) {
          return reject(new Error(err));
        }
        if (response.InserirPedidoResult.ResultadoOperacao.Codigo !== 200001) {
          return reject(new Error(response.InserirPedidoResult.ResultadoOperacao.ExceptionMessage));
        }
        return resolve(response.InserirPedidoResult.Rows);
      });
    });
  }
}

AlanMarcell: I call the service this way:

```javascript
[...]
        const kplClient = await KplService.createConnection(url)

        args.ListaDeClientes = {
          'DadosClientes': DadosClientes
        }

        // I tried this way:
        // args.ListaDeClientes = {  
        //   'DadosClientes[]': DadosClientes
        // }

        // And this way:
        // args.ListaDeClientes = [{
        //   'DadosClientes': DadosClientes
        // }]
        // With no success

        let res = await KplService.createClient(kplClient, args)
[...]

herom:
This is probably (or better to say, I'm sure it is) describing what the failing test added with #933 describes.

herom:
@AlanMarcell would you mind checkin' out the `broken` branch of my forked repo at https://github.com/herom/node-soap/tree/broken and try your luck with the latest fix? I really do think that this relates to your issue and is probably a fix which allows us to fix the regression and keep the newly introduced possibility of having a "sequenced" output 👍

jsdevel:
Reverting #914: https://github.com/vpulim/node-soap/pull/937

herom:
@AlanMarcell would you mind trying the new `0.19.1` build and see if this fixes your bug?

smahdi:
Hi everyone
I'm sorry I mention it here, I could not find the issue place!!! :-)
I have an issue:
`console.log(client.describe().soap_ansarMobApp.soapPort.login)`
.
.
Output:
`login: {
    "input": {
        "context": {
            "data[]": {
                "key": "xs:string",
                "value": "xs:string",
                "targetNSAlias": "tns",
                "targetNamespace": "http://sample.com/"
            },
            "targetNSAlias": "tns",
            "targetNamespace": "http://sample.com/"
        },
        "request": {
            "password": "xs:string",
            "username": "xs:string",
            "targetNSAlias": "tns",
            "targetNamespace": "http://sample.com/"
        },
        "targetNSAlias": "tns",
        "targetNamespace": "http://sample.com/"
    },
    "output": {
        "return": "xs:string",
        "targetNSAlias": "tns",
        "targetNamespace": "http://sample.com/"
    }
}`
I want to know what object should I use to send?
`client.login(loginObj, function (err, result) {})`

I use : `loginObj: {
       "context": [],
       "request": {
              "username": "9382742",
              "password": "62835143"
       }`
but get **ValidatorWSException**.

Thank you
Always happy and victorious.

PR #935, opened 2017-05-29
[FIX] [DO NOT MERGE] Possible fix #933 (regression because of introduction of new be…

herom:
…haviour in #914)

This is very much "work in progress", so please feel free to review/comment/contribute until it's ready to merge and ready to use 😺

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -18,7 +18,7 @@ var HttpClient = require('./http'),
 
 var Client = function(wsdl, endpoint, options) {
   events.EventEmitter.call(this);
-  options = options || {};
+  options = Object.assign({}, options);
   this.wsdl = wsdl;
   this._initializeOptions(options);
   this._initializeServices(endpoint);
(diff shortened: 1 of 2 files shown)


herom:
What's sad about this is, that we'd have a tradeoff and introducing yet another "option" which may or may not get passed to the `createClient()` method, but I couldn't think of any better solution in order to prevent the client from forcing the sequencing behaviour which was introduced earlier and broke our latest release - which means it made it unusable for a lot of users.... 😞

herom:
can I have your opinion on this fix @jsdevel ? 😸

jsdevel:
I don't think an option for this is good.  Perhaps we should just revert https://github.com/vpulim/node-soap/pull/914.  Do we know what releases that PR made it into?

jsdevel:
https://github.com/vpulim/node-soap/pull/937

herom:
Closing due to #937

PR #936, opened 2017-05-30
upgrade to lodash 4.17.4

rflament:
Lodash 4.x merge function does not take customizer function as the third argument anymore : 

https://lodash.com/docs/4.17.4#merge

Instead we have to use the function mergeWith

That's problematic when a dependency other than node-soap requires lodash 4.x : lodash 4.x ends up being used and wsdl inclusion does not work anymore.

This pull request switches to lodash 4.17.4 and use mergeWith function to fix this.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -88,7 +88,7 @@ function trim(text) {
 }
 
 function deepMerge(destination, source) {
-  return _.merge(destination || {}, source, function(a, b) {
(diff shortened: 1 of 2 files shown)


jsdevel:
please fix the broken tests

herom:
@jsdevel the test which are failing are the ones that we merged in with #933 

@rflament could you please rebase your branch so that it's on par with the latest master (with #937 merged in) - the test should pass afterwards 👍

PR #940, opened 2017-06-09
Failing test case for recursive types

DullReferenceException:
XML generation in a service with a recursively-defined type is causing a stack overflow. There appears to be infinite recursion going on. Here's the repeating stack trace fragment:

```
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1978:19)
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1984:20)
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1984:20)
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1978:19)
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1984:20)
      at WSDL.findChildSchemaObject (/Users/jpage/Code/node-soap/lib/wsdl.js:1984:20)
```

This boiled-down version of our code in the new test case reproduces it.

Note that versions before `0.15.0` did _not_ have this bug.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1932,10 +1932,22 @@ WSDL.prototype.findSchemaType = function(name, nsURI) {
   return schema.complexTypes[name];
 };
 
-WSDL.prototype.findChildSchemaObject = function(parameterTypeObj, childName) {
+WSDL.prototype.findChildSchemaObject = function(parameterTypeObj, childName, backtrace) {
   if (!parameterTypeObj || !childName) {
     return null;
   }
+
+  if (!backtrace) {
+    backtrace = [];
+  }
+
+  if (backtrace.indexOf(parameterTypeObj) >= 0) {
+    // We've recursed back to ourselves; break.
+    return null;
+  } else {
+    backtrace = backtrace.concat([parameterTypeObj]);
+  }
+
   var found = null,
       i = 0,
       child,
@@ -1975,13 +1987,13 @@ WSDL.prototype.findChildSchemaObject = function(parameterTypeObj, childName) {
     }
     var typeDef = this.findSchemaType(typeInfo.name, childNsURI);
     if (typeDef) {
-      return this.findChildSchemaObject(typeDef, childName);
+      return this.findChildSchemaObject(typeDef, childName, backtrace);
     }
   }
 
   if (object.children) {
     for (i = 0, child; child = object.children[i]; i++) {
-      found = this.findChildSchemaObject(child, childName);
+      found = this.findChildSchemaObject(child, childName, backtrace);
       if (found) {
         break;
       }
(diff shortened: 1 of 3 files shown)


jsdevel:
Thanks @DullReferenceException !  Hopefully someone can pick this up and submit a PR.  Any ideas @herom on what may have introduced this?

DullReferenceException:
Aw yeah, found a fix which is basically just maintaining a recursion backtrace. Does this fix look ok?

jsdevel:
Amazing @DullReferenceException !  Thanks!

DullReferenceException:
Thank you for the quick merge!

jsdevel:
Np!

DullReferenceException:
@jsdevel: any estimate on when this will be published?

jsdevel:
Published as `v0.19.2` @DullReferenceException

lib/wsdl.js
```diff
@@ -1932,10 +1932,22 @@ WSDL.prototype.findSchemaType = function(name, nsURI) {
...
+
+  if (!backtrace) {
+    backtrace = [];
+  }
+
+  if (backtrace.indexOf(parameterTypeObj) >= 0) {
+    // We've recursed back to ourselves; break.
+    return null;
+  } else {
+    backtrace = backtrace.concat([parameterTypeObj]);
```
herom: just to understand it better: what's the purpose of using `concat()` and creating a new `Array` instead of just `push()`ing ?

DullReferenceException: If we mutate the array (in the case when it's passed in), then after this function returns, the array will erroneously carry the new value to the next iteration, making it no-longer reflect the backtrace.

herom: I'm always eager to learn new things but I'm afraid I not able to follow you here 😸 

Would it be possible for you to create a jsbin or gist or something like that in order for me to get the difference? Maybe it's because it's already late in the evening (for me) :austria:  I would be very thankful for this 👍 💯

DullReferenceException: Sure. Let's say we're traversing this schema:

```
foo
  bar
    foo
    baz
  blah
zub
```

If we used `.push()`, here's what the array would look like as we do the depth-first traversal:

```
[foo]
[foo, bar]
// (here we encounter foo again, so we skip)
[foo, bar, baz]
[foo, bar, baz, blah] // Incorrect; should be [foo, blah]
[foo, bar, baz, blah, zub] // Incorrect; should be [zub]
```

herom: thanks a ton @DullReferenceException - it seems it was really too late for me 😄 - totally clear now 👍

PR #941, opened 2017-06-15
Adding a note about consulting in the README.

nonedeactivated:
exchangeId on Client Events does not seems to work. It appears that exchangeId is not being passed as described by the documentation. I even tried setting exchangeId on the initial request and it didn't seem to make any difference.


jsdevel:
@nonedeactivated I think your branch has some issues.

PR #946, opened 2017-06-26
Fix vulnerability dependency

tobiaslins:
in order to fix the `ms` vulnerability we should consider to update the finalhandler dependency


jsdevel:
works for me!  thanks!

PR #947, opened 2017-07-10
Fix for obj.hasOwnProperty is undefined if obj was created with Object.create(null)

NikitaKurpas:
Fix for obj.hasOwnProperty is undefined if obj was created with Object.create(null).

This is a little issue that bugged us a lot. We are creating objects with `Object.create(null)` and passing these objects to node-soap and it fails.


jsdevel:
How about a refactor to `Object.keys` or  `_.toPairs` instead?

drbignik:
Yeah, I can do that instead.

herom:
@jsdevel @drbignik  while `Object.keys` or `_.toPairs` will do the job, would it guarantee that it's iterating over the objects own properties, which are not inherited? If it's just used as a guard for the `for...in` loop, it would probably be good enough to only check if it's there, like `if (obj[key])`?

jsdevel:
@herom per the docs for [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys):

```
The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).
```

I tested locally with the following code snippet and it worked just fine:

```js
var a = Object.create(null);
a.foo = 5;
Object.keys(a);
```

herom:
@jsdevel thanks - I just made a (not very well thought out) guess out of the blue 😸

jsdevel:
is there a way to add a test for this?

PR #948, opened 2017-07-10
Remove not needed header declaration on request function

dcolonv:
Issue: Allocation of unused variables is not recommended.
Fix: Remove unused declaration for var header on request function


jsdevel:
Thanks!

PR #949, opened 2017-07-10
Use children's base type as the type of restriction and simpleType elements

mfred488:
While using node-soap, I realized that the type of restricted simple types (as described, for instance, in the first example [here](https://www.w3schools.com/xml/schema_facets.asp)) is not correctly taken into account.

In wsdl.js:
- the description of a simple type was the concatenation of the simple type's name and its child's description, separated by a "|"
- the description of a restriction element was the concatenation of the restriction's name, a "|", then the list of comma-separated names of the restriction's children
It seems that this has been around for quite some time, but as far as I can see this is used nowhere else in the code base.

In the case of a "restricted integer" (as described in the previously mentioned example of w3schools), this unusual type description is incompatible with the logic of wsdl.js which, for instance, converts a field to an integer when needed (see wsdl.js:1428). As a result, this kind of fields are simply returned by node-soap as strings, while we could be already parsed into integers according to the schema definition.

This pull requests simplifies the description of `restriction` and `simpleType` elements, so that the previously mentioned example is correctly parsed as an integer. A test case is added to reproduce the issue and ensure that 

The second commit is just a formatting fix, so that `npm test` passes.

### DIFF
--- a/test/client-customHttp-xsdinclude-test.js
+++ b/test/client-customHttp-xsdinclude-test.js
@@ -77,8 +77,8 @@ it('should allow customization of httpClient, the wsdl file, and associated data
           DummyPortType: {
             Dummy: {
               "input": {
-                "ID": "IdType|xs:string|pattern",
-                "Name": "NameType|xs:string|minLength,maxLength"
+                "ID": "xs:string",
+                "Name": "xs:string"
               },
               "output": {
                 "Result": "dummy:DummyList"
--- a/test/wsdl-test.js
+++ b/test/wsdl-test.js
@@ -129,7 +129,7 @@ wsdlStrictTests['should get empty namespace prefix'] = function(done) {
 };
 
 wsdlNonStrictTests['should load same namespace from included xsd'] = function(done) {
-  var expected = '{"DummyService":{"DummyPortType":{"Dummy":{"input":{"ID":"IdType|xs:string|pattern","Name":"NameType|xs:string|minLength,maxLength"},"output":{"Result":"dummy:DummyList"}}}}}';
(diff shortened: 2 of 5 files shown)


test/wsdl-test.js
```diff
@@ -129,7 +129,7 @@ wsdlStrictTests['should get empty namespace prefix'] = function(done) {
 };
 
 wsdlNonStrictTests['should load same namespace from included xsd'] = function(done) {
-  var expected = '{"DummyService":{"DummyPortType":{"Dummy":{"input":{"ID":"IdType|xs:string|pattern","Name":"NameType|xs:string|minLength,maxLength"},"output":{"Result":"dummy:DummyList"}}}}}';
+  var expected = '{"DummyService":{"DummyPortType":{"Dummy":{"input":{"ID":"xs:string","Name":"xs:string"},"output":{"Result":"dummy:DummyList"}}}}}';
```
jsdevel: breaking change

test/client-customHttp-xsdinclude-test.js
```diff
@@ -77,8 +77,8 @@ it('should allow customization of httpClient, the wsdl file, and associated data
           DummyPortType: {
             Dummy: {
               "input": {
-                "ID": "IdType|xs:string|pattern",
-                "Name": "NameType|xs:string|minLength,maxLength"
+                "ID": "xs:string",
```
jsdevel: breaking change

romankrejciapitree:
Hi,
i have same problem (exactly with xs:int) and want to ask, if author can repair this PR please?
thanks in advance

PR #952, opened 2017-07-27
Adds a wsdlOption 'returnNilAsNull' which returns xml tags which have…

aconran:
… been marked as nil as null values in JavaScript.


aconran:
When a SOAP response contains xml tags with nil attributes they are intentionally omitted from the JS object (in the XML -> JS conversion). For my use-case I need to know when they are set to nil and would prefer that it is converted to null. See https://github.com/vpulim/node-soap/blob/master/lib/wsdl.js#L1370-L1372

I need to distinguish when nil values are returned in the SOAP response. I saw a request on stackoverflow for this behavior as well https://stackoverflow.com/questions/40388453/node-soap-including-nil-values-in-result

lib/wsdl.js
```diff
@@ -1112,6 +1112,12 @@ WSDL.prototype._initializeOptions = function (options) {
   } else {
     this.options.returnFault = false;
   }
+  if (options.returnNilAsNull !== undefined) {
```
herom: we could set it like `this.options.returnNilAsNull = !!options.returnNilAsNull` which would set it to the expected boolean value in every case.

in general - and especially when looking through your tests - wouldn't it be a better fit if we name the option `handleNilAsNull` instead?

aconran: @herom I went through a few different iterations of naming (convertNilToNull, omitNullValues, etc). I thought that returnNilAsNull was consistent with the option directly above in the code returnFault.  Glad to change to whatever maintainers feel is consistent with project.

aconran: @herom please advise how you'd like me to proceed. I will update PR as you see fit.

herom: @aconran sorry, somehow I managed to oversee your notification 😞 

I'd favor for `handleNilAsNull` as this is describing the scenario 💯 % 😄

aconran: @herom done, rebased and squashed

herom:
Thanks a ton @aconran 👍 

Could you please do me a favour and open an accompanying PR where you document this new option in our `README.md`? This would be awesome 😸

brunogasparetto:
This feature is working? I tried, but the properties is still missing.

aconran:
@brunogasparetto returnNilAsNull got renamed to handleNilAsNull; looks like it's still in the current code base
https://github.com/vpulim/node-soap/blob/master/src/wsdl/index.ts#L363

brunogasparetto:
@aconran I see. But the Client isn't set the wsdl option: https://github.com/vpulim/node-soap/blob/4660fa2c19c1cae869436173ea038c3292a5a8de/src/client.ts#L173C21-L173C21

I tried pass the wsdl option after create the Client and it worked.

```javascript
const client = await soap.createClientAsync(uri);
client.wsdl.options.handleNilAsNull = true;
```

brunogasparetto:
Sorry. It was cache. Passing `disableCache: true` in the options worked as well.

PR #953, opened 2017-07-30
Adding promise based async to client creation and soap client with Bluebird Promise library

siimeon:
# Adding promise returning variants of soap client creation and function calls within client

I wanted to use promises to handle async functionality of client and added [Bluebird](http://bluebirdjs.com/docs/getting-started.html) Promises to `createClientAsync` and all client function calls. This helps to keep code clean in my mind.

With these changes its possible to use `.then()` and nodes new `async/await` functionality

## Examples

```javascript
function createSoapClientAndMakeApiCall () {
  var soap = require('soap');
  var args = {name: 'value'};
  soap.createClientAsync(url).then((client) => {
    return client.MyFunctionAsync(args);
  }).then((result) => {
    console.log(result);
  });
}
```

```javascript
async function createSoapClientAndMakeApiCall () {
  var soap = require('soap');
  var args = {name: 'value'};
  var client  = await soap.createClientAsync(url);
  var result = await MyFunctionAsync(args);
  console.log(result);
}
```

## Updated .travis.yml

For my testing I also added `node` for NodeJS environments

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -14,6 +14,7 @@ var HttpClient = require('./http'),
   findPrefix = require('./utils').findPrefix,
   _ = require('lodash'),
   concatStream = require('concat-stream'),
+  Promise = require("bluebird"),
   uuid = require('uuid');
 
 var Client = function(wsdl, endpoint, options) {
@@ -23,6 +24,7 @@ var Client = function(wsdl, endpoint, options) {
   this._initializeOptions(options);
   this._initializeServices(endpoint);
   this.httpClient = options.httpClient || new HttpClient(options);
+  Promise.promisifyAll(this);
 };
 util.inherits(Client, events.EventEmitter);
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ node_js:
   - 4.0
   - 6.9.4
   - 7.4.0
+  - node
 env:
   - CXX=g++-4.8
 addons:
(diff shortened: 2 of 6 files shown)


lib/client.js
```diff
@@ -14,6 +14,7 @@ var HttpClient = require('./http'),
   findPrefix = require('./utils').findPrefix,
   _ = require('lodash'),
   concatStream = require('concat-stream'),
+  Promise = require("bluebird"),
```
herom: could you please change the name from `Promise` to `BPromise` or something the like so that we don't collide with the **native** `Promise` class. 👍  (in `soap.js` too)

herom:
Thanks a lot @siimeon - that's much appreciated as I think that a lot of people are actually using `bluebird.js` to promisify our library anyways 👍

herom:
I didn't mean to offend you @siimeon - it would be enough if you'd rename it and add a new commit - thanks to GitHub we're able to "squash & merge" directly 👍 

So, if you'd reopen your PR and add an additional commit with the renaming I'd be happy to merge it :smiley:

siimeon:
I just tried to fix it so I will send new PR with that change :)

siimeon:
@herom there is now new PR https://github.com/vpulim/node-soap/pull/956 with the change that you asked 😃 
Sorry that I messed up and closed this PR by accident 😄

PR #954, opened 2017-07-31
EnvelopeKey header end tag bug

ahtaja:
Not using the optional envelopeKey when ending the headertag

### DIFF
--- a/lib/security/WSSecurityCert.js
+++ b/lib/security/WSSecurityCert.js
@@ -62,7 +62,7 @@ WSSecurityCert.prototype.postProcess = function (xml, envelopeKey) {
     id: this.x509Id
   });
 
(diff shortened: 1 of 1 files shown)


herom:
hey @ahtaja - may I ask you what was the reason for you to close (give up on) this PR?

ahtaja:
I dont know how to create tests you require and the fix took just one minute to do... so I didnt feel like it was worth the hassle. :/

herom:
@ahtaja Oh, that's a pitty... we request the tests in order to reduce/prevent the occurence of regressions/bugs/code removal and keep our library "clean". Sorry if this has bothered you, but the codebase is rather large and I don't think we could keep it like it is without this strict request for tests.

PR #956, opened 2017-08-01
Adding promise based async to client creation and soap client with Bluebird Promise library

siimeon:
# Adding promise returning variants of soap client creation and function calls within client

I wanted to use promises to handle async functionality of client and added [Bluebird](http://bluebirdjs.com/docs/getting-started.html) Promises to `createClientAsync` and all client function calls. This helps to keep code clean in my mind.

With these changes its possible to use `.then()` and nodes new `async/await` functionality

## Examples

```javascript
function createSoapClientAndMakeApiCall () {
  var soap = require('soap');
  var args = {name: 'value'};
  soap.createClientAsync(url).then((client) => {
    return client.MyFunctionAsync(args);
  }).then((result) => {
    console.log(result);
  });
}
```

```javascript
async function createSoapClientAndMakeApiCall () {
  var soap = require('soap');
  var args = {name: 'value'};
  var client  = await soap.createClientAsync(url);
  var result = await MyFunctionAsync(args);
  console.log(result);
}
```

## Updated .travis.yml

For my testing I also added `node` for NodeJS environments

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -14,6 +14,7 @@ var HttpClient = require('./http'),
   findPrefix = require('./utils').findPrefix,
   _ = require('lodash'),
   concatStream = require('concat-stream'),
+  BluebirdPromise = require("bluebird"),
   uuid = require('uuid');
 
 var Client = function(wsdl, endpoint, options) {
@@ -23,6 +24,7 @@ var Client = function(wsdl, endpoint, options) {
   this._initializeOptions(options);
   this._initializeServices(endpoint);
   this.httpClient = options.httpClient || new HttpClient(options);
+  BluebirdPromise.promisifyAll(this);
 };
 util.inherits(Client, events.EventEmitter);
 
--- a/lib/soap.js
+++ b/lib/soap.js
@@ -10,6 +10,7 @@ var Client = require('./client').Client,
   HttpClient = require('./http'),
   security = require('./security'),
   passwordDigest = require('./utils').passwordDigest,
+  BluebirdPromise = require('bluebird'),
   wsdl = require('./wsdl'),
   WSDL = require('./wsdl').WSDL;
 
@@ -59,6 +60,20 @@ function createClient(url, options, callback, endpoint) {
   });
 }
 
+function createClientAsync(url, options, endpoint) {
+  if (typeof options === 'undefined') {
+    options = {};
+  }
+  return new BluebirdPromise(function(resolve, reject) {
+    createClient(url, options, function(err, client) {
+      if (err) {
+        reject(err);
+      }
+      resolve(client);
+    }, endpoint);
+  });
+}
+
 function listen(server, pathOrOptions, services, xml) {
   var options = {},
     path = pathOrOptions,
@@ -84,6 +99,7 @@ exports.ClientSSLSecurity = security.ClientSSLSecurity;
 exports.ClientSSLSecurityPFX = security.ClientSSLSecurityPFX;
 exports.BearerSecurity = security.BearerSecurity;
 exports.createClient = createClient;
+exports.createClientAsync = createClientAsync;
 exports.passwordDigest = passwordDigest;
 exports.listen = listen;
 exports.WSDL = WSDL;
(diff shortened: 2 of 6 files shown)


lib/client.js
```diff
@@ -14,6 +14,7 @@ var HttpClient = require('./http'),
   findPrefix = require('./utils').findPrefix,
   _ = require('lodash'),
   concatStream = require('concat-stream'),
+  BluebirdPromise = require("bluebird"),
```
siimeon: Promise renamed to BluebirdPromise 😃

lib/client.js
```diff
@@ -23,6 +24,7 @@ var Client = function(wsdl, endpoint, options) {
   this._initializeOptions(options);
   this._initializeServices(endpoint);
   this.httpClient = options.httpClient || new HttpClient(options);
+  BluebirdPromise.promisifyAll(this);
```
siimeon: Promise renamed to BluebirdPromise 😃

lib/soap.js
```diff
@@ -10,6 +10,7 @@ var Client = require('./client').Client,
   HttpClient = require('./http'),
   security = require('./security'),
   passwordDigest = require('./utils').passwordDigest,
+  BluebirdPromise = require('bluebird'),
```
siimeon: Promise renamed to BluebirdPromise 😃

siimeon:
I would guess that mixing native promises and bluebird promises don't cause any problems especially now that bluebird promises are renamed.
Also native promises don't offer possibility `promisifyAll`so this kind simple adding of promises isn't possible with native promise

aleung:
@siimeon The `BluebirdPromise` naming is only for internal.  The function `createClientAsync(url)` returns a Bluebird Promise so application code has to handle Promises created by both Bluebird and other promise library it chooses.

You code seems to be just wrapping the `soap` API. What if you create a new module `soap-promise-bluebird` and leave `soap` module un-promisified? Then other users who don't use Bluebird can create another wrapper with their favorite promise library.

`request-promise` is created in this way.

jsdevel:
Looks good to me!  Thoughts @herom ?

herom:
>Is there a way to make promise library replaceable?
I'd love to see that too - that would be awesome! Especially, if we could make it really "easy cheesy" to replace it by just setting `soap.Promise = myPromiseLib;` and that's it 🥇 

@siimeon @jsdevel I'm fine with this PR as I think that it's a valuable addition for the `soap` library 👍 - nonetheless I'd appreciate it if don't loose track of the idea/advice @aleung gave us here... as we don't have the issues active anymore, should we open a placeholder PR for this?

siimeon:
I can publish that promisified version of `soap` if you feel like it doesn't suit for your plans for this library 🙂 I anyway need to do that since I'm already using my version so its much more convenient for me to create package from that.

herom:
@siimeon I'm happy to merge your PR (if you're ok with it too @jsdevel) 💯 👍 

I just don't want to loose track of the future(?) addition of being able to replace the used `Promise` by "your own lib" or even use native `Promise` if you wish.

aleung:
The code uses bluebird `promisefyAll`, which may not be available in other promise lib. Maybe it's not so easy to replace it by just setting `soap.Promise = myPromiseLib;`. 

If decomposing the library into core and plug-ins, you may create an organization in Github then put all related projects under this organization. One project or multiple projects, each has pros and cons. Hard to say which is better.

My application is using native promise right now. If `soap` introduce Bluebird, I'm considering to switch to Bluebird as well. So it's also OK for me.

herom:
> The code uses bluebird promisefyAll, which may not be available in other promise lib

I see your point. My (at the time just partial and not well thought through) approach would be to wrap the needed methods with `Native Promise` known functions (so, no `finally()`) or the like, which would make it easy to switch `Promise` libs like I described.

> you may create an organization in Github then put all related projects under this organization

yeah, I also thought about this... perhaps we should discuss this with @vpulim @jsdevel?


@jsdevel ready to merge? 😸

jsdevel:
yup!

PR #958, opened 2017-08-08
Lookup definitions in child element first

lfantone:
I have this case where the `xmlns` definition is added as a `xmlns:{prefix}` attribute to the element and the _wsdl_ library wasn't able to find the Schema Object because It only looks into the global definitions.
Example of a not working xml schema:

```xml
  <xs:extension xmlns:q1="http://schemas.org/Request" base="q1:Info">
    <xs:sequence>
      <xs:element name="id" nillable="true" type="xs:string"/>
    </xs:sequence>
  </xs:extension>
```

This PR solves the issue describe above by looking the prefix in the child xmlns with a fallback to the `this.definitions.xmlns`.

I'm not sure if this this needs an specific unit test to being added, if so please let me know I'll try to add it.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -2048,7 +2048,7 @@ WSDL.prototype.findChildSchemaObject = function(parameterTypeObj, childName, bac
       if (child.$base) {
         var baseQName = splitQName(child.$base);
         var childNameSpace = baseQName.prefix === TNS_PREFIX ? '' : baseQName.prefix;
-        childNsURI = this.definitions.xmlns[baseQName.prefix];
+        childNsURI = child.xmlns[baseQName.prefix] || this.definitions.xmlns[baseQName.prefix];
 
         var foundBase = this.findSchemaType(baseQName.name, childNsURI);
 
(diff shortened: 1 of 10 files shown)


herom:
@lfantone thanks a lot for your contribution 👍 

please add a `request/repsonse` sample test, as required in our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) to make sure your fix won't get overwritten accidentially 🎲

jsdevel:
@lfantone this fix is _very_ much appreciated!  Echoing what @herom said, we hope you can add some `request/response` samples as this fix _is_ important enough to demand some.

crash7:
@jsdevel @herom Hi! I'm trying to create the request/response test for this, any hints? Thanks!

jsdevel:
> @jsdevel @herom Hi! I'm trying to create the request/response test for this, any hints? Thanks!
I'd just follow existing examples.

jsdevel:
Awesome job @crash7 !  I love seeing all that green and no red!  That's what we're after!

crash7:
@jsdevel excellent :smile:

PR #959, opened 2017-08-17
path.resolve cannot resolve a null path

0x4139:
changed the uri from null to empty string due to the fact the path.resolve can't resolve a null path.


jsdevel:
I'm fine with this.  How about you @herom ?

PR #962, opened 2017-08-27
Added useEmptyTag option to wsdlOptions to prefer <Tag /> format over <Tag></Tag>

whythecode:
One of the web services I was working with for some reason required that tags without a body be <Tag /> (empty, self-closing) rather than <Tag />. The XML spec makes no distinction between the two, but I had to work with the service.

**Change**: added useEmptyTag option to wsdlOptions that will prefer <Tag /> to <Tag></Tag> if the tag has no contents.

**Accompanying changes**:

- README.md documentation on extra switch
- Two tests:
  - should_use_empty_tags_when_directed: checks that when useEmptyTag is set, empty tags are preferred
  - should_use_empty_tags_only_when_directed: checks that when useEmptyTag is not set, the current behavior does not change

All credit belongs with @pxpeterxu . I'm reopening this since the minimum node requirement has actually changed, so that the failing tests in 0.10 are no longer an issue.


jsdevel:
Awesome!  Thanks @whythecode !

whythecode:
you're very welcome! thanks for taking care of this essential node module! one quick question, @jsdevel , can you tell me when there's a release planned? just so I know to keep an eye out and can update our package.json accordingly, no rush.

jsdevel:
Published as `v0.21.0`

jsdevel:
Done @whythecode https://github.com/vpulim/node-soap/pull/962#issuecomment-325395836

Concurrency in action!

whythecode:
wow, LOL, thanks, you rock!!!

PR #967, opened 2017-09-01
allow http endpoints for client

markbjerke:
With a trivial one line change our shop can connect to http endoints.  Behind firewalls there are enterprise soap services just on http.  We could use self-signed certs but it's a hassle and the change is so minor why not allow it  ? 

I added a test case.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -2215,7 +2215,7 @@ function open_wsdl(uri, options, callback) {
   var request_options = options.wsdl_options;
 
   var wsdl;
-  if (!/^https?:/.test(uri)) {
+if (!/^(http?:|https?:)/.test(uri)) {                
     debug('Reading file: %s', uri);
     fs.readFile(uri, 'utf8', function(err, definition) {
       if (err) {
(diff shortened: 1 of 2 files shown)


jsdevel:
Thanks @markbjerke.  I'm leaning towards making this an option.  What do you think?  It's somewhat of a breaking change, as previous behavior was to fail.  By blindly allowing both we could be breaking user's reliance on the check.  I'm open to your thoughts on this.  @herom ?

markbjerke:
Hello:
Yes, option would be fine, there are some security ramifications but it’s so rare that a public facing service would be over http, this is primarily a developer change, although we have a soap service that runs behind firewalls over http.

The change is trivial, the test passed.

Thanks
m.

jsdevel:
@markbjerke actually, upon looking at the regex, it seems the existing regex allows for http.  I tested it in my console:

regex

jsdevel:
Unless I'm missing something, I'm closing this PR @markbjerke.  http endpoints are already allowed by the existing regex.

markbjerke:
You’re right I was in the debugger and watched it enter the file path but I must have not typed the uri correctly, like http// instead of http://.

My bad…
-m

PR #968, opened 2017-09-01
Fix performance bug when converting POJO to xml

zarend:
Fix bug where converting PJO to xml created too may calls to findChildSchemaObject and caused a significant performance hit.
See: https://github.com/CumberlandGroup/node-ews/issues/58

Fix is for findChildSchemaObject to avoid recurring to irrelevant parts
of the schema.


zarend:
Thank you for responding quickly 😄 💥

PR #971, opened 2017-09-06
Remove compress dependency

tdecaluwe:
Functionality of the compress dependency is available through the node `zlib` module. This pull request removes the `compress` dependency and replaces it with `zlib`.

The PR also adds a test for the server compression. This test consists of sending the same request twice, once through a soap client, once through a gzip stream. Afterwards both responses are checked for equality.

I based the test on an example from the `request-response-samples` folder. This might not be ideal, so feel free to suggest another wsdl file and example request that I can use.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -18,13 +18,13 @@ function getDateString(d) {
 }
 
 var url = require('url'),
-  compress = null,
+  zlib = null,
   events = require('events'),
   util = require('util'),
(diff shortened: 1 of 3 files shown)


herom:
please rebase your PR with the latest master - I updated our `travis.yml` in order to fix the failing `node 4.x` tests 👍

herom:
Thanks a lot @tdecaluwe 👍

PR #972, opened 2017-09-27
Added preserveWhitespace option to prevent trimming leading / trailin…

samuelms1:
…g whitespace from text and cdata values

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,6 +126,7 @@ Client.prototype._initializeOptions = function(options) {
   this.streamAllowed = options.stream;
   this.wsdl.options.attributesKey = options.attributesKey || 'attributes';
   this.wsdl.options.envelopeKey = options.envelopeKey || 'soap';
+  this.wsdl.options.preserveWhitespace = options.preserveWhitespace || false;
   if(options.ignoredNamespaces !== undefined) {
     if(options.ignoredNamespaces.override !== undefined) {
       if(options.ignoredNamespaces.override === true) {
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1403,7 +1403,9 @@ WSDL.prototype.xmlToObject = function(xml, callback) {
   };
 
   p.oncdata = function (text) {
-    text = trim(text);
+    if (!self.options || !self.options.preserveWhitespace) {
+      text = trim(text);
+    }
     if (!text.length) {
       return;
     }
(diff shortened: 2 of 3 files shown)


samuelms1:
My team uses this soap library and we ran into an issue where a soap service we consume has values that begin with a space and the space is intended to be preserved (part of an identifier).

This change shouldn't break any existing behavior, rather it gives the ability to disable trimming whitespace for those who wish to turn that off.

lib/client.js
```diff
@@ -126,6 +126,7 @@ Client.prototype._initializeOptions = function(options) {
   this.streamAllowed = options.stream;
   this.wsdl.options.attributesKey = options.attributesKey || 'attributes';
   this.wsdl.options.envelopeKey = options.envelopeKey || 'soap';
+  this.wsdl.options.preserveWhitespace = options.preserveWhitespace || false;
```
herom: could you please use `!!` to convert into boolean (`this.wsdl.options.preserverWhitespace = !!options.preserveWhitespace`)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.21.0",
+  "version": "0.22.0",
```
herom: Please revert the version change - we'd like to change it ourselves when we create the next release (which will be really soon, promised 👍 )

lib/wsdl.js
```diff
@@ -1403,7 +1403,9 @@ WSDL.prototype.xmlToObject = function(xml, callback) {
   };
 
   p.oncdata = function (text) {
-    text = trim(text);
+    if (!self.options || !self.options.preserveWhitespace) {
```
herom: I guess it's save to check for `self.options.preserveWhitespace` only as the options should be set up anyways(?)

herom:
please rebase your PR with the latest master - I updated our `travis.yml` in order to fix the failing `node 4.x` tests 👍

samuelms1:
@herom Thanks for the feedback.  I think I've addressed it all, but let me know.

herom:
@samuelms1 awesome! thanks a lot for you contribution 💯 

Could you please do me a last favour and add a test which guarantees that the whitespaces get cut if the `preserverWhitespace` option is not set? Thus we have proof that both cases do still work and can't change secretly with further PRs being accepted 👍

samuelms1:
@herom Good idea.  I'll add that in now.

herom:
Awesome! 💯 Thanks a ton @samuelms1 👍 😺

PR #974, opened 2017-10-02
Allow agent reuse for ClientSSLSecurity if forever option is set

Mik13:
It should be possible for SSL connections to reuse the tls session. Every request, the https agent is recreated.

This PR changes this:
If the `forever` flag is set in the options per request (3rd parameter) or by default when setting the security (4th parameter), it should reuse the same agent as the previous requests also using `forever`.

The `request`-Module handles everything else, like setting the `keep-alive`-flag to the agent and the https module will cache the tls-session to the agent. Because of this, it is important that the same agent is used.

Tests added

### DIFF
--- a/lib/security/ClientSSLSecurity.js
+++ b/lib/security/ClientSSLSecurity.js
@@ -48,6 +48,8 @@ function ClientSSLSecurity(key, cert, ca, defaults) {
 
   this.defaults = {};
   _.merge(this.defaults, defaults);
+
+  this.agent = null;
 }
 
 ClientSSLSecurity.prototype.toXML = function(headers) {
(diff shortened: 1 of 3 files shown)


herom:
thanks a lot @Mik13 - could you please document the feature in our `README.md` file too? 👍

lib/security/ClientSSLSecurity.js
```diff
@@ -55,11 +55,26 @@ ClientSSLSecurity.prototype.toXML = function(headers) {
...
   options.cert = this.cert;
   options.ca = this.ca;
   _.merge(options, this.defaults);
-  options.agent = new https.Agent(options);
+
+  if (!!options.forever) {
+    if (!this.agent) {
+      options.keepAlive = true;
+
+      this.agent = new https.Agent(options);
```
herom: Please initialize this property also within the constructor-function

test/security/ClientSSLSecurity.js
```diff
@@ -49,4 +49,39 @@ describe('ClientSSLSecurity', function() {
...
+      var firstOptions = {};
+      var secondOptions = {};
+
+      instance.addOptions(firstOptions);
+      instance.addOptions(secondOptions);
+
+      firstOptions.agent.should.not.equal(secondOptions.agent);
+    });
+
+    it('should not generate different agents if paramter given', function () {
```
herom: Please change the description to "should return the same agent if paramter is present"

Mik13:
You're welcome @herom. I've changed the test description and added a short documentation to the readme. I hope this works

herom:
Awesome 💛 Thanks a ton @Mik13 😸

PR #976, opened 2017-10-02
Use assert.ifError function in tests

atavakoli:
All instances of `assert.ok(!err)` were replaced with `assert.ifError(err);`,
which provides better output when errors are encountered during testruns.
The former would only output the less-than-useful message,
"Uncaught AssertionError [ERR_ASSERTION]: false == true", while the latter
would output the actual error object that caused the assert to fail.

### DIFF
--- a/test/client-customHttp-test.js
+++ b/test/client-customHttp-test.js
@@ -100,7 +100,7 @@ it('should allow customization of httpClient and the wsdl file download should p
     {httpClient: httpCustomClient},
     function(err, client) {
       assert.ok(client);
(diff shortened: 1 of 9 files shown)


herom:
uargs... I messed up with the commit merge order :sheep: ... could you please rebase this PR @atavakoli ? thanks 👍

atavakoli:
@herom Done

herom:
thanks a ton @atavakoli 🥇

atavakoli:
My pleasure!

PR #977, opened 2017-10-02
Add function support for server addSoapHeader

atavakoli:
The purpose of this enhancement is to allow outgoing server headers to be
customized with information from the associated request. For servers only,
the `soapHeader` to addSoapHeader and changeSoapHeader can be a function,
which is called with the following arguments for each received request:

 - `methodName`     The name of the request method
 - `args`           The arguments of the request
 - `headers`        The headers in the request
 - `req`            The original request object

The return value of the function must be an Object({rootName: {name: 'value'}})
or strict xml-string, which will be inserted as an outgoing header of the
response to that request.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -352,7 +352,35 @@ They provide the following methods to manage the headers.
 
 #### *addSoapHeader*(soapHeader[, name, namespace, xmlns]) - add soapHeader to soap:Header node
 ##### Parameters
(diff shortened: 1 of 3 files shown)


herom:
Thanks a lot @atavakoli 👍

atavakoli:
@herom Thanks!

PR #978, opened 2017-10-06
replace non identifier chars to underscore

djjanse:
Hi vpulim, I have a wsdl with operation names like `prefix-methodname`. node-soap creates functions on the Client object with these names, but these can only be called like `client['prefix-methodname']()` instead of `client.prefix-methodname()`, because `-` is not allowed as identifier name in Javascript. While this works for node-style functions, this isn't working for Promise style functions. Bluebird doesn't promisify these invalid function names, so no `prefix-methodnameAsync` methods will be available.

I chose to change the invalid chars to `_` for both node-style and promise-style functions names. This is simpler and more explicit and consistent to the user than changing invalid characters only for promise-style functions.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -17,6 +17,8 @@ var HttpClient = require('./http'),
   BluebirdPromise = require("bluebird"),
   uuid4 = require('uuid/v4');
 
+var nonIdentifierChars = /[^a-z$_0-9]/i;
+
 var Client = function(wsdl, endpoint, options) {
   events.EventEmitter.call(this);
   options = options || {};
@@ -151,7 +153,6 @@ Client.prototype._defineService = function(service, endpoint) {
   return def;
 };
 
-var nonIdentifierChars = /[^a-z$_0-9]/i;
 Client.prototype._definePort = function(port, endpoint) {
   var location = endpoint,
     binding = port.binding,
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,6 +126,7 @@ Client.prototype._initializeServices = function(endpoint) {
 
 Client.prototype._initializeOptions = function(options) {
   this.streamAllowed = options.stream;
+  this.normalizeNames = options.normalizeNames;
   this.wsdl.options.attributesKey = options.attributesKey || 'attributes';
   this.wsdl.options.envelopeKey = options.envelopeKey || 'soap';
   this.wsdl.options.preserveWhitespace = !!options.preserveWhitespace;
@@ -160,8 +161,8 @@ Client.prototype._definePort = function(port, endpoint) {
     def = {};
   for (var name in methods) {
     def[name] = this._defineMethod(methods[name], location);
-    var normalizedName = name.replace(nonIdentifierChars, '_');
-    this[normalizedName] = def[name];
+    var methodName = this.normalizeNames ? name.replace(nonIdentifierChars, '_') : name;
+    this[methodName] = def[name];
   }
   return def;
 };
--- a/lib/client.js
+++ b/lib/client.js
@@ -151,6 +151,7 @@ Client.prototype._defineService = function(service, endpoint) {
   return def;
 };
 
+var nonIdentifierChars = /[^a-z$_0-9]/i;
 Client.prototype._definePort = function(port, endpoint) {
   var location = endpoint,
     binding = port.binding,
@@ -158,7 +159,8 @@ Client.prototype._definePort = function(port, endpoint) {
     def = {};
   for (var name in methods) {
     def[name] = this._defineMethod(methods[name], location);
-    this[name] = def[name];
+    var normalizedName = name.replace(nonIdentifierChars, '_');
+    this[normalizedName] = def[name];
   }
   return def;
 };
--- a/test/client-customHttp-test.js
+++ b/test/client-customHttp-test.js
@@ -91,7 +91,7 @@ it('should allow customization of httpClient and the wsdl file download should p
       });
     }
     //Now write the response with the wsdl
-    var state = httpResStream.write('HTTP/1.1 200 OK\r\nContent-Type: text/xml; charset=utf-8\r\nContent-Length: 1904\r\n\r\n'+wsdl);
+    var state = httpResStream.write('HTTP/1.1 200 OK\r\nContent-Type: text/xml; charset=utf-8\r\nContent-Length: 2497\r\n\r\n'+wsdl);
   });
 
   var httpCustomClient = new MyHttpClient({}, socketStream);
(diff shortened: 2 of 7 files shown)


herom:
Thanks a lot for your contribution @djjanse, we appreciate it a lot 😃 

Please see our guidelines [Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) and add at least 1 test in order to get your PR merged 👍

djjanse:
Hi @herom, it took me a while to get into the testing code. I added some tests. Forgot to read the Contribution.md before, my bad. 
I read that I also should create a branch in my fork and then commit to that branch. Should I create the changes again in a new fork, or is it ok to leave it like this?

lib/client.js
```diff
@@ -151,14 +151,16 @@ Client.prototype._defineService = function(service, endpoint) {
   return def;
 };
 
+var nonIdentifierChars = /[^a-z$_0-9]/i;
```
herom: Please, move the `var`-declaration to the top of the file, where all the other "file-wide variables" are defined.

herom:
> Should I create the changes again in a new fork, or is it ok to leave it like this?
It's good like it is, you don't have to create a new branch for this 👍

lib/client.js
djjanse: OK, moved it just above the constructor of `Client`.
Also fixed some JSHint warnings I introduced.

jsdevel:
so, I'm somewhat hesitant to merge this.  Consider this scenario:

```
some-thing
some:thing
```

is it contrived?  you bet, but it's possible.  I'm not sure we want potential overwriting of existing method names.

herom:
> I'm leaning towards hiding this behind an option, like normalizeNames: true or something.

If we use an option, would it make sense to say that it's a `String` instead of `Boolean` where the passed in option is used as the `replace`-character? Like `normalizeNames: '__'` and we use it to have `some-thing` normalized to `some__thing`?

jsdevel:
> If we use an option, would it make sense to say that it's a String instead of Boolean where the passed in option is used as the replace-character? Like normalizeNames: '__' and we use it to have some-thing normalized to some__thing?

I think we're getting a bit complicated with this.  Being that this hasn't come up as an issue before, I'm somewhat inclined to just reject this.  While it may be a pain to use brackets, it can be even more of a pain to get something wrong for a WSDL that doesn't really follow standard naming conventions.

djjanse:
> With this change `some-thing would` overwrite `some_thing` if both were defined. I'm not sure we want potential overwriting of existing method names.

I agree that methods shouldn't be silently overwritten.

> While it may be a pain to use brackets, it can be even more of a pain to get something wrong for a WSDL that doesn't really follow standard naming conventions.

A WSDL like this could be changed manually and be saved to a local file to follow standard naming conventions or just use JS identifier chars as a workaround, but then the WSDL should be changed every time it is updated. I think there should be a clean solution to consume WSDL's like this. Often the WSDL is a given and one doesn't have influence on it. (In my case I use the API of a third-party product).

If we use a `normalizeNames` option I should stick with a `Boolean`, document the possible overwriting of names, and log a warning or throw an error when this is the case. A more sophisticated solution with a `String` or maybe a `Map` (with per character translations) could be used if there's a need for it in the future.

herom:
> Being that this hasn't come up as an issue before, I'm somewhat inclined to just reject this. 

@jsdevel I would not want to reject this based on the fact that it wasn't an issue so far, as I guess it will be an issue, now that we have the `Promise`-support. See:
> While this works for node-style functions, this isn't working for Promise style functions. Bluebird doesn't promisify these invalid function names, so no prefix-methodnameAsync methods will be available.

So, I guess we can agree on using a new option called `normalizeNames` which does replace unqualified chars with `_` if set? @djjanse @jsdevel

jsdevel:
i think it should be a new option

djjanse:
I added the normalizeNames option. It took some time due to other (work-related) tasks. I didn't add a log or a throws when names are overwritten, as I found out that throwing from the `_definePort` method in `client.js` would need a change to the createClient method in `soap.js`. 

Besides that I found out that there are other cases where methods will be overwritten in the `_definePort` method. For example the `self_referencing.wsdl` contains two operations with the same name in two different bindings. This may be OK, but I'm not very experienced with WSDL definitions and don't know if these operations should be the same by definition.

test/client-customHttp-test.js
```diff
@@ -111,12 +111,6 @@ it('should allow customization of httpClient and the wsdl file download should p
               },
               output: {
               }
-            },
```
jsdevel: Why is this being removed?

test/wsdl/default_namespace.wsdl
```diff
@@ -24,12 +24,6 @@
       <wsdl:output message="OutputMessage">
     </wsdl:output>
     </wsdl:operation>
-    <wsdl:operation name="prefixed-MyOperation">
```
jsdevel: why is this being removed?

test/wsdl/default_namespace.wsdl
```diff
@@ -43,15 +37,6 @@
         <soap:body use="literal" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/>
       </wsdl:output>
     </wsdl:operation>
-    <wsdl:operation name="prefixed-MyOperation">
```
jsdevel: why is this being removed?

test/wsdl/default_namespace.wsdl
djjanse: In an earlier change I added unittests using a changed version of the existing `default_namespace.wsdl` with added operations that contained non identifier characters. I decided to move these to a new wsdl (`non_identifier_chars_in_operation.wsdl`) as to not possibly changing the behaviour of other tests (altough they didn't fail).

test/wsdl/default_namespace.wsdl
djjanse: In an earlier change I added unittests using a changed version of the existing `default_namespace.wsdl` with added operations that contained non identifier characters. I decided to move these to a new wsdl (`non_identifier_chars_in_operation.wsdl`) as to not possibly changing the behaviour of other tests (altough they didn't fail).

jsdevel: ideally we shouldn't be changing existing tests.  Period.  Can you revert this?

djjanse: @jsdevel: I think we agreed already: this change reverts my earlier commit https://github.com/vpulim/node-soap/pull/978/commits/e327a60cd70930c5e0fa258d5c2f737aa9a3fdbd

jsdevel: ahh, I see.  Thanks for clarifying @djjanse !

jsdevel:
@herom these LGTM, but I'd like your final review as well.

herom:
I'm fine with it 👍

PR #979, opened 2017-10-06
Do not explicitly set Connection-Header.

Mik13:
This commit is a follow up  to #974 

The Connection-Header should not be explicitly set, because the core modules are handling them (see https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js, everything connected to the `shouldKeepAlive` field).

This might be a breaking change, because connections are keep-alive by default (see https://github.com/nodejs/node/issues/10774) as long as there are running requests. The forever parameter, which sets the keep-alive flag for the agent, activates pooling (see https://github.com/nodejs/node/blob/2043944a4c62de60e7a04e8c914b565114c4b11b/lib/_http_agent.js#L78-L103), which offers a great performance gain on top.

I did not find a possibility to test this in node, because I've only got problems when connecting to SOAP-Servers written in another language, where every second request fails.

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -43,7 +43,7 @@ HttpClient.prototype.buildRequest = function(rurl, data, exheaders, exoptions) {
     'Accept': 'text/html,application/xhtml+xml,application/xml,text/xml;q=0.9,*/*;q=0.8',
     'Accept-Encoding': 'none',
     'Accept-Charset': 'utf-8',
-    'Connection': 'close',
+    'Connection': !!exoptions.forever ? 'keep-alive' : 'close',
     'Host': host + (isNaN(port) ? '' : ':' + port)
   };
   var attr;
--- a/Readme.md
+++ b/Readme.md
@@ -441,7 +441,7 @@ The `args` argument allows you to supply arguments that generate an XML document
 The `options` object is optional and is passed to the `request`-module.
 Interesting properties might be:
 * `timeout`: Timeout in milliseconds
-* `forever`: Enables keep alive connections
+* `forever`: Enables keep-alive connections and pools them
 
 ### Client.*method*Async(args) - call *method* on the SOAP service.
 


jsdevel:
> This might be a breaking change, because connections are keep-alive by default

If that's the case, are sure sure we know _exactly_ what the repercussions are?  Would it be possible to have an option instead?  Like `connection: 'keep-alive'`?

jsdevel:
what issue did this solve for you?

Mik13:
I communicate with a Java SOAP-Service and without this change, every second request fails, because the service closed the connection, but the https module got this connection still in its pool after the first successful request (because with #974 pooling is possible, but pooling closed connections does not make any sense).

After my change, the connection will not be closed by the counterpart and our pooled connection can be reused correctly, means every request works again.

jsdevel:
> I communicate with a Java SOAP-Service and without this change, every second request fails, because the service closed the connection, but the https module got this connection still in its pool after the first successful request (because with #974 pooling is possible, but pooling closed connections does not make any sense).

I see.  Well, being that most connections are keep-alive anyway, I think I'm comfortable merging this knowing that it could be a breaking change.  Can you document this in the README?

Mik13:
I've added the change to the readme

lib/http.js
```diff
@@ -43,7 +43,8 @@ HttpClient.prototype.buildRequest = function(rurl, data, exheaders, exoptions) {
     'Accept': 'text/html,application/xhtml+xml,application/xml,text/xml;q=0.9,*/*;q=0.8',
     'Accept-Encoding': 'none',
     'Accept-Charset': 'utf-8',
-    'Connection': 'close',
+    // Connection-Header is set in the http or https core module
+    // 'Connection': !!exoptions.forever ? 'keep-alive' : 'close',
```
jsdevel: is this meant to be commented out?

Mik13: See: https://github.com/vpulim/node-soap/pull/979#issuecomment-335083728

Should I add "next line enables keep-alive connection only for pooled connections" to clearify that?

jsdevel: I'd replace both lines with `'Connection': !!exoptions.forever ? 'keep-alive' : 'close',` and call it good.

Mik13:
Alright, I've changed that and now the old default behaviour is restored and the forever option will now pool the keep alive connections.

jsdevel:
Thanks @Mik13 !

Mik13:
Thank you too!

PR #981, opened 2017-10-10
Fixed checking for empty obj.Body before further actions

vlad-rz:
Moved check for empty obj.Body before further actions with it and before checking for type of obj.Body
It throws an error even if response was successfull, but without obj.Body.


herom:
Your PR breaks 6 tests @vlad-rz, please take a look at it over at Travis or run `npm test` on your local machine in order to see what failed 👓

jsdevel:
closing due to inactivity

PR #983, opened 2017-10-16
only detect xsi:nil if its value is `true`

wvanderdeijl:
we had an example where the SOAP service outputs xml with `xsi:nil="false"` which was interpreted by node-soap as having a xsi:nil attribute and thus removing the element from the response. This PR makes the check more strict and verifies that xsi:nil is actually true


jsdevel:
@wvanderdeijl  This is great!  Can you remove `package-lock.json`?

wvanderdeijl:
oops. I missed your comment. Just removed the package-lock so this PR should be good to go

jsdevel:
Lame.  Looks like npm doesn't support node version `v9.2.0`

npm

Going to merge this PR anyway as the other builds passed.  Thanks for the contribution @wvanderdeijl !

PR #984, opened 2017-10-17
Make `HttpClient#buildRequest` headers and options parameter mandatory

Mik13:
With #979 I sadly introduced a bug, because the two parameters are optional, despite the fact that they are not marked as such in the jsdoc comment.

This commit makes them mandatory and should fix the tests.

### DIFF
--- a/lib/soap.d.ts
+++ b/lib/soap.d.ts
@@ -159,7 +159,7 @@ export function listen(server: any, options: IServerOptions): Server;
 
 export class HttpClient {
     constructor(options?: IOptions);
-    buildRequest(rurl: string, data: any | string, exheaders?: { [key: string]: any }, exoptions?: { [key: string]: any }): any;
+    buildRequest(rurl: string, data: any | string, exheaders: { [key: string]: any }, exoptions: { [key: string]: any }): any;
     handleResponse(req: any, res: any, body: any | string): any | string;
     request(rurl: string, data: any | string, callback: (err: any, res: any, body: any | string) => void, exheaders?: { [key: string]: any }, exoptions?: { [key: string]: any }): any;
     requestStream(rurl: string, data: any | string, exheaders?: { [key: string]: any }, exoptions?: { [key: string]: any }): any;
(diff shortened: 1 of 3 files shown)


lib/soap.d.ts
```diff
@@ -159,7 +159,7 @@ export function listen(server: any, options: IServerOptions): Server;
 
 export class HttpClient {
     constructor(options?: IOptions);
-    buildRequest(rurl: string, data: any | string, exheaders?: { [key: string]: any }, exoptions?: { [key: string]: any }): any;
+    buildRequest(rurl: string, data: any | string, exheaders: { [key: string]: any }, exoptions: { [key: string]: any }): any;
```
Mik13: Sync ts with jsdoc

lib/wsdl.js
```diff
@@ -2218,8 +2218,8 @@ function open_wsdl(uri, options, callback) {
 
   // initialize cache when calling open_wsdl directly
   var WSDL_CACHE = options.WSDL_CACHE || {};
-  var request_headers = options.wsdl_headers;
-  var request_options = options.wsdl_options;
+  var request_headers = options.wsdl_headers || {};
+  var request_options = options.wsdl_options || {};
```
Mik13: Those are the only instances where the two parameter may become falsy

lib/soap.d.ts
jsdevel: I'm a bit confused.  They weren't mandatory before.  Why should they be mandatory now?

Mik13: Why not. The jsdoc does not mark them as optional, and there is only one instance, which does not set them to an empty object like all the others. So if we just set them to an empty object by default in this one instance, we can make the parameters mandatory and do not have to handle a falsy value in this function, thus taking complexity out of it

jsdevel: Documentation is one thing.  If the implementation had them as optional in the past, then I'm of the opinion that we should update the documentation i.e. ts file or jsdoc.

jsdevel: Is this even a method that most users of `node-soap` would use?  If not, then I'm fine with this change.

jsdevel:
Being that these have been optional for a while, I'm deciding that we shouldn't make them required now.  It is possible that others have been interfacing with the HttpClient directly.

I updated master so the tests should pass now.  If you'd like to update the JSDoc/tsdoc to mark the parameters as optional feel free.

Mik13:
Works for me, thank you!

PR #985, opened 2017-10-18
parse double and floats

leon:
When type is xs:double or xs:float we use parseFloat to convert it to javascript.

I'm unsure why this hasn't been done.
Maybe you have a good reason as to why you are not converting floats or doubles.
But since usually people want the data in their native form, I thought I'd suggest this atleast.

Tests:
There were 28 failing tests in master, so I was unable to write any tests.
I'd gladly write tests once master is passing again.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1450,6 +1450,8 @@ WSDL.prototype.xmlToObject = function(xml, callback) {
     else {
       if (name === 'int' || name === 'integer') {
         value = parseInt(text, 10);
+      } else if (name === 'double' || name === 'float') {
+        value = parseFloat(text);
       } else if (name === 'bool' || name === 'boolean') {
         value = text.toLowerCase() === 'true' || text === '1';
       } else if (name === 'dateTime' || name === 'date') {


herom:
thanks a lot @leon - it seems that #979 broke our build when it was merged in... somehow Travis CI/Coveralls wasn't running 🤕 

I hope we get this fixed very soon - seems like #984 is a follow up to the breaking PR which will fix these issues.... I'll notify you once we're up and running again 💯

jsdevel:
@leon I've updated master with a fix for the bug introduced from #979.  Can you rebase and update your branch?

herom:
@leon I restarted Travis CI and all current tests pass - could you please add tests for your PR? 👍

leon:
Where do we test all the other serializations?
I couldn't find where we test all the simple types.

cgilindellmo:
Could you also add parsing of xs:long and xs:decimal ?

alexb-uk:
I've just come across this issue as didn't realise that type "double" wasn't being converted.

Can I help get this PR merged? Given how old this PR is I'm guessing @leon has probably moved on to something else?

Do you want me to submit a new PR with some tests?

alexb-uk:
Having said that I think it's safer to convert using `Number(text)`.

jsdevel:
need a test added and conflicts resolved

PR #986, opened 2017-11-09
Fixed checking for empty obj.Body before further actions

raviolin:
An error is thrown even if response was succesful.

This is the same as #981 only this doesn't break tests. I decided to make a new pull request because it have been almost a month since #981 was created, and it's author doesn't seem willing to fix tests.


raviolin:
I swear it passed all tests on my local machine, but Travis failed to `npm install`
https://travis-ci.org/vpulim/node-soap/jobs/299410819

herom:
@raviolin sorry for letting you down for this long... I had some time off and lost sight of this issue. Could you please try and rebase against the latest `master`?

I updated our `.travis.yaml` file in order to test against the latest LTS instead of the latest available version (which seems to fail the tests) - now it should run as expected.

herom:
Thanks a lot @raviolin 👍 

In order to get your PR pushed I'll have to ask for at least 1 test for your contribution as requester per our [Guidelines on Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request)

jsdevel:
why check for `obj.html` anyway?  Does this lib _ever_ use `obj.html`?  Seems we could just move the if statement up and call this good.

raviolin:
@jsdevel apparently it's necessary in case the client gets an HTML response. There are tests for that.

@herom I'll add a test with a response sample, please don't close this PR yet.

jsdevel:
LGTM

raviolin:
Can you remove 'needs tests' label? Or should I add more tests?

herom:
@raviolin I don't know why the label is still there and we haven't merged your PR yet. 

Sorry for that... again 🐑 🐑 🐑 😞

PR #987, opened 2017-11-12
Namespace for referenced element was wrong.

blissi:
Hallo,
I found a bug with a wrong namespace for a referenced element. I think this only occurs under special circumstances - please have a look at the unit test that I've included. The bug was that the element "name2" didn't get the namespace prefix "cis:" and therefore the XML file was invalid.

There is still another bug in the library, though:
The "streetName"-node should also have the "cis:"-prefix, but that one is also omitted. That's the reason why the unit test still fails.

I tried to find out the cause of this bug but haven't been successful so far. I'll try again tomorrow, but if anybody has an idea where I have to look please give me a note.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1824,7 +1824,7 @@ WSDL.prototype.objectToXML = function(obj, name, nsPrefix, nsURI, isFirst, xmlns
               var resolvedChildSchemaObject;
               if (childSchemaObject.$type) {
                 var typeQName = splitQName(childSchemaObject.$type);
-                var typePrefix = typeQName.prefix;
+                var typePrefix = childSchemaObject.$baseNameSpace || typeQName.prefix;
                 var typeURI = schema.xmlns[typePrefix] || self.definitions.xmlns[typePrefix];
                 childNsURI = typeURI;
                 if (typeURI !== 'http://www.w3.org/2001/XMLSchema' && typePrefix !== TNS_PREFIX) {
@@ -2057,7 +2057,6 @@ WSDL.prototype.findChildSchemaObject = function(parameterTypeObj, childName, bac
 
           if (found) {
             found.$baseNameSpace = childNameSpace;
-            found.$type = childNameSpace + ':' + childName;
             break;
           }
         }
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -1769,7 +1769,7 @@ WSDL.prototype.objectToXML = function(obj, name, nsPrefix, nsURI, isFirst, xmlns
             //find sub namespace if not a primitive
             if (childSchemaObject &&
               ((childSchemaObject.$type && (childSchemaObject.$type.indexOf('xsd:') === -1)) ||
-                childSchemaObject.$ref || childSchemaObject.$name)) {
+                childSchemaObject.$ref || childSchemaObject.resolvedNameWithPrefix || childSchemaObject.$name)) {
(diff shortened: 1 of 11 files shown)


jsdevel:
@blissi this is great!  Looks like the test is failing.  Looks like elements names are different in the schema?

blissi:
Have a look at my unit test update I just checked in:

I split the test into two tests:
1. doAction__should_handle_namespaces_of_referenced_element_correctly: here the bug in the library was that name2 was not prefixed with "cis:" => I have fixed this bug in my previous checkin.

2. doAction__should_handle_namespaces_in_referenced_complex_type_correctly: this is the other bug I'm still struggling with. streetName is a member of the complexType "PersonAddressType" and should also have the prefix "cis:".

jsdevel:
Hmm, it looks like you're using element names that aren't defined in your schema.  Before your last update, `cis:name` appeared in the failed assertion, now only `name` appears in the assertion.

blissi:
@jsdevel I don't fully understand...in the new unit test,
`<name>My name</name>`
is correct since it is defined under <Person> and thus has the namespace bcs. It's the streetName that isn't correct => the "cis:"-prefix is missing in the generated request and therefore it also mistakenly inherits the bcs-namespace.

jsdevel:
In your first commit you had ref cis for name in your schema.  Build output used cis namespace for the name elements.  Now you're just defining type xs:string for name and streetName.  What does the spec say about namespace behavior?

blissi:
Yes, I changed it when adjusting the unit tests for the second checkin. I did it on purpose:
I wanted to see in the result-XML of the unit test that the name below `<Person>` has the correct namespace (omitted here because inherited). The name below `<Address>` on the other hand must have the namespace. Sorry for the confusion.

I don't know where this is officially specified. I found this bug because the webservice I am calling is rejecting the request. When I specify the namespace prefix, it works. The webservice I am calling seems to behave correctly: when I use the WSDL in SoapUI it also creates a sample request where the "cis:"-prefix is used at the name below `<Address>`.

I think I am quite close to a solution...I have to dig deeper in the evening and try some more things. I'll keep you informed.

blissi:
@jsdevel  The bug is fixed now.

blissi:
My last fix had a severe regression: when one SOAP operation was called twice, the second call failed because I had manipulated the $name and therefore findChildSchemaObject() couldn't find that schema object anymore.

herom:
Thanks a lot for your contriubtion @blissi - could you please rebase against the latest `master` as I updated our `.travis.yaml` so that the latest LTS version of node is used for testing

jsdevel:
@blissi any chance you can clean this PR up?

herom:
ping @blissi - if you could take some time to clean up and rebase your PR we'll happily merge it 👍 :trollface:

blissi:
@herom  I'm not a git expert - how do I do this?

herom:
@blissi just take a look into our section on [Submitting a Pull Request](https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md#submitting-a-pull-request) of our (CONTRIBUTING.md)[https://github.com/vpulim/node-soap/blob/master/CONTRIBUTING.md] file. There are 2 links to the (very good) GitHub Help Pages on this topic.

kapilkarda:
This pull request is still not fixed ?

jsdevel:
closing due to inactivity

PR #988, opened 2017-11-28
Enable multiArgs during promisification

rmobis:
Because the `multiArgs` parameters wasn't enabled when promisification was implemented on #956, it is impossible to have access to the `raw` and `soapHeader` parameters that are passed when the response is received. Enabling `multiArgs` causes the promisified function to return an array with all but the first callback parameters, solving the issue.

**Example `multiArgs: false`**
```
let response = await client.someMethodAsync();
console.log(response); // { attr1: value1, attr2: value2 }
```

**Example `multiArgs: true`**
```
let response = await client.someMethodAsync();
console.log(response); // [{ attr1: value1, attr2: value2 }, '<xml>...</xml>', soapHeaders]
```

Please note, **this is a breaking change**.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -448,7 +448,10 @@ Interesting properties might be:
 
 ``` javascript
   client.MyFunctionAsync({name: 'value'}).then((result) => {
+    // result is a javascript array containing result, raw and soapheader
     // result is a javascript object
+    // raw is the raw response
+    // soapHeader is the response soap header as a javascript object
   })
 ```
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -26,8 +26,11 @@ var Client = function(wsdl, endpoint, options) {
   this._initializeOptions(options);
   this._initializeServices(endpoint);
   this.httpClient = options.httpClient || new HttpClient(options);
-  var suffixOption = options.overridePromiseSuffix ? { suffix: options.overridePromiseSuffix } : null;
-  BluebirdPromise.promisifyAll(this, suffixOption);
+  var promiseOptions = { multiArgs: true };
+  if (options.overridePromiseSuffix) {
+    promiseOptions.suffix = options.overridePromiseSuffix;
+  }
+  BluebirdPromise.promisifyAll(this, promiseOptions);
 };
 util.inherits(Client, events.EventEmitter);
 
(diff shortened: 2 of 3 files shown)


rmobis:
Updated it with tests. Instead of writing specific tests for the functionality, I just extended `request-response-samples-test.js` files to run all those tests with the promisified version as well. 

I would have liked to also port the _Method invocation_ tests under `client-test.js` but for some strange reason whenever I call the promisified method I get a DNS error when trying to lookup for the WSDL endpoint.

I assume this is either due to the custom request handler or some other black magic I couldn't figure out. Either way, if you think those are necessary as well, I'd need at least some pointers to the correct path forward.

rmobis:
Any update on this? Is there anything missing to get this merged?

jsdevel:
Sorry for the wait had to restart one of the node builds.  Thanks!

rmobis:
No problem. Thanks for the merge!

mbrevda:
Thanks for this! I realize this lib is at < v1 and hence is like to break at any time. If possible however, it would be helpful if breaking changes like this would be more explicit in the release notes (this one bit me until I went through all the changes).

jsdevel:
@mbrevda we apologize for any inconvenience you may have had.  If you have time,  can you document what you had to change and what version of node-soap you were on?

mbrevda:
No need to apologize! I was on v0.23. 

First I tried to see if `multiArgs` could be passed as an option to `soap.createClientAsync`, and simply disable it, but it doesn't seem to.

Instead, I'm now destructuring the response. 

Before:
```js
let res = await client.methodAsync(opts)
```

After:
```js
let [res] = await client.methodAsync(opts)
```

jsdevel:
hmm, I think this is pretty bad.  Thoughts @herom ?

mbrevda:
Two thoughts: A. if your following semver, breaking changes in `v < 1.0` is ok. Im simply recommending that it be labeled as such. B. Kinda late to fix this now, as "fixing" it will just break it for others

herom:
I don't think it would be a good idea to "just revert" it, but maybe we could make it configurable through the passed in options, so that `multiArgsPromise` (?) could be set through the consumer, just like `overridePromiseSuffix`?

jsdevel:
I think that's a good option @herom

jsdevel:
@rmobis can you implement the feedback herein?  We'll have to revert unless the functionality can be behind an option as @herom mentioned above.

bat020:
Just had a v similar experience to @mbrevda – upgraded from v0.23.0 to v0.25.0 and was caught out by this change. Having said that, this is clearly a useful enhancement and shouldn't be reverted. Fixing the problem at our end was trivial (we're also destructuring the response). But +1 to calls for breaking changes like this to be flagged up a bit more clearly in the readme and history files.

PR #992, opened 2017-12-14
feat(client): add rawRequest to callback arguments

forstermatth:
__Use case:__ Allows clients who send requests asynchronously to accurately log the requests they are making.

Adds the XML raw request as an argument to the client callback. This allows correct tracking of the requests, as requests made asynchronously on the same client will overwrite the `client.lastRequest` attribute.

I added a test to the headers describe block as that is where it seemed similar code was being tested.


jsdevel:
@herom what are your thoughs?

herom:
@jsdevel @forstermatth This seems like a really useful addition to our library - thanks a lot 👍

PR #993, opened 2018-01-04
Update Readme.md

tbranko:
Option to use pfx self enclosed certificate (p12) already exists in the code, and it works. It just wasn't documented.


jsdevel:
Thanks!

PR #1000, opened 2018-02-14
Error on custom deserializer example

mrganser:
Probably a typo, there is an assignment operator inside the wsdlOptions object where a colon is expected.


herom:
a typo for sure, thanks a lot @mrganser 👍

PR #1002, opened 2018-03-01
Implemented optional asynchronous authentication

schantaraud:
Changes:
* The server.authentication function can either return a value or call an optional callback to allow for authentication from databases or other webservices. If both are used, the first one wins, the second is ignored.
* In case of authentication error, a SOAP Fault is returned with a status code of 401 (previously: standard exception with optional stacktrace, which soapClient doesn't like).

All existing tests results unchanged, new tests: test/server-authentication-test.js.

All feedback is welcome.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -294,6 +294,26 @@ To change the HTTP statusCode of the response include it on the fault.  The stat
 
 If `server.authenticate` is not defined then no authentication will take place.
 
+Asynchronous authentication:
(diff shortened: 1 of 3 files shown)


jsdevel:
@schantaraud the node 4 tests are failing.  can you take a look?

jsdevel:
nvm.  node 4 is dead

PR #1004, opened 2018-03-27
Add one-way response configuration options

leachiM2k:
Some clients, for instance SAP SOAP Client, expect different behavior when calling one-way operations. For example SAP expects a 202 status code with no body or an empty body with a status code 200.
This little change makes it possible to configure both options.

This is my first contribution to this project, so please give me a constructive feedback ;-)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -20,6 +20,7 @@ This module lets you connect to web services using SOAP.  It also provides a ser
   - [Options](#options)
   - [Server Logging](#server-logging)
   - [Server Events](#server-events)
+  - [Server Response on one-way calls](#server-response-on-one-way-calls)
   - [SOAP Fault](#soap-fault)
   - [Server security example using PasswordDigest](#server-security-example-using-passworddigest)
   - [Server connection authorization](#server-connection-authorization)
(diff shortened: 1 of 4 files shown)


Readme.md
```diff
@@ -258,6 +259,16 @@ Server instances emit the following events:
 The sequence order of the calls is `request`, `headers` and then the dedicated
 service method.
 
+### Server Response on one-way calls
+
+The so called one-way (or asynchronous) calls occur when an operation is called with no output defined in WSDL.
+The server sends an response to the client disregarding the result of the operation.
```
jsdevel: nit: `... sends a response`

lib/server.js
```diff
@@ -39,6 +39,7 @@ var Server = function (server, path, services, wsdl, options) {
   this.wsdl = wsdl;
   this.suppressStack = options && options.suppressStack;
   this.returnFault = options && options.returnFault;
+  this.onewayOptions = options && options.oneWay;
```
jsdevel: Why not add ` || {}` here so `onewayOptions` is _always_ an object?  Should reduce typing elsewhere.

leachiM2k:
@jsdevel I've adjusted the code to your review remarks and added two unit tests. I hope this is fine now.

jsdevel:
@leachiM2k can you address this? https://github.com/vpulim/node-soap/pull/1004/files#r177319579

leachiM2k:
@jsdevel after your review, I've implemented the initialization in the method `_initializeOptions`. I left line 42 as it was to match the style of other lines in the constructor.

lib/server.js
jsdevel: Please change this to `this.onewayOptions = options && options.oneWay || {};`.  We can follow up later to change existing style.

Further down in this file, you can assume it'll always be an object.

jsdevel:
@leachiM2k see my comment https://github.com/vpulim/node-soap/pull/1004#discussion_r178121914

leachiM2k:
@jsdevel I've updated the code line. Hope it's ok right now.

jsdevel:
Thanks @leachiM2k !

PR #1006, opened 2018-04-04
Change signature of authorizeConnection

angeloassis13:
Change signature of authorizeConnection to allow changing response parameters, for example, setting status code 401 (Unauthorized)


jsdevel:
thanks!

PR #1007, opened 2018-04-16
Fix cert with auth

eliranha:


jsdevel:
@eliranha all the builds are failing.  Can you take a look?

PR #1009, opened 2018-06-12
Added bluebird types on devDependencies

LsKoder:
Just import bluebird types on devDependencies to fix a bug occurred importing node-soap on ts projects:

> error TS7016: Could not find a declaration file for module 'bluebird'. '/node_modules/bluebird/js/release/bluebird.js' implicitly has an 'any' type.
>  Try `npm install @types/bluebird` if it exists or add a new declaration (.d.ts) file containing `declare module 'bluebird';`
> 
> 4 import * as BluebirdPromise from 'bluebird';

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,22 @@
 {
   "name": "soap",
-  "version": "0.23.0",
+  "version": "0.24.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@types/bluebird": {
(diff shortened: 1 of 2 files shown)


herom:
Please also add the `package-lock.json`, as otherwise all the tests will fail... 👍

herom:
Thanks a lot @LsKoder 👍 

As all the checks/tests pass I think we're good to go with the other upgrades in there too :trollface:

PR #1013, opened 2018-07-16
Add attributes parsing

nastassja90:
Hi,

this PR adds `$attributes` property to element description. That property is not enumerable in order to preserve the back compatibility.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -216,6 +216,8 @@ var ComplexTypeElement = Element.createSubClass();
 var ComplexContentElement = Element.createSubClass();
 var SimpleContentElement = Element.createSubClass();
 var SequenceElement = Element.createSubClass();
+// attribute
+var AttributeElement = Element.createSubClass();
 var AllElement = Element.createSubClass();
 var MessageElement = Element.createSubClass();
 var DocumentationElement = Element.createSubClass();
@@ -240,10 +242,12 @@ var ElementTypeMap = {
   choice: [ChoiceElement, 'element sequence choice any'],
     // group: [GroupElement, 'element group'],
   enumeration: [EnumerationElement, ''],
-  complexType: [ComplexTypeElement,  'annotation sequence all complexContent simpleContent choice'],
+  complexType: [ComplexTypeElement,  'annotation sequence all complexContent simpleContent choice attribute'],
   complexContent: [ComplexContentElement,  'extension'],
   simpleContent: [SimpleContentElement,  'extension'],
   sequence: [SequenceElement, 'element sequence choice any'],
+  // attributes
+  attribute: [AttributeElement, ''],
   all: [AllElement, 'element choice'],
 
   service: [ServiceElement, 'port documentation'],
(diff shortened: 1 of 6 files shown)


herom:
@nastassja90 are there some more commits to come or is your PR "feature complete" ?

nastassja90:
Hi herom, 
yes we have finished you can proceed with the merge, thank you

lib/wsdl.js
```diff
@@ -807,17 +811,41 @@ EnumerationElement.prototype.description = function() {
...
+      isFirstChild = true;
+      ret = child.description(definitions, xmlns);
     }
   }
-  return {};
+
+  if(hasAttributes){
+    Object.defineProperty(ret, '$attributes',{
+      value: $attributes,
+      enumerable: false,
```
herom: Why would we want this to not be enumerable and therefore not appearing when calling `Object.keys()`, `Object.values()` and similar on this property?

lib/wsdl.js
```diff
@@ -870,13 +898,24 @@ ElementElement.prototype.description = function(definitions, xmlns) {
...
         }
         else {
           Object.keys(description).forEach(function (key) {
             elem[key] = description[key];
           });
+          
+          
+          if(description.$attributes){
+            Object.defineProperty(elem, '$attributes',{
+              value: description.$attributes,
```
herom: The same question appears as above, why do we want this to be non-enumerable?

lib/wsdl.js
```diff
@@ -216,6 +216,8 @@ var ComplexTypeElement = Element.createSubClass();
 var ComplexContentElement = Element.createSubClass();
 var SimpleContentElement = Element.createSubClass();
 var SequenceElement = Element.createSubClass();
+// attribute
```
herom: I guess we don't need the comment here, as the name of the `var` speaks for itself 😺

lib/wsdl.js
```diff
@@ -240,10 +242,12 @@ var ElementTypeMap = {
   choice: [ChoiceElement, 'element sequence choice any'],
     // group: [GroupElement, 'element group'],
   enumeration: [EnumerationElement, ''],
-  complexType: [ComplexTypeElement,  'annotation sequence all complexContent simpleContent choice'],
+  complexType: [ComplexTypeElement,  'annotation sequence all complexContent simpleContent choice attribute'],
   complexContent: [ComplexContentElement,  'extension'],
   simpleContent: [SimpleContentElement,  'extension'],
   sequence: [SequenceElement, 'element sequence choice any'],
+  // attributes
```
herom: same here 👍

lib/wsdl.js
nastassja90: I wanted to limit as much as possible side-effects in order to manage attributes just in that piece of code.

herom: I'd vote for the property to be `enumerable` so that the fellow developer is able to read it within a loop or something the like. What about you @jsdevel ?

herom: @nastassja90 would you please update your PR and make it `enumerable`? this would be awesome 👍

starsoccer:
Is there any work being done to fix this? From the above reference issue on wsdl-tsclient this blocks auto generating correct types which makes using the client more or less pointless if there are attributes since all typing will be lost.

jsdevel:
the community needs to provide a PR or pick this one back up.

icholy:
@jsdevel are you still the maintainer around here? I'm willing to pick up this work, but I want to make sure there will be someone to look at it.

roopakv:
@jsdevel or @herom I'd love to pick this up if you all would be willing to look at, let me know and I'm happy to send an updated PR

PR #1014, opened 2018-07-19
soap client describe() resolves namespace cross references between schema

PaoloSarti:
Hi,

this PR adds support to cross references in namespaces between types from different schema.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -870,7 +870,7 @@ ElementElement.prototype.description = function(definitions, xmlns) {
   if (type) {
     type = splitQName(type);
     var typeName = type.name,
-    ns = xmlns && xmlns[type.prefix] || /* this.schemaXmlns[type.prefix] || */ definitions.xmlns[type.prefix],
+    ns = xmlns && xmlns[type.prefix] || (definitions.xmlns[type.prefix] !== undefined && this.schemaXmlns[type.prefix]) || definitions.xmlns[type.prefix],
     schema = definitions.schemas[ns],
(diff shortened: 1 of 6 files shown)


lib/wsdl.js
```diff
@@ -116,6 +117,12 @@ var Element = function(nsName, attrs, options) {
       }
     }
   }
+  for (var schemaKey in schemaAttrs) {
+    var schemaMatch = /^xmlns:?(.*)$/.exec(schemaKey);
+    if (schemaMatch && schemaMatch[1] && schemaMatch) {
```
herom: did you really want to check for `schemaMatch` twice here? 😄

lib/wsdl.js
```diff
@@ -855,7 +862,9 @@ ElementElement.prototype.description = function(definitions, xmlns) {
   if (type) {
     type = splitQName(type);
     var typeName = type.name,
-      ns = xmlns && xmlns[type.prefix] || definitions.xmlns[type.prefix],
+      ns = xmlns && xmlns[type.prefix] ||
+        (definitions.xmlns[type.prefix] !== undefined && this.schemaXmlns[type.prefix]) ||
```
herom: when checking for `!== undefined` explicitly, we are aware that `null` would pass - is this intentional?

lib/wsdl.js
PaoloSarti: fixed :)

lib/wsdl.js
PaoloSarti: Anyway, I could change `definitions.xmlns[type.prefix] !== undefined` into `definitions.xmlns[type.prefix]`

herom: >  I checked the presence or absence of the property

good catch, I didn' t think of that... 👍

PR #1015, opened 2018-07-21
Improve deserialization of inline simpleType declarations

kouak:
I encountered an issue deserialization of elements declared in such way :
```xml
<xs:element name="nonICAOAerodromeOfDeparture" minOccurs="0" maxOccurs="1">
  <xs:simpleType>
    <xs:restriction base="xs:boolean">
    </xs:restriction>
  </xs:simpleType>
</xs:element>
```

I'm no WSDL/XSD expert but I believe this is equivalent to this syntax :
```xml
<xs:element name="nonICAOAerodromeOfDeparture" minOccurs="0" maxOccurs="1" type="xs:boolean">
</xs:element>
```
While node-soap correctly deserializes the latter to a javascript boolean, the former is deserialized to a string (i.e `"true"` or `"false"`).

Furthermore, the lack of `type` attribute on the `xs:element` declaration makes it impossible to use `customDeserializer` to properly handle such fields.

This PR fixes both aspects by:
* Allowing `xs:element` to have a `xs:simpleType` child
* Computing `xs:element` type if missing from child `xs:simpleType`'s `name` attribute, or `restriction`'s `base` attribute

This PR was inspired by #949, and aims to achieve the same effect, without breaking existing tests and preserving the ability to use `customDeserializer` if needed.

See attached test case for the complete behaviour.

NB: This PR includes ES6 syntax and was tested with node v6.5.0 and node v8.9.3 locally.


herom:
LGTM, what about you @jsdevel ?

herom:
thanks a lot for your contribution @kouak , we appreciate it a lot 👍 🚀

kouak:
@herom Any chance to get this released ?

We're currently wrapping a large SOAP API in a JS lib which uses those weird definitions. Having this sooner rather than later would save us the trouble to refactor call sites later on.

If this is not much of a hassle for the maintainers, this would be greatly appreciated !

herom:
@kouak I now we've been a lazy with releasing new versions, sorry for that - I will take a look at it tomorrow as I'm a bit tight on schedule 👍

kouak:
@herom: Thanks a lot for the quick release !

herom:
You're welcome - although I wouldn't call my "response time" quick 😸 - sorry for that.

PR #1017, opened 2018-08-01
WSSecurityCert without Timestamp and only xml-exc-c14n#

RaymondKroon:
For a client we are writing a Timestamp field is not allowed, and also http://www.w3.org/2000/09/xmldsig#enveloped-signature is not supported. Therefore I made some changes to make both configurable: { hasTimeStamp: false, signatureTransformations: [] }

It should have no breaking changes, and should be fully backwards compatible.

### DIFF
--- a/lib/security/WSSecurityCert.js
+++ b/lib/security/WSSecurityCert.js
@@ -44,7 +44,7 @@ function WSSecurityCert(privatePEM, publicP12PEM, password, options) {
     passphrase: password
   };
   this.x509Id = "x509-" + generateId();
-  this.hasTimeStamp = typeof options.hasTimeStamp === 'undefined' ? true : !!options.hasTimestamp;
+  this.hasTimeStamp = typeof options.hasTimeStamp === 'undefined' ? true : !!options.hasTimeStamp;
   this.signatureTransformations = Array.isArray(options.signatureTransformations) ? options.signatureTransformations
     : ["http://www.w3.org/2000/09/xmldsig#enveloped-signature", "http://www.w3.org/2001/10/xml-exc-c14n#"];
 
--- a/lib/security/WSSecurityCert.js
+++ b/lib/security/WSSecurityCert.js
@@ -44,8 +44,8 @@ function WSSecurityCert(privatePEM, publicP12PEM, password, options) {
     passphrase: password
   };
   this.x509Id = "x509-" + generateId();
-  this.addTimestamp = options.addTimestamp !== undefined ? options.addTimestamp : true;
-  this.signatureTransformations = options.signatureTransformations !== undefined ? options.signatureTransformations
+  this.hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true;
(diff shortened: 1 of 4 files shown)


lib/security/WSSecurityCert.js
```diff
@@ -43,6 +44,9 @@ function WSSecurityCert(privatePEM, publicP12PEM, password) {
     passphrase: password
   };
   this.x509Id = "x509-" + generateId();
+  this.hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true;
```
herom: why do we need the `||`? wouldn't it be the same, if we'd just take the right side of the `or`?

RaymondKroon: Ah, ok. 
To make those situations work, the left side of the check is needed, however it should be better defined, because:
```js
const options = {
  hasTimeStamp: 0
};
_hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true; 
console.log(_hasTimeStamp); // true
```
which should be false.

```js
_hasTimeStamp = options.hasTimeStamp !== 'undefined' ? !!options.hasTimeStamp : true;
``` 
should work ...

herom: The following snippet will do the trick, transforming falsy/truthy values and leave `boolean` values untouched...

```javascript
const options = {
  hasTimestamp: 0
};

const hasTimestamp = !!options.hasTimestamp;
```

RaymondKroon: Ok, changed it.

lib/security/WSSecurityCert.js
```diff
@@ -44,8 +44,8 @@ function WSSecurityCert(privatePEM, publicP12PEM, password, options) {
     passphrase: password
   };
   this.x509Id = "x509-" + generateId();
-  this.hasTimeStamp = options.hasTimeStamp || typeof options.hasTimeStamp === 'boolean' ? !!options.hasTimeStamp : true;
-  this.signatureTransformations = options.signatureTransformations && Array.isArray(options.signatureTransformations) ? options.signatureTransformations
+  this.hasTimeStamp = typeof options.hasTimeStamp === 'undefined' ? true : !!options.hasTimestamp;
```
herom: I'm sorry to be a bit picky here, but I think we should stuck with either `hasTimeStamp` or `hasTimestamp` in order to have it consistent between the `options` and the internal value. I'd vote for `hasTimestamp` as I haven't seen the other version too often, but it's up to you as long as it's the same 😉 👍

RaymondKroon: I only use hasTimeStamp because WSSecurity uses this, I actually prefer hasTimestamp, because de xml field has the same casing.

And the inconsistency I just missed. The typeof field has casing. It got through the test because !!undefined === false. The code works as expected.

It is ok to have different field names for the same functionality in different classes? If yes, I happily change it to hasTimestamp.

herom: Ah, I see. No, just leave it that way (`hasTimeStamp`) but make it consistent in the `options` too, so that it doesn't get too confusing ;)

herom:
thanks a ton @RaymondKroon - we appreciate it a lot 👍

PR #1018, opened 2018-08-16
fix double transformationentries in WSSecCert

RaymondKroon:
Upon multiple client calls, everytime the reference is added. This is fixed by checking the xpath on already added references.

I was tempted to add it in the constructor, because it is not likely that one would change the body namespace on an existing client. It would also add some extra changes though ...


ktj:
@jsdevel Any chance for this to be merged. I've tried this change in my own project and it worked as intended.

PR #1019, opened 2018-08-25
WSSecurity - escaping special characters in usernames and passwords

kulfoniasty:
Current version of node-soap is unable to handle special characters in username and passwords. And as in most cases usernames are simple then it is often the case that passwords do contain & for example. Although PasswordDigest works properly PasswordText causes XML syntax problems. 

Example failing test scenario:
``
  const wsSecurity = new soap.WSSecurity("abc", "Ihaveamper&and");
``


lib/security/WSSecurity.js
```diff
@@ -80,11 +80,17 @@ WSSecurity.prototype.toXML = function() {
...
     "<wsse:UsernameToken xmlns:wsu=\"http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd\" wsu:Id=\"SecurityToken-" + created + "\">" +
-    "<wsse:Username>" + this._username + "</wsse:Username>" +
+    "<wsse:Username>" + wrapInCdata(this._username) + "</wsse:Username>" +
     password +
     (this._hasTokenCreated ? "<wsu:Created>" + created + "</wsu:Created>" : "") +
     "</wsse:UsernameToken>" +
     "</wsse:Security>";
 };
 
+function wrapInCdata(value) {
```
herom: could you please move the declaration of your function onto the top of this file, so that it's declared before we use it? thanks 👍

test/security/WSSecurity.js
```diff
@@ -39,8 +39,8 @@ describe('WSSecurity', function() {
   });
 
   it('should insert a WSSecurity when postProcess is called', function() {
-    var username = 'myUser';
-    var password = 'myPass';
+    var username = 'special&<<>>Characters]]>username</wsse:Username>';
+    var password = 'even&>><<"moreSpecial]]>password';
```
herom: could you please remove the `]]` characters from any of the two variables (either `username` or `password`) so that the test ensures that it'll work without them too? 👍

gjasny:
Hello,

I got hit by the same error. Could you please polish the fix and apply it?

Thanks,
Gregor

jsdevel:
@gjasny we haven't heard back from @kulfoniasty.  If you'd like to pick up where he left off and address the feedback left by @herom feel free.

jsdevel:
thanks @michaelbrawn !! hopefully someone else can pick this up and address @herom 's feedback.  @kulfoniasty: you able to respond?

PR #1020, opened 2018-08-29
If an object doesn't have the hasOwnProperty inheritance, be safe abo…

rschultzgraycor:
If an object doesn't have the hasOwnProperty function inherited, then the original code will fail with obj.hasOwnProperty is not a function. This takes care of that.


lib/wsdl.js
```diff
@@ -1774,7 +1774,7 @@ WSDL.prototype.objectToXML = function(obj, name, nsPrefix, nsURI, isFirst, xmlns
     }
   } else if (typeof obj === 'object') {
     for (name in obj) {
-      if (!obj.hasOwnProperty(name)) continue;
+      if (!({}).hasOwnProperty.call(obj, name)) continue;
```
herom: Please, refer to it as `Object.hasOwnProperty`, so that it's more readable what is going on here, thanks a lot 👍

PR #1023, opened 2018-09-13
WSDL: make merging external schema works correctly

peat-psuwit:
- Remove $targetNamespace check, as the key in definitions.schemas
already guarantee this. Fix importing schema with \<wsdl:import\> where
namespace == schema's targetNamespace.
- Also merge attributes from source schema. Otherwise, attributes such
as elementFormDefault="qualified" specified in included schema will not
work properly.

### DIFF
--- a/lib/wsdl.js
+++ b/lib/wsdl.js
@@ -349,12 +349,19 @@ DocumentationElement.prototype.init = function() {
 
 SchemaElement.prototype.merge = function(source) {
   assert(source instanceof SchemaElement);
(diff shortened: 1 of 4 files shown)


saarivirtajCGI:
Confirmed, tested this on a service that used to produce no input or output from `client.describe()` and it now produces proper description and WS calls work as expected. Would love to see this merged.

jsdevel:
you're kicking it @peat-psuwit !

PR #1024, opened 2018-09-15
Arrays deserve namespace override too

Briggybros:


barboni:
this is huge. @vpulim please please merge this

PR #1025, opened 2018-09-19
emit response events allowing user override on response XML

prasunsultania:
this gives flexibility to override the response XML String if required


prasunsultania:
@herom can this be merged for upcoming release ?

PR #1027, opened 2018-10-05
Added an option field named allDetails.

Kankan023:
The flag allDetails in the option will give all the properties of an element. Currently the output shows only name and type of an element. With allDetails, we can get other properties like minOccurs, nullable, maxOccurs etc. apart from name and type.


Kankan023:
@herom 
Can you check the PR and let me know if I need to do something about it?

PR #1029, opened 2018-10-22
Handle error http status code without response body

aleung:
Prior to v0.24.0 soap client reports error when server returns 404 error without body. This behavior was broken by #986.

And also I suspect that the test case 'Handle non-success http status codes' is invalid and doesn't cover the non-success http status code scenario. I added new test case but didn't remove the old one and let reviewer to check it.


aleung:
Close by #1053

PR #1030, opened 2018-10-24
fix: catch request error event when create client

killagu:


killagu:
@jsdevel Build failures have been resolved. But the `Node.js: 9` build was been terminated.

> No output has been received in the last 10m0s

Can you retry it?

PR #1033, opened 2018-11-22
fix a problem about Multi-Byte character

rhjyy:
If the data contains Multi-Byte character,it may be cut by the tcp packets,then the join function's result will contain some unintelligible text,because it turns the object in the array into string before the concat action.So,we should replace it with the Buffer class's member function concat.


rhjyy:
will create a new request by the current commit to fix the checks

PR #1034, opened 2018-11-22
Update Type Definitions (soap.d.ts)

onury:
- Removed `BluebirdPromise` definition in favour of native `Promise` which will still support Bluebird users.
- Added missing members of `ISecurity` interface
- Added missing members of `IWsdlBaseOptions` interface
- Added missing members of `IOptions` interface
- Added missing members of `IOneWayOptions` interface
- Added missing members of `Client` class
- Made `Client` class extendable with `any` members (Required if you need to extend the class with custom members other than SOAP methods.)
- Added missing members of `Server` class
- Added `INTLMSecurityOptions` interface
- Added `NTLMSecurity` class

### DIFF
--- a/lib/soap.d.ts
+++ b/lib/soap.d.ts
@@ -1,7 +1,6 @@
 /// <reference types="node" />
 
 import { EventEmitter } from 'events';
-import * as BluebirdPromise from 'bluebird';
 
 export interface ISoapMethod {
     (args: any, callback: (err: any, result: any, raw: any, soapHeader: any) => void, options?: any, extraHeaders?: any): void;
@@ -38,6 +37,8 @@ export interface ISoapFault12 {
 
 export interface ISecurity {
     addOptions(options: any): void;
+    addHeaders?(headers: any): void;
+    postProcess?(xml: string, envelopeKey: string): string;
     toXML(): string;
 }
 
@@ -66,10 +67,15 @@ export interface IWsdlBaseOptions {
     ignoredNamespaces?: boolean | string[] | { namespaces?: string[]; override?: boolean; };
     ignoreBaseNameSpaces?: boolean;
     escapeXML?: boolean;
+    preserveWhitespace?: boolean;
     returnFault?: boolean;
     handleNilAsNull?: boolean;
     wsdl_headers?: { [key: string]: any };
     wsdl_options?: { [key: string]: any };
+    useEmptyTag?: boolean;
+    namespaceArrayElements?: boolean;
+    strict?: boolean;
+    customDeserializer?: { [name: string]: (text: string, top: any) => string };
 }
 
 export interface IOptions extends IWsdlBaseOptions {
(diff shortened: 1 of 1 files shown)


Pajn:
Would be great if this was merged and released!

maximeloizeau:
I am trying to define a more strict type for the client custom SOAP methods, something like that:

```
export type CustomSoapMethod = <T>(
  arg1: any,
  ...args: any
) => Promise<[T, string, { Header: { [key: string]: string } }, string]> | any;

export class Client extends EventEmitter {
  constructor(wsdl: WSDL, endpoint?: string, options?: IOptions);
  addBodyAttribute(bodyAttribute: any, name?: string, namespace?: string, xmlns?: string): void;
  addHttpHeader(name: string, value: any): void;
  addSoapHeader(soapHeader: any, name?: string, namespace?: any, xmlns?: string): number;
  changeSoapHeader(index: number, soapHeader: any, name?: string, namespace?: string, xmlns?: string): void;
  clearBodyAttributes(): void;
  clearHttpHeaders(): void;
  clearSoapHeaders(): void;
  describe(): any;
  getBodyAttributes(): any[];
  getHttpHeaders(): { [k: string]: string };
  getSoapHeaders(): string[];
  setEndpoint(endpoint: string): void;
  setSOAPAction(action: string): void;
  setSecurity(security: ISecurity): void;
  wsdl: WSDL;
  [key: string]: CustomSoapMethod | any;
}
```

And then using it like that:
```
interface ApiResponse {
  Field1: {
    Subfield2: string;
    Subfield3: string;
  };
}

export default async function myfunction(data, data2): Promise<Status1> {
  ...
  const client = await soap.createClientAsync(WSDLDescription);
  client.setSecurity(new soap.ClientSSLSecurity(credentials[0], credentials[1]));
  client.addSoapHeader({
    "ns:Header": {
     .....
    },
  });
  if (isCustomSoapMethod(client.ValidateEmpClockStatusAsync)) {
    return client.ValidateEmpClockStatusAsync<ApiResponse>({ ...data3 
}).then(([responseObject]) => ({
      ....
    }));
  }
}

function isCustomSoapMethod(field: any | soap.CustomSoapMethod): field is soap.CustomSoapMethod {
  return typeof field === "function";
}
```
That way, `responseObject` is typed with `ApiResponse`


Can I get your feedback on this implementation? If that's something we could include, or if it's too specific to my use case?
Also I'm not sure if we can include `isCustomSoapMethod` in the soap.d.ts file.

onury:
@jsdevel almost a year. A note on why you didn't merge this would be nice.

jsdevel:
@onury i apologize!  we had way too many open prs from unresponsive developers, and this one was lost in the weeds if you will.  would you be willing to update this pr with the latest from master?  i'll merge it in once that's done.

jsdevel:
`@mention` me once you have it updated with the latest from `master` and i'll re-open and merge.

onury:
Fair enough. thanks. I'll look into it as soon as I find time.

PR #1037, opened 2018-12-17
Stringify wsdl-related error messages to avoid non-descriptive "[object Object]" output.

barboni:
It happens quite often that `body.Fault.detail` is an object. Right now I'm getting validation errors which I can't debug because I can't read into `[object Object]` and parsing `error.root` again only to get the same value of `detail` also doesn't make a lot of sense imo


barboni:
thanks @jsdevel

PR #1040, opened 2019-01-09
Apply custom namespace to the array items

LadnovSasha:
Fixes an issue with custom namespaces with arrays. Also fixed jshint errors and tests
```
client.getSomeMethod({
  'ns:item': [
    foo: 'bar'
  ]
});
```

### DIFF
--- a/test/client-test.js
+++ b/test/client-test.js
@@ -94,16 +94,16 @@ var fs = require('fs'),
       });
     });
 
-    
+
     it('should allow passing in XML strings', function (done) {
       soap.createClient(__dirname + '/wsdl/default_namespace.wsdl', _.assign({ envelopeKey: 'soapenv' }, meta.options), function (err, client) {
         assert.ok(client);
         assert.ifError(err);
-        
+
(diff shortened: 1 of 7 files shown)


test/client-test.js
```diff
@@ -94,16 +94,16 @@ var fs = require('fs'),
...
-    
+
     it('should allow passing in XML strings', function (done) {
       soap.createClient(__dirname + '/wsdl/default_namespace.wsdl', _.assign({ envelopeKey: 'soapenv' }, meta.options), function (err, client) {
         assert.ok(client);
         assert.ifError(err);
-        
+
         var xmlStr = '<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">\n\t<head>\n\t\t<title>404 - Not Found</title>\n\t</head>\n\t<body>\n\t\t<h1>404 - Not Found</h1>\n\t\t<script type="text/javascript" src="http://gp1.wpc.edgecastcdn.net/00222B/beluga/pilot_rtm/beluga_beacon.js"></script>\n\t</body>\n</html>';
-        client.MyOperation({_xml: xmlStr}, function (err, result, raw, soapHeader) {
```
jsdevel: can you add a new test instead of modifying this one?

jsdevel:
i updated the cert/key in master, so you can revert those files too

RishikeshDarandale:
@LadnovSasha , @jsdevel , Is there any possibility to include override the namespace of arrays if I provided as `:array`?  I am facing an issue after specifying `:array`, its taking parent namespce.

I have added a check to exclude adding namespace, if regex does not match(wsdl.js). Is that correct approach?

jsdevel:
@RishikeshDarandale not sure.  feel free to make a proposal.  i'm just here to ensure that certain development practices are followed.

RishikeshDarandale:
> @RishikeshDarandale not sure. feel free to make a proposal. i'm just here to ensure that certain development practices are followed.

Thanks @jsdevel . I will create another pull request.

PR #1041, opened 2019-01-24
processRequestXml only if req.body is not empty object

amity0:
When using `bodyParser.json()` I get empty req.body object which translate to [Object object] and throws exception 
"Non-whitespace before first tag.\nLine: 0\nColumn: 1\nChar: ["

This PR only uses processRequestXml if req.body is not empty

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -203,7 +203,7 @@ Server.prototype._requestListener = function (req, res) {
 
     //request body is already provided by an express middleware
     //in this case unzipping should also be done by the express middleware itself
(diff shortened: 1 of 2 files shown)


amity0:
@jsdevel which file would you say is a good place to add a test? 

Thanks

jsdevel:
please study the current tests and decide the best approach.

amity0:
@jsdevel I added a test. let me know what you think

jsdevel:
thanks!

PR #1042, opened 2019-02-05
fix a problem about Multi-Byte character

rhjyy:
If the data contains Multi-Byte character,it may be cut by the tcp packets,then the join function's result will contain some unintelligible text,because it turns the object in the array into string before the concat action.So,we should replace it with the Buffer class's member function concat.

### DIFF
--- a/test/post-data-concat-test.js
+++ b/test/post-data-concat-test.js
@@ -7,6 +7,16 @@ var http = require('http');
 
 
 describe('post data concat test', function () {
+    var server = http.createServer(function (req, res) { });
+
+    before(function () {
+        server.listen(51515);
+    });
+
+    after(function () {
+        server.close();
+    });
+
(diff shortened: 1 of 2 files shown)


jsdevel:
@rhjyy please add a test

rhjyy:
@jsdevel the test has been added.

test/post-data-concat-test.js
```diff
@@ -0,0 +1,52 @@
...
+                    MyOperation: function (arg) {
+                        console.log(arg);
+                        check(arg, postdata);
+                        return "0";
+                    }
+                }
+            }
+        };
+
+        server.listen(51515);
```
jsdevel: can you move the server declaration/initialization to a `before` block, then in an `after` block tear it down?

rhjyy: done.pls check if it is right.

PR #1043, opened 2019-02-08
server: send response to client with res.end()

peat-psuwit:
Using res.end() to write result instead of res.write() followed by
res.end() will cause node not to use chunked encoding and include
Content-Length header. (See nodejs/node#26005) This small distinction is
important with some client. For example, Windows 10's MDM enrollment
system will not accept chunked response
(https://docs.microsoft.com/en-us/windows/client-management/mdm/on-premise-authentication-device-enrollment).

This might improve compatibility with some other clients, too. However,
there's a small chance that some client may expect the response to be
chunked.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -219,6 +219,7 @@ Server options include the below:
 - `ca`: An array of strings or Buffers of trusted certificates in PEM format. If this is omitted several well known "root" CAs will be used, like VeriSign. These are used to authorize connections.
 - `crl` : Either a string or list of strings of PEM encoded CRLs (Certificate Revocation List)
 - `ciphers`: A string describing the ciphers to use or exclude, separated by  :. The default cipher suite is:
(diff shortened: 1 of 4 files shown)


peat-psuwit:
OK, I'll make this configurable. However, I want your suggestion on option name.

Also, if I add an option, I have to edit the README too, right?

jsdevel:
> Also, if I add an option, I have to edit the README too, right?

yes

> OK, I'll make this configurable. However, I want your suggestion on option name.

maybe `enableChunkedEncoding`?

peat-psuwit:
>maybe `enableChunkedEncoding`?

So, server option `enableChunkedEncoding` with default to `true`?

jsdevel:
> So, server option enableChunkedEncoding with default to true?

That works

jsdevel:
thanks @peat-psuwit !

peat-psuwit:
Thank you for merging my PR, @jsdevel. If you don't mind, could you also take a look at my other PRs, #1022 & #1023, too please?

PR #1044, opened 2019-02-18
converting the project to TypeScript

andygreenegrass:
I have done the work of converting the project to TypeScript.
after cloning simply run `npm i` followed by `npm run build`
you can also run `npm run watch` which will compile the project while you edit.

I wanted to respect the existing code base and made as few changes as possible to the actual implementation. I did my best to fill in types and referenced the existing soap.d.ts for exposed interfaces. I wanted to minimize the use of the any type, but there were places in the code where adding typings would require additional type guards or rework, so I simply typed them as any.

The conversion to classes did require the reorganization of code, but the implementation itself is largely unaffected. I also retained the file structure with the exception of wsdl.js which, in my opinion, was just too large. I created a sub-directory wsdl and broke out the elements objects into a separate file.

I added tslint for linting and disabled many of the recommended rules, again to minimize impact to the existing implementation. These rules can be enabled in the future if desired.

The TypeScript compilation is currently targeting es3. I wanted to ensure that backwards compatibility was not broken. TypeScript will allow the project to utilize some more modern features of JavaScript while still remaining compatible with older versions of node.

I also added typedoc which generates HTML documentation for the project. You can check that out by running `npm run docs`

It is currently passing all tests.


jsdevel:
https://www.youtube.com/watch?v=My2FRPA3Gf8

jsdevel:
@herom thoughts?  i'm reviewing this, and honestly can't find much about it that i don't agree with.  it _will_ invalidate just about _every_ PR out there right now lol.  but perhaps that's a good thing?  (progress anyone?)

overall i'm very impressed @andygreenegrass , and you basically left all the tests untouched!

herom:
This is beyond awesome @andygreenegrass ! 🚀 🥇 I did start a conversion to TypeScript myself but due to new job endeavours I lost focus about it... I think that's really great and much needed in order to get all the things structured and fit for the future! 👍 

@jsdevel I think that this should be merged/done as quick as possible and tagged as ```v1.0.0``` 🐬 as this is clearly marking a big step forward and we can begin to work with `semver` as it was intended 😃 
What do you think about that?

jsdevel:
@herom i could go either way on this.  all the existing tests were untouched, which leads me to believe that it's relatively safe to promote this in the next minor release.  this is mostly refactor/maintenance work.  if we release it as a minor version, we're more likely to find issues with it sooner rather than later, as a lot of users aren't likely to pick up a major bump (unless they have `*` for the version id and aren't using shrinkwrap or package-lock.json).  my preference would be a minor bump, and then we can release fixes that are reported in patch releases.

my main concern is all the existing prs that are open, but most of those have likely been abandoned by now and would probably benefit from being closed (thinking out loud).

RishikeshDarandale:
@jsdevel, if the changes are non breaking and client of this package would not have to  change the code, then it make sense to bump the `minor` version.

If it's `major` release, then probably you need to maintain two versions. 

Rest PRs needs to rebased and do the change according to new one if it's `minor` release. If `major` one, then probably target can be changed for existing PRs.

jsdevel:
either way, going to merge this.  all existing prs will need to be rebased with the latest, which means all prs are likely in conflict.

@andygreenegrass hit me up if you're ever in phoenix and i'll buy you lunch :joy: thanks again!

jsdevel:
i'll need to make some updates to the publish process, but the impact should be minimal

RishikeshDarandale:
Looks like tests are failing, try increasing the `timeout`.

jsdevel:
@RishikeshDarandale it's actually an issue with coveralls.  they're looking into it

andygreenegrass:
@jsdevel thanks for the merge! I really appreciate the positive responses. I'm a big fan of what you guys have built and I'm looking forward to helping make it even better.

PR #1046, opened 2019-02-19
npm upgrade;  removing ejs and external template files

andygreenegrass:
I ran an `npm upgrade` to reduce vulnerabilities in the audit.
The only outstanding vulnerabilities are from `nyc` and it requires a major version update. I didn't want to break any code coverage stuff, so I left it alone.

I also removed the dependency on external files in `WSSecurityCert.ts`, which also removed the need for `ejs`.


jsdevel:
i'm in awe @andygreenegrass .  thank you!

PR #1047, opened 2019-02-21
adding source-map-support for ts mapping in stack trace; enabling tsl…

andygreenegrass:
…int rules; added linting to travis

I added the source-map-support dev dependency which will resolve ts file and line numbers in stack traces during the tests.
I enabled some of the previously disabled tslint rules and made the updates to fix those.
I also added linting to travis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,17 +26,21 @@
   "main": "./index.js",
   "types": "./lib/soap.d.ts",
   "directories": {
-    "lib": "./lib"
+    "lib": "./lib",
+    "test": "./test",
+    "example": "./src/examples"
   },
   "scripts": {
+    "prepare": "npm run build",
     "build": "tsc -p .",
     "clean": "rm -rf lib",
     "watch": "tsc -w -p .",
+    "lint": "tslint -p tsconfig.json",
     "toc": "./node_modules/.bin/doctoc Readme.md --github --maxlevel 3",
-    "cover": "nyc --reporter=lcov --reporter=html --reporter=text mocha --exit test/*-test.js test/security/*.js",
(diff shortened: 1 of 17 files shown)


RishikeshDarandale:
How about using [husky](https://www.npmjs.com/package/husky) for linting and formatting at the time of commit?

andygreenegrass:
I haven't used husky. I'll check it out. It _is_ completely unnecessary that linting runs for every node version in travis.

Also, can I rebase an active pull request? I usually rebase my commits before I push to github.

andygreenegrass:
I added husky and lint-staged to do a partial lint of staged files on commit. If there are linting error it will attempt to fix what it can or fail to commit if the errors cannot be automatically fixed.

jsdevel:
Absolutely no husky sorry.  Please remove that.

package.json
```diff
@@ -60,13 +75,16 @@
     "express": "^4.16.4",
     "finalhandler": "^1.1.1",
     "glob": "^7.1.3",
+    "husky": "^1.3.1",
```
jsdevel: Please remove husky

RishikeshDarandale: @jsdevel , any particular reason to remove  `husky`?

andygreenegrass:
husky is gone.

jsdevel:
Thanks.  In a publicly shared repo, I'd like to limit what runs on my machine when making contributions.

andygreenegrass:
I just went through and removed all unnecessary `self` variables. I wanted to get that in there along with the move to arrow functions. There will be no more changes for this pull request.

andygreenegrass:
well, that's not good. I'm a little confused by the coverage results.

andygreenegrass:
I think it's just because of the removed `const self = this;` lines.

jsdevel:
i can accept the coverage decrease.  thanks @andygreenegrass !

PR #1049, opened 2019-02-22
npmignore cleanup; adding some types to Client

andygreenegrass:
- Added examples - I personally find example code to be helpful in terms of getting up and running with a new library. I also think it gives the library developer an opportunity to define best practices since it will be copy/pasted by everyone.

- Added `/docs` to `.npmignore` - I think ultimately the project should use an npm white list.

- It looks like the `last*` properties of `Client` should be public so I changed that.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,3 +6,4 @@ History.md
 PUBLISHING.md
 .*
 /src
+/docs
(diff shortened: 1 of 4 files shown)


andygreenegrass:
That's fair. examples removed.

jsdevel:
thanks!

PR #1050, opened 2019-02-23
adding promise support for server method handlers.

andygreenegrass:
Allows service method handlers to return a promise.

- Added a test for promise return value.
- moved `--require source-map-support/register` to `mocha.opts` so that it's used all the time.
so running something like `npx mocha test/server-test.js` will also have ts mapping in stack traces.
- updated `Readme.md` with promise example

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -30,6 +30,10 @@ function isExpress(server): server is IExpressApp {
   return (typeof server.route === 'function' && typeof server.use === 'function');
 }
 
+function isPromiseLike<T>(obj): obj is PromiseLike<T> {
+  return (!!obj && typeof obj.then === 'function');
+}
+
 function getDateString(d) {
   function pad(n) {
     return n < 10 ? '0' + n : n;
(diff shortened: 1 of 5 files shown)


andygreenegrass:
There is currently an issue with async errors (for callbacks) which also affects this implementation. An error returned via the callback actually follows a different path than an error thrown synchronously.

_Any_ error thrown synchronously ends up being converted to a full blown soap error on line 395 of server.ts. However due to the ambiguity of error and result in handleResult starting at line 457, if the error is not itself a soapFault (line 463), it will be handled as if it were the result (467).

I can fix this for the promise implementation, but it will still be an issue for callbacks (again due to the ambiguity presented by the backward compatibility).

src/server.ts
```diff
@@ -30,6 +30,10 @@ function isExpress(server): server is IExpressApp {
   return (typeof server.route === 'function' && typeof server.use === 'function');
 }
 
+function isPromiseLike<T>(object): object is PromiseLike<T> {
+  return (typeof object.then === 'function');
```
jsdevel:
```
return !!object && ...
```

andygreenegrass: will do

andygreenegrass:
- made the requested change.
- added the fix for async errors for Promises.
- added a new test for promise based async error.

PR #1052, opened 2019-03-24
add rsa-sha256 support for WSSecurityCert

nvbach91:
Looks like the only supported encryption currently supported is rsa-sha1 (default in xml-crypto)
http://www.w3.org/2000/09/xmldsig#rsa-sha1

Added option to specify rsa-sha256
http://www.w3.org/2001/04/xmldsig-more#rsa-sha256

### DIFF
--- a/src/security/WSSecurityCert.ts
+++ b/src/security/WSSecurityCert.ts
@@ -54,14 +54,14 @@ export class WSSecurityCert implements ISecurity {
       .replace(/(\r\n|\n|\r)/gm, '');
 
     this.signer = new SignedXml();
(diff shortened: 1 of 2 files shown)


jsdevel:
I'll merge once the build is fixed

nvbach91:
Hi @jsdevel, I tried to fix the build but I got lost in the code. It shows 
```Unhandled rejection AssertionError: undefined deepEqual { return: '-1', bstrError: '' }``` but I have no idea why it happened. Could you please have a quick look and point out where exactly is the error? I can then fix it myself. Thanks a lot

nvbach91:
@jsdevel nevermind, it's fixed.

PR #1053, opened 2019-03-26
Handle response with error status code and without response body

aleung:
Prior to v0.24.0 soap client reports error when server returns 404 error without body. This behavior was broken by #986.

And also I suspect that the test case 'Handle non-success http status codes' is invalid and doesn't cover the non-success http status code scenario. I added new test case to non-success status code but didn't remove the old one and let reviewer to check it.

This PR supersedes #1029 since code had been changed to TypeScript.


jsdevel:
thanks @aleung !

PR #1055, opened 2019-03-28
Added callback for soap.listen

rymdo:
Added ability to specify callback to soap.listen, which is used when wsdl.onReady is called.


jsdevel:
i'll merge once the build passes

PR #1057, opened 2019-04-11
Fixed some issues with xsd elements

zanemcca:
- Namespace was not fetched properly when `definitions.xmlns[type.prefix]` is `undefined`. There appears to be a small typo in the logic that was fixed. Please comment if this logic was intentional and we can discuss it.
- Some types are still strings when the `defaultsDeep` is called. Explicitly checking for strings before calling `defaultsDeep` resolves the issue


jsdevel:
please fix the build

zanemcca:
I updated the first patch to be more specific to the particular use case. I am a `node-soap` newb so any input on the fix would be appreciated.

jsdevel:
can you add a test?

zanemcca:
Test added :)

jsdevel:
thanks!

zanemcca:
Thank you @jsdevel! When should I expect this to be published?

jsdevel:
we're due for a publish soon.  i may have time in the next few days

PR #1058, opened 2019-04-15
dedup xmlns links

13W:


jsdevel:
@13W please fix the build and provide a test

13W:
I hurried, the problem is somewhere deeper, but this fix works for me

I have a wsdl request that creates namespaces with similar schemes (more than 100 schemes, after deduplication - 7), and the service is not responding

PR #1061, opened 2019-04-24
Adds outputName type to soap response for server. 

chriswiggins:
This fixes issues when using this with some SOAP clients, namely ONVIF Device Manager


chriswiggins:
This actually causes more problems currently. Will re-open the PR when I work out what else needs solving

PR #1062, opened 2019-04-25
client.addSoapHeader() dynamic SOAP header

kqkq:
I need to add a "message ID" (a random UUID) on each SOAP request. But I found `client.addSoapHeader()` won't accept a function as a parameter. The dynamic SOAP header feature only works in `server.addSoapHeader()`. So I made some change to enable it for the `client`.

I'm relatively new here. I wonder if you are interested in this feature or I just misunderstand the documentation and the dynamic header should be added in another way.

If I am right and you also believe the feature should be added; I will create test cases for it and get the documentation updated.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -58,7 +58,7 @@ export class Client extends EventEmitter {
 
   private wsdl: WSDL;
   private httpClient: HttpClient;
-  private soapHeaders: string[];
+  private soapHeaders: any[];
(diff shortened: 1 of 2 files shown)


jsdevel:
please add a test

kqkq:
Tests updated :grinning:

jsdevel:
thanks!

PR #1063, opened 2019-04-29
types: move forceSoap12Headers to IWsdlBaseOptions

peat-psuwit:
"forceSoap12Headers" works on server since bcc41e6. So, this option
should be available to server as well.


jsdevel:
thanks!

PR #1064, opened 2019-05-03
Fixes for ONVIF Server methods

chriswiggins:
I was trying to use this library to create an ONVIF server (similar to the `breeze/rpos` project) and was having similar issues to them as to what this library was doing to some of the XML elements.

These changes pass the tests but I don't know whether they'd affect anyone else's applications. They certainly fixed mine! :-)


jsdevel:
@chriswiggins can you add tests?  we're not concerned about breaking things that don't have tests.  we _do_ care about breaking existing tests.

chriswiggins:
Sure thing! I'll try and sort this today 👍

PR #1065, opened 2019-05-11
Added support for parsing of doubles and floats.

alexb-uk:
Also package-lock updated due to out-of-date version currently checked into `master`.

Minor tweak to existing tests to better cover the parsing of double and float.

**Background:**

After reviewing the existing very old PR that fixes doubles and floats thought best to create a new one with a slightly different approach

See: https://github.com/vpulim/node-soap/pull/985

I've gone for `Number(text)` instead of `parseFloat(text)` as I think it's safer to throw an error on unexpected characters rather than allowing an assumption.

E.g. "1a" !== 1

There is a good explanation of the pro's con's of `Number()` vs `parseFloat()` here: https://stackoverflow.com/questions/12227594/which-is-better-numberx-or-parsefloatx

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.27.1",
+  "version": "0.26.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jsdevel:
please revert package-lock.json.  we'll update it.

alexb-uk:
> please revert package-lock.json. we'll update it.

Done, thanks a lot.

alexb-uk:
I'm a little confused as to why Travis failed for Node v9 and v7, is that a known issue?

alexb-uk:
If it helps the build passes on my branch so guess it was an intermittent test timeout?

See: https://travis-ci.org/alexb-uk/node-soap/builds/531663881

jsdevel:
thanks @alexb-uk !

PR #1066, opened 2019-05-15
Added Options object for signer.computeSignature

LorneCurrie:
Locked the XML-Crypto to 1.2.0
updated the readme with details about options for WSSecurityCert function
Added the signer Options to the WSSecurityCerts OPtions to be used in xml-crypto.  Helps with a breaking change in 1.4.0. Detail about change below.

XML-crypto 1.4.0, would add the `xmlns` `xmlns:wsse=""` to the `<wsse:SecurityTokenReference xmlns:wsse="">` when the `signer.computeSignature` function was called.  
This was different behaviour to the version (1.2.0) locked in the project.  
When Node-Soap was imported into my project, the version of XML-crypto that was set using the hat `^` would pick the latest minor version release, which had different behaviour.

Other option for the fix would be to lock the version fo xml-crypto to 1.2.0 minor range

Pull request: 1067 is another option that would fix this option and is less evasive.

XML before the change:

```
    <KeyInfo>
        <wsse:SecurityTokenReference xmlns:wsse="">
          <wsse:Reference URI="#x509-35f93d1596fa461cb092cc0014197dcd" ValueType="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3"/>
        </wsse:SecurityTokenReference>
      </KeyInfo>
```
After the change
```
      <KeyInfo>
        <wsse:SecurityTokenReference xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd">
          <wsse:Reference URI="#x509-35f93d1596fa461cb092cc0014197dcd" ValueType="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3"/>
        </wsse:SecurityTokenReference>
      </KeyInfo>
```


jsdevel:
@LorneCurrie please update your git config so we can properly track you in git history later on.

also, please add a test

holmmikael:
I'm using version 0.28.0 from node-soap and I'm facing this error. When the fix will be released?

jsdevel:
@holmmikael when tests are added.  if you'd like to copy this effort and add tests in your own pr feel free, otherwise you'll need to reach out to @LorneCurrie .

LorneCurrie:
Sorry, I have been caught up on some things.  I'll update the tests this weekend

LorneCurrie:
Tests added

jsdevel:
Thank you!

PR #1067, opened 2019-05-15
Lock xml-crypto

LorneCurrie:
XML crypto  version 1.4.0 has a breaking change
Locked version to 1.2.0 as that still works with the current version of node soap

This is a less invasive option to Pull request #1066 I submitted

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.26.0",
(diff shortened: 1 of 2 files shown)


jsdevel:
can you revert package-lock.json?  we'll update it.

jsdevel:
can you add this to #1066 ?

LorneCurrie:
will do

PR #1071, opened 2019-05-31
Expose the WSDL class

ties:
While it is not a documented use case, the WSDL class can be used
directly to (de-)serialize XML. This enables users to use an SOAP
endpoint that does not expose bindings in the WSDL.

In my specific use case, the SOAP endpoint is an interface to Microsoft Dynamics with URL's documented in documentation. The service can alledgedly be used directly by `svcutil.exe`, but is less idiomatic to consume from python or nodejs.

I hope you will consider making the WSDL class part of the public API.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -670,6 +670,87 @@ Example :
   }, {exchangeId: myExchangeId})
 ```
 
+## WSDL
+
+A WSDL instance can also be instantiated directly when you want to (un)marshal
+messages without doing SOAP calls. This can be used when a WSDL does not contain
+bindings for services (e.g. some Windows Communication Foundation SOAP web
+services).
+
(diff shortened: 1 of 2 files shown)


jsdevel:
@ties this looks good.  can you add documentation in the README?

ties:
I added documentation in the readme. I do not know the project very well. I'm confident the example code works (it's modeled after my real usage), however it is probably useful if you double check that my interpretation of the functionality is correct.

The usage of `objectToXML` over `objectToRpcXML` or `objectToDocumentXML` was the first one I started using after switching from Jsonix after that was not compatible with some types in my WSDL...

jsdevel:
thanks!

PR #1073, opened 2019-06-05
added possibility to add action to content-type header

bonanzakrak:
I have added support to add soapAction to Content-Type header.

For example:
https://docs.intersystems.com/latest/csp/docbook/DocBook.UI.Page.cls?KEY=ROBJ_method_soapaction

It allows to set Content-Type to `'application/soap+xml; charset=utf-8'; action=Action`

Added option param: `addHeadersAction` that allows to turn this on/off.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -194,6 +194,7 @@ export class Client extends EventEmitter {
       this.wsdl.options.overrideRootElement = options.overrideRootElement;
     }
     this.wsdl.options.forceSoap12Headers = !!options.forceSoap12Headers;
+    this.wsdl.options.addHeadersAction = !!options.addHeadersAction;
   }
 
   private _defineService(service: ServiceElement, endpoint?: string) {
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -364,7 +360,13 @@ export class Client extends EventEmitter {
       soapAction = ((ns.lastIndexOf('/') !== ns.length - 1) ? ns + '/' : ns) + name;
     }
 
-    if (!this.wsdl.options.forceSoap12Headers) {
+    if (this.wsdl.options.forceSoap12Headers) {
+      headers['Content-Type'] = 'application/soap+xml; charset=utf-8';
+      if (this.wsdl.options.addHeadersAction) {
```
gbar: According to specification from https://docs.intersystems.com/latest/csp/docbook/DocBook.UI.Page.cls?KEY=ROBJ_method_soapaction action param should be added within forceSoap12Headers flag, not some additional one

jsdevel:
thanks!

src/types.ts
```diff
@@ -94,6 +94,8 @@ export interface IWsdlBaseOptions {
   wsdl_options?: { [key: string]: any };
   /** set proper headers for SOAP v1.2. */
   forceSoap12Headers?: boolean;
+  /** set content type header action for SOAP v1.2 */
+  addHeadersAction?: boolean;
```
jsdevel: @gbar can you follow up with a PR to add this to the README please?

bonanzakrak: Will add that when I get home. Will create another PR with that

PR #1076, opened 2019-06-06
Fixes issue with unknown ReadableStream type

pkuczynski:
When using the latest `soap` version 0.27.1, I am getting the following error:

```
node_modules/soap/lib/http.d.ts:10:11 - error TS2304: Cannot find name 'ReadableStream'.

10     body: ReadableStream;
```

The proposed change is fixing it. Typescript version 3.5.1

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -22,7 +22,7 @@ export interface IAttachment {
   name: string;
   contentId: string;
   mimetype: string;
-  body: ReadableStream;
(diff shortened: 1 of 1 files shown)


shaungreen:
Any thoughts when this fix will be published to npm?

jsdevel:
soon.  we give master some time to simmer.  probably in the next few days.

PR #1077, opened 2019-06-07
Bad description for MessageElement when reference a type

lafeuil:
When generating the description of MessageElement, the description of each parts are not call and return a return the object not the description. This issue happens when the part references a type.

With the wsdl-to-ts project, this generates bad interface : https://github.com/TimLuq/wsdl-to-ts/issues/7


jsdevel:
please add a test

PR #1078, opened 2019-06-18
Fix description for recursive wsdl with extended element

mastodon0:
I found a bug when using a WSDL that includes recursive types combined with extended elements.
The description of recursive objects would be empty because `_.defaultsDeep` creates copies of child objects. Using `_.defaults` only creates the necessary object references and solves the issue.

The testcase shows the problem: We have the types `Person` -> `Department` -> `Person (HeadOfDepartment)`
`personDescription.Department.HeadOfDepartment` would be an empty object, but should be the same object as `personDescription`


jsdevel:
thanks @mastodon0 !

PR #1080, opened 2019-06-21
Fix server type error

david-templeton:


jsdevel:
@david-templeton can you remove the submodule?

PR #1081, opened 2019-06-21
Update server.ts

david-templeton:
Fixed issue causing error message, "TypeError: Cannot read property &apos;output&apos; of undefined"


jsdevel:
@david-templeton can you add a test?

PR #1082, opened 2019-07-11
Adjusts WSDL URL detection to be case insensitive

cmptrgeekken:
Per [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.1), the URL schema (e.g., http:, https:) can be case insensitive. This PR adjusts the URL detection regex to ensure that capitalized schemas (e.g., HTTP, HTTPS) are supported.

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -1169,7 +1169,7 @@ export class WSDL {
     }
 
     let includePath: string;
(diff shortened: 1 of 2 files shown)


jsdevel:
please add tests

cmptrgeekken:
Unit test has been added.

PR #1086, opened 2019-08-08
added forceMTOM option and updated the Readme

g-m-a:
* cleaned some of the previous code (lame console.log leftover)
* added forceMTOM in the request options so that you can send MTOM request even without having binary attachments to the request. Magically enough some services always use MTOM.
* updated the Readme with some info on the `attachments` and `forceMTOM` options


jsdevel:
thanks!

PR #1089, opened 2019-08-27
Allow a fixed file path for local includes

mritzco:
NetSuite wsdl uses relative paths but when downloaded NS flattens all files in a single directory and fails to load.
This option loads all the files from the directory as provided by NetSuite zip file.

Using with node-suitetalk
https://github.com/felipechang/node-suitetalk


mritzco:
Modified readme.md 
I tried to ammend the previous commit but now the story looks very dirty. Let me know if I should create a new pull request with a single commit.

PR #1092, opened 2019-10-31
Add WSDL xmlns header attribute

mateioprea:
I've encounter a case where the SOAP Header value needs to have an xmlns:value attribute. 
In Bing Ads SOAP requests (maybe for every Microsoft Service SOAP api) the xmlns attribute in the header element is required. 

Example request here: https://docs.microsoft.com/en-us/advertising/campaign-management-service/getaccountmigrationstatuses?view=bingads-13#request-header 

This provides a way of adding it through ```client.wsdl.xmlnsInHeader```. 

For above example, ```client.wsdl.xmlnsInHeader = 'xmlns="https://bingads.microsoft.com/CampaignManagement/v13"'```; should work just fine.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -409,7 +409,7 @@ export class Client extends EventEmitter {
       this.wsdl.xmlnsInEnvelope + '>' +
       ((decodedHeaders || this.security) ?
         (
-          '<' + envelopeKey + ':Header>' +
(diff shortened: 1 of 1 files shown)


mateioprea:
Sorry, i've rushed a bit. Let me redo the PR.

PR #1094, opened 2019-11-05
Allows SchemaElement instance to use import namespace as targetNamesp…

ndeniche:
### Allows SchemaElement instance to use import namespace as targetNamespace when the attribute is not set.

Prevents the following error message when schema elements do not include targetNamespace attribute:

> Target-Namespace "undefined" already in use by another Schema!

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -802,8 +802,10 @@ export class TypesElement extends Element {
   // fix#325
   public addChild(child) {
     assert(child instanceof SchemaElement);
-
-    const targetNamespace = child.$targetNamespace;
+    const targetNamespace = child.$targetNamespace ||
+      child.includes.find(e => {
+        return e.hasOwnProperty("namespace");
+      }).namespace;
 
     if (!this.schemas.hasOwnProperty(targetNamespace)) {
       this.schemas[targetNamespace] = child;
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -802,10 +802,11 @@ export class TypesElement extends Element {
   // fix#325
   public addChild(child) {
     assert(child instanceof SchemaElement);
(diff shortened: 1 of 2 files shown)


jsdevel:
@ndeniche can you explain why coverage dropped?

ndeniche:
@jsdevel

Actually, I cannot. I'm having issues with the tests setup, since its failing on httpClient in client-customHttp-test.js.

Is there a previous step before `run ci` or `run cover` that I should be executing to setup the test environment?

src/wsdl/elements.ts
```diff
@@ -802,8 +802,11 @@ export class TypesElement extends Element {
   // fix#325
   public addChild(child) {
     assert(child instanceof SchemaElement);
-
-    const targetNamespace = child.$targetNamespace;
+    const childInclude = child.includes.find((e) => {
+      return e.hasOwnProperty('namespace');
+    });
+    const targetNamespace = child.$targetNamespace ||
+      typeof childInclude !== 'undefined' ? childInclude.namespace : childInclude;
```
jsdevel: can you add parens around the ternary?  i think you mean this:

```
    const targetNamespace = child.$targetNamespace ||
      (typeof childInclude !== 'undefined' ? childInclude.namespace : childInclude);
```

also, the tests are failing because of a TypeError when trying to resolve `.namespace`.  please see the travis output in the builds.

ndeniche: I was so invested on fixing the first test, I didn't stop to see my attempt at fixing the lint errors broke the logic.

ndeniche:
Closing pull request to follow correct procedure: fix errors and add tests and documentation.

jsdevel:
@ndeniche please don't close prs and open new ones.  you can update your current PR.  if you're not sure how to do that, please take the time to learn how.  the git community will appreciate it.

PR #1097, opened 2019-11-13
 SOAP envelope 1.2 schema valid Fault response 

sigitas-stumbras:
Current SOAP 1.2 Fault response does not comply with schema: http://www.w3.org/2003/05/soap-envelope


jsdevel:
@sigitas-stumbras please fix the build and add tests

jsdevel:
please don't open and close PRs

PR #1098, opened 2019-11-15
SOAP envelope 1.2 schema valid Fault response

sigitas-stumbras:
SOAP envelope 1.2 schema valid Fault response (http://www.w3.org/2003/05/soap-envelope)


jsdevel:
please don't open and close a pr

PR #1100, opened 2019-12-09
Fix min/maxOccurs parsing and handling

kouak:
Consider this XSD :

```xml
                <xs:element name="DummyList" minOccurs="1" maxOccurs="10">
                    <xs:complexType>
                        <xs:sequence>
                            <xs:element name="DummyFirst" type="xs:string" />    
                            <xs:element name="DummySecond" type="xs:string" />
                        </xs:sequence>
                    </xs:complexType>
                </xs:element>
```

We should expect node-soap to deserialize DummyList as an array regardless of the number of <DummyList> elements in the document.

Consider this response :
```xml
<Parent>
    <DummyList>
        <DummyFirst>foo</DummyFirst>
        <DummySecond>foo</DummySecond>
    </DummyList>
</Parent>
```

Before this PR, this would be transformed to :
```json
{
  "DummyList": {
    "DummyFirst": "foo",
    "DummySecond": "foo"
  }
}
```

After this PR :
```json
{
  "DummyList": [
    {
      "DummyFirst": "foo",
      "DummySecond": "foo"
    }
  ],
}
```

We noticed this regression upgrading our app from node-soap 0.26.0 to 0.30.0. We pinpointed its introduction between 0.26 and 0.27, possibly during the typescript rewrite.

Also [the former typings](https://github.com/vpulim/node-soap/blob/af93a6f45720431371ab3f23866d3ad3f2efae1b/src/wsdl/elements.ts#L188-L189) were wrong: AFAIK, sax parses minOccurs and maxOccurs attributes to strings making [this code](https://github.com/vpulim/node-soap/blob/af93a6f45720431371ab3f23866d3ad3f2efae1b/src/wsdl/elements.ts#L200-L203) not perform correctly unless `maxOccurs` was set to "unbounded".


jsdevel:
please fix the build

kouak:
Coveralls seems to be the issue for the failed builds : https://travis-ci.org/vpulim/node-soap/jobs/622742785#L1622-L1625

Not sure how I can fix that though.

kouak:
I ran the pipeline again by force pushing to the source branch.

Travis CI is all good now. The previous error was probably just a fluke.

jsdevel:
please revert package-lock.json.  also, i can't tell what was changed in the src file because the entire file was formatted.  please revert space changes.

kouak:
My bad, the PR should be reviewable now.

kouak:
Yeah I had a look into that but I'm at lost here ...

If I read the following image correctly, coveralls reports a decreased coverage in `wsdl/` but an increase in both `wsdl/index.ts` and `wsdl/element.ts` ?

Capture d’écran de 2020-01-10 21-45-43

Also, it looks like all the added lines are covered if I read this report correctly.

Capture d’écran de 2020-01-10 21-46-10

jsdevel:
thanks @kouak !

jsdevel:
@kouak could it be that we now have dead code elsewhere?

kouak:
@jsdevel: As far as I'm aware I'd say no.

Could it be possible to prepare a new release including this fix ? That would greatly help our project. Thanks !

jsdevel:
soon @kouak

PR #1101, opened 2020-01-10
Fix for circular descriptions

mastodon0:
There seems to be an edge case when using refs in wsdls like this:
```xml
      <xsd:element name="ExampleContent" type="v1:ExampleContent"/>
      <xsd:element name="ExampleRequest" type="v1:ExampleRequestType"/>

      <xsd:complexType name="ExampleContent">
        <xsd:sequence>
          <xsd:element name="MyID" type="xsd:string" />
        </xsd:sequence>
      </xsd:complexType>

      <xsd:complexType name="ExampleRequestType">
        <xsd:sequence>
          <xsd:element ref="v1:ExampleContent"/>
        </xsd:sequence>
      </xsd:complexType>
```

The reference to `ExampleContent` would cause a circular description because there is an element and type of the same name.

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -224,7 +224,10 @@ export class ElementElement extends Element {
         if (!(typeName in definitions.descriptions.types)) {
 
           let elem: any = {};
-          definitions.descriptions.types[typeName] = elem;
+          if (!this.$ref) {
+            definitions.descriptions.types[typeName] = elem;
+          }
+
           const description = typeElement.description(definitions, xmlns);
           if (typeof description === 'string') {
             elem = description;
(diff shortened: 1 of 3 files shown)


jsdevel:
thanks @mastodon0 .  can you add more test cases to keep coverage up?

mastodon0:
I could add an additional test but it would be unrelated to the fixed bug.
This looks like a problem with coveralls since it doesn't list any file that got decreased coverage.

jsdevel:
yea, it does seem like something is up.  thanks!

jsdevel:
actually, @mastodon0 could it be with this fix that we now have dead code elsewhere?

mastodon0:
I looked into that and doubt it. I know it sounds like the classic "it works on my machine" but running npm run cover has higher coverage after the commit than before.

jsdevel:
ok.  thanks for checking @mastodon0

PR #1103, opened 2020-01-29
Clear http client header

KurtzL:
The `public clearHttpHeaders (): void` method in the` client.ts` file does not really remove the Http headers. 
In the `_invoke` method, another variable was created to manage the headers. This variable doesn't even care if the variable `this.httpHeaders` was cleaned by user


jsdevel:
@KurtzL please squash to 1 commit.  i don't want the package-lock.json change in git history

KurtzL:
sorry, i hope i did good now

PR #1107, opened 2020-02-20
Escape username and password in wss

holm:
We found out that the username and password in WSS does not do any escaping. With a password with an ampersand, the entity will not be properly encoded and the XML payload will be invalid.


danrpts:
will there be a version bump for this soon?

jsdevel:
@danrpts the next publish will be a month from the last one.

PR #1109, opened 2020-04-14
Prefix was not correct when using arrays

efreibe:
If using array as parameters then you cannot have a way to prefix the method with namespace using overrideRootElement

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -767,7 +767,7 @@ export class WSDL {
         }
 
         if (isFirst) {
-          value = this.objectToXML(child, name, nsPrefix, nsURI, false, null, schemaObject, nsContext);
+          value = this.objectToXML(child, name, nsPrefix, nsURI, Array.isArray(child), null, schemaObject, nsContext);
         } else {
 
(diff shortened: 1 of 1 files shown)


jsdevel:
please add a test

grandpaSam:
Is there a temporary work around for this? I'm having this issue at the moment

vtrphan:
since 0.37.0 i'm having problem with prefixes too. For example i have an attribute Roles which is an array, like `Roles : [{ RoleName: "someName", RoleDesc: "a Description"}]`, then i prefix it manually so that it becomes like: `"ns1:Roles": [{ "ns1:RoleName": "someName", "ns1:RoleDesc": "a Description"}]`, then i pass this array into node-soap, the xml request it produces will be like `<Roles><ns1:RoleName>someName</ns1:RoleName><ns1:RoleDesc>a Description</ns1:RoleDesc></Roles>`. Notice the namespace "ns1" has been stripped off from "Roles". This caused my request to fail. Any idea or fix?
This behavior is since v0.37.0. Before that, all worked well.

PR #1110, opened 2020-05-04
Bug where no prototypical inheritence chain exists

NoMan2000:
- Some transpilers will create objects using `Object.create(null)`.  This creates an object with no prototypical inheritance, so there is no `hasOwnProperty` method on it.

The fix is if that if such an object exists, use `Object.assign` to copy the prototype chain from an empty object into the null created obj.

Otherwise, you just get a TypeError that `obj.hasOwnProperty is not a function`.

Signed-off-by: NoMan2000 <webart.video@gmail.com>

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -730,6 +730,10 @@ export class WSDL {
       }
     } else if (typeof obj === 'object') {
       for (name in obj) {
+        // Happens when Object.create(null) is used, it will not inherit the Object prototype
+        if (!obj.hasOwnProperty) {
+          obj = Object.assign({}, obj);
+        }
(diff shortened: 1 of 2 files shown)


ZacharyKearns:
Is anyone reviewing this? After updating to v0.31.0 this issue broke my app I had to roll back to a previous version.

NoMan2000:
The fix is dead simple, so I'm guessing that they haven't seen it?

ZacharyKearns:
This didn't fix my issue unfortunately. Any null values in the args still causes the request to fail.

NoMan2000:
> This didn't fix my issue unfortunately. Any null values in the args still causes the request to fail.

Can you submit some sample code and tag me on it?

PR #1112, opened 2020-05-18
Fixed preserveWhitespace not taken in account if text contains only spaces

DeadEye0112:
The preserveWhitespace option is not taken in account if the text only contains spaces. This leads to cases where the returned text should be ' ' but leads to ''.

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -418,7 +418,7 @@ export class WSDL {
     p.ontext = (text) => {
       const originalText = text;
       text = trim(text);
-      if (!text.length) {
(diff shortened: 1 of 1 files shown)


jsdevel:
@DeadEye0112 i apologize for the late response.  it appears that this change is breaking some of the tests.  is it possible to invert the flag to preserve the old default behavior?  i.e. instead of `preserveWhitespace`, maybe `discardWhitespace`?

w666:
Hi @DeadEye0112,

Seems like this feature was already added in https://github.com/vpulim/node-soap/pull/1211, could you please check it and close PR if it is no longer needed?

PR #1114, opened 2020-05-27
Add overrideImportLocation option

silverwind:
To mock a remote server for tests, I need a way to override both the endpoint used for the SOAP services as well as the URI used in `import` statements inside the WSDL.

The existing `Client.setEndpoint` works for the first case and this adds a new `overrideImportLocation` wsdl option to override the import statements.

I think this option is generally useful, for example if a remotely obtained WSDL imports another unreachable WSDL, it allows users to supply the imported WSDL locally.


wesley102:
Ok

PR #1115, opened 2020-06-21
feat(Server): method called with apply

bruno-brant:
When using node-soap to create a soap server, one could prefer to declare and instantiate a class and pass that as the service object (_although the current documentation does instruct developers to use anonymous objects_). This would allow us to create more complex service implementations by using methods, properties, etc.

However, right now, doing so results in little gain since methods are being called directly, without correctly setting the "this" variable.

The correction is very simple, though: the owning object for `method` is `port`; it's just a question of instead of invoking it directly, just call `apply` and pass the `port` as it's `this` variable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.29.0",
+  "version": "0.31.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/server.ts
+++ b/src/server.ts
@@ -6,7 +6,7 @@
 import { EventEmitter } from 'events';
 import * as http from 'http';
 import * as url from 'url';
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.29.0",
+  "version": "0.31.0",
```
jsdevel, line 3: please revert this.  maintainers will bump

smokhov: @w666 -- it appears that you guys already have the tests covering this due to the nature of the change? The only thing holding this up was that premature version bump; once removed, this should OK, no?

w666: Added to my backlog. Will look a bit later this or next week.

w666: erm, can't easily check these changes, @bruno-brant has deleted forked repo. Will return to this later when have time.

PR #1116, opened 2020-06-29
fix: serverless-http compatibility

wik89:
if you use serverless-http to serve soap server the url lib is compiled http://example.com/?wsdl to ?wsdl=.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -239,7 +239,8 @@ export class Server extends EventEmitter {
     }
 
     if (req.method === 'GET') {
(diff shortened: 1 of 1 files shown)


paulish:
This breaks handling of wsdl request where some letters are not in lower case. E.g.:
?WSDL
?Wsdl

If you are going to compare using startingWith the please replace `reqQuery.startsWith('?wsdl')` by `reqQuery.toLowerCase().startsWith('?wsdl')`

jsdevel:
@paulish can you open a PR to fix this?

PR #1118, opened 2020-07-30
Convert decimals

simhnna:


simhnna:
@jsdevel could we get a release with the fix soonish?

jsdevel:
done

PR #1119, opened 2020-07-30
Added MTOM support for multipart/related responses with attachments

chrisgh1:
Pretty new to soap and typescript so please check over the code.  

I've got this working in my environment but have not tested outside of that.  

Basically what is going on is I have added the option to have the request module return binary and not utf8.  I then check the headers and see if their is a boundary in the header that would indicate a multipart response.  if no boundary I use toString(utf8) to convert back to utf8 so node-soap runs as it used to.  

If we have a boundary I get the boundary id with some regex and pass the res.body and the boundary to the mime-reader.js (feel free to convert this to type-script I do not know how to do that)

In mime-reader.js we create a buffer to hold the first part of the multipart response which is the soap xml response.  then I use a temp file to write the binary data for each of the parts to a file.  I had this using a buffer but on large files I would get slow performance and sometimes out of memory errors.  I package all the headers and the buffer in a parts array and return it.  

then I use the first element in the parts array and convert the buffer to a utf8 string and assign it to the body variable

the other parts get assigned to res.attachments

client.ts then looks to see if response.attachments exists and if it does we assign it to obj.Body.getContentResponse.return.attachments

I use the multipart parser from formidable v1.2.2 and I code written by:

//Original code from Yaron Naveh, 
//https://github.com/yaronn/ws.js/

for the mime-reader.  

Again this has only been tested in one environment and I'm new to soap and typescript so please check it over for issues.  It's currently working in my environment and fulfilling my needs

Also all my edits were done in the javascript files and I tried to convert them to the type-script but may have errors.  I've only been running my javascript files not the type-script ones.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.29.0",
+  "version": "0.32.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


jsdevel:
please add tests and avoid using `var`.

chrisgh1:
Can someone tell me how to add the mime-reader.js file so that Travis CI build can find it?  not sure how to do this in type-script.  In javascript I had var mimeReader = require('./mime-reader.js');  or if I use import I thought I have it alphabetized but apparently not?  Import sources within a group must be alphabetized.

davefej:
Issue can be closed because It is a duplication of https://github.com/vpulim/node-soap/pull/1148 which is already merged

PR #1122, opened 2020-09-26
Avoid lodash where possible

TrySound:
Lodash is quote big project and it's required less with a time in modern
environments. In this diff I bumped node support to 8+ and used native
methods where possible instead of lodash. Later we can go further and
completely remove it from the project.

This can make the project much smaller.
https://packagephobia.com/result?p=soap

### DIFF
--- a/soap-stub.js
+++ b/soap-stub.js
@@ -1,5 +1,3 @@
-var _ = require('lodash');
-
 var aliasedClientStubs = {};
 var clientStubs = {};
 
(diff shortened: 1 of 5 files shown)


jsdevel:
you're killing it @TrySound !  can you rebase onto latest master?

soap-stub.js
```diff
@@ -1,5 +1,3 @@
-var _ = require('lodash');
```
jsdevel, line 1: can we remove lodash altogether?

TrySound: It's just one module. I will try to cover other cases in upcoming PRs.

TrySound:
@jsdevel Done

PR #1123, opened 2020-09-28
Replace concat-stream with get-stream

TrySound:
Ref https://packagephobia.com/result?p=concat-stream https://packagephobia.com/result?p=get-stream

Concat-stream is a big package with not relevant in node 10 dependencies.
In this diff I replaced it with dependency-free get-stream with promise
based api.


jsdevel:
thank you!

PR #1125, opened 2020-09-29
Drop bluebird

TrySound:
Using bluebird for promisifying is a bit overhead. Especially since the
package is quite massive. https://packagephobia.com/result?p=bluebird

In this diff I replaced Bluebird.promisifyAll with custom promisifying
specifically for services methods instead of all class methods.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -151,12 +151,6 @@
         }
       }
     },
-    "@types/bluebird": {
-      "version": "3.5.26",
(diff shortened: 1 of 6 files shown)


Biktop:
This pull request broke async methods without parameters. Now I have to pass empty object as workaround.

jsdevel:
@Biktop can you submit a pr?

Biktop:
Sorry but I don't think that I'm able to do it. I've just discussed this workaround on gitter and tested it.

PR #1126, opened 2020-09-30
Setup prettier

TrySound:
In many places code has inconsistent formatting which makes developing
quite hard. In this diff I setup prettier as linter. Users should format
code by their editors. Or with format script.

Tell me if you want to change some rules.
https://prettier.io/docs/en/options.html

You can review with hidden whitespaces.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,3 @@
-"use strict";
+'use strict';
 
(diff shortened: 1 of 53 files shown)


TrySound:
No idea why coveralls fails.

PR #1131, opened 2020-10-28
upgraded xml-crypto to latest version

Kifah:
Upgrading `xml-crypto` extension, in order to avoid a severe vulnerability reported by GitHub.
Bildschirmfoto 2020-10-28 um 16 51 48

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,124 +5,120 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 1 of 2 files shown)


juuxstar:
👍

manishgadhock-monotype:
Awaiting for this change. +1

atishayjain:
+1

gaurkeshav:
+1

jsdevel:
thank you!

mllanes:
It seems this change didn't make it to v 0.34.0?

danleegion:
Package version should be updated with 0.34.1 or 0.35.0 as NPM was not updated due to version

jansenlau:
Any idea on when a new package version update that includes this change will be available?

jsdevel:
0.35.0 has it

marjandjokic:
Hi, I have problem when switching to version 0.35.5 of node-soap as now request are sent without body.const client = await createClientAsync(serviceConfig.wsdl, soap_client_options);
const result = await client.healthCheckAsync();
This works well with 0.32.0 but not with 0.35.0 I see my request has empty bodies. Any idea what needs to be done, maybe some additional options?

Thanks in advane

marjandjokic:
> Hi, I have problem when switching to version 0.35.0 of node-soap as now request are sent without body.const client = await createClientAsync(serviceConfig.wsdl, soap_client_options);
> const result = await client.healthCheckAsync();
> This works well with 0.32.0 but not with 0.35.0 I see my request has empty bodies. Any idea what needs to be done, maybe some additional options?
> 
> Thanks in advance, Marjan

I noticed that when I don't send any parameters in method like I do above: const result = await client.healthCheckAsync();
Then request looks like: (Request method is ommited)
<?xml version=\"1.0\" encoding=\"utf-8\"?>
<soap:Envelope...>
	<soap:Body></soap:Body>
</soap:Envelope>
But If I add some dummy parameters then all good, I fixed it like that, but I just wonder if that can be done in some better way.

PR #1135, opened 2021-01-21
add handling for extension base without children

KnutEdelbert:
When parsing the wsdl, the trace don't recognize the reference in the base attribute of an ExtensionElement without child nodes, when it comes to findChildSchemaObject. It should be handled like the type attribute on ElementElements


KnutEdelbert:
ok, finally, it looks good.
@jsdevel: is there anything else i have to do now?

jsdevel:
@KnutEdelbert please remove the change to package-lock.json

KnutEdelbert:
@jsdevel  as you wish - it's deleted

KnutEdelbert:
ok, i think it was better to revert the file instead of deleting it....
@jsdevel Everything ok now?

PR #1138, opened 2021-03-15
update xml-crypto version

germankrause:
I have the `node-soap` in my dependency tree and I've got a security alert from `dependabot`, it says: update `xmldom` to `0.5.0`. `xmldom` is a dependency of `xml-crypto`, which is a dependency of the `node-soap`, and they have already updated the `xmldom` to 0.5.0, so it's your turn :)
[The alert](https://github.com/advisories/GHSA-h6q6-9hqw-rwfv)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "sax": ">=0.6",
     "strip-bom": "^3.0.0",
     "uuid": "^8.3.0",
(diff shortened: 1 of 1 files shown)


Loksly:
Would you please consider publishing a new version of `soap` package on _npmjs_?

PR #1142, opened 2021-05-03
Circular element references inside wsdl - assistance needed

doubleaxe:
Hello.

This fix https://github.com/vpulim/node-soap/pull/1101 fixed one bug and introduced another. Now parsing wsdl with recursive ref elements will cause stack overflow. I created 2 test cases which demonstrate this bug, but I need advice of someone who knows this project better than me.

I can quick fix stack overflow with one kludge - just like https://github.com/vpulim/node-soap/pull/1101, something like this, or maybe a little more complex - because second test will fail with this:
```
          if (!this.$ref) {
            definitions.descriptions.types[typeName] = elem;
          } else {
            const refType = splitQName((typeElement as ElementElement).$type);
            if(refType.name != typeName)
              definitions.descriptions.types[typeName] = elem;
          }
```

As pedantic person, I think kludge is not the way to do. In my opinion main problem is more deep in internals: both element names (with references) and type names go to the same object `definitions.descriptions.types`, they are mixed, and this introduces all these bugs with elements and types, but I'm not sure in this, because I'm new to node-soap internals, I just use it as a regular user.

So I need advice from someone, I can try to fix this by splitting `definitions.descriptions.types` for elements and for types, but it seems this is core internal feature, I'm afraid to break something and waste my time making PR which will not be accepted.

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -231,6 +231,7 @@ export class ElementElement extends Element {
           definitions.xmlns[type.prefix];
       const schema = definitions.schemas[ns];
       const typeElement = schema && ( this.$type ? schema.complexTypes[typeName] || schema.types[typeName] : schema.elements[typeName] );
+      const typeStorage = this.$type ? definitions.descriptions.types : definitions.descriptions.elements;
 
       if (ns && definitions.schemas[ns]) {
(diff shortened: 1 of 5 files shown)


doubleaxe:
I updated PR, this is my vision on how this bug should be fixed. All tests seems passing, I only doubt in backward compatibility.

PR #1148, opened 2021-06-02
Support for soap attachments in response.

davefej:
### Summary
With this modification the lib can parse the multipart/related http response, and get the (binary) attachments from it.

### Configuration
The funcionality is behind a boolean option called **parseReponseAttachments** (default: false, not to affect previous functionality)
If mtomResponse is set to true we expect a multipart response however it can handle normal soap response (without attachment) as well.

### Modification list

- encoding is set to null before sending http request (it cause response to be Buffer instead of string)
- checking response header for multipart content-type and get boundary
- parse multipart content with formidable MultipartParser (first part is the soap response xml)
- attachments are reachable in soapClient.lastResponseAttachments
- request-response-samlples test is extended with attachment.js and responseHttpHeaders.json

### Dependencies
- https://www.npmjs.com/package/formidable
- https://www.npmjs.com/package/content-type-parser

### Example usage
`const client = soap.createClientAsync(wsdl, { parseReponseAttachments: true} );  `
`await client.methodAsync(json)`
`const attachments = client.lastResponseAttachments`

### Docs
W3C not about soap attachments: https://www.w3.org/TR/SOAP-attachments/

### Other
The modification is similar to https://github.com/vpulim/node-soap/pull/1119 but configurable and working.


DPM1:
@davefej great work, thank you! I just tried it and it worked out of the box.
I think it would be a good idea to also include mtomAttachements in the results instead of having to use lastResponseAttachments. What do you think?
MR draft here:
[https://github.com/davefej/node-soap/pull/1](https://github.com/davefej/node-soap/pull/1)

davefej:
@DPM1 Thanks, Absolutely agreed on include attachments into the result woult be great and the preferred solution.
However that would be a major modification.
Maybe I could implement that in the future in another pull request what do you think of this solution?

davefej:
Hi There!
Travis ci is not started automatically for https://github.com/vpulim/node-soap/pull/1148/commits/49e2e63c0dfd275f6c8eb7413284fcd65d6527e6

Is it becaues of travis-ci.org is deprecated since June 15?

DPM1:
Would be great to get this into the next release. If anyone would be nice enough to do a review, I would be happy to then do any work required to get this released.
For a start it probably needs at least 1 unit test for the mtomAttachments return, I did not add one in my draft MR. I will try to add one this week. 
Additional review comments are always welcome. 
Thanks

jsdevel:
please resolve the conflicts

davefej:
Hi There!
@DPM1 Tests are already done for MTOM. Those are in the request-response-samlpes.
(should_handle_MTOM_response and should_handle_not_MTOM_response)

@jsdevel Merge done with new axios mods on master

DPM1:
@davefej great! I just quickly tested the rebase on module I have running with MTOM attachments and it worked perfectly. 
I noticed the "request" lib was re-added on rebase so created [MR ](https://github.com/davefej/node-soap/pull/2/) to remove it again.
Thanks!

package.json
```diff
@@ -1,79 +1,82 @@
...
-    "semver": "^5.6.0",
-    "serve-static": "^1.14.1",
-    "should": "^13.2.3",
-    "sinon": "^1.17.7",
-    "source-map-support": "^0.5.10",
-    "timekeeper": "^2.1.2",
-    "tslint": "^5.18.0",
-    "typedoc": "^0.20.36",
-    "typescript": "^3.3.3333"
+    "name": "soap",
```
jsdevel: please revert the white space changes in this file.

jsdevel (changes requested):

test/request-response-samples-test.js
```diff
@@ -135,8 +151,15 @@ function generateTest(name, methodName, wsdlPath, headerJSON, securityJSON, requ
 
   suite[name] = function(done){
     if(requestXML) requestContext.expectedRequest = requestXML;
-    if(responseXML) requestContext.responseToSend = responseXML;
+    if (responseXML) {
+      if (wsdlOptions.parseReponseAttachments) {//all LF to CRLF
+          responseXML = responseXML.replace(/\r\n/g, "\n");
```
jsdevel, line 156: why is this needed?

jsdevel: you're converting rn, to n, then you're converting n back to rn.  please revert this.

jsdevel (changes requested):

test/request-response-samples-test.js
davefej: Hi!

It is needed, might be implemented in different way, but needed.

The reason is only CRLF allowed in the multipart/related response. And in linux systems (for example travis-ci) git repo pulled with LF only. However if response xml has LF line endings the standart http multipart response parser cannot parse it so tests fails.

That two lines of code does is to replace all "\n" to "\r\n" .
If line 156 would have been left out `responseXML = responseXML.replace(/\r\n/g, "\n");` and only line 157 would be executed than it might occur that some line endings will have "\r\r\n" and that is not correct of course.

So someone with better regexp knowledge might add a solution to replace all LF to CRLF without replacing the LF parts of CRLF's but it is needed.

package.json
davefej: Hi!
You compared with a previous commit, it is already fixed in the master.
https://github.com/vpulim/node-soap/pull/1148/files

test/request-response-samples-test.js
jsdevel: so you're saying there may be a mix of rn initially, so your strategy is to convert those to n then convert all n to rn?

davefej: Yes.
These lines are because of I noticed that tests fails on linux system, and the reason was that there were \n -s in the file.
So I replaced all to \n to \r\n in a dirty way I mean taking care of not to replace \r\n to \r\r\n...

dursk:
@jsdevel Is there a target to get this into a release? I desperately need this!

PR #1146, opened 2021-05-19
replace request & httpntlm with axios & axios-ntlm

vision10:
This is my attempt to replace request and httpntlm packages with axios and axios-ntlm

Its not perfect yet and there are still some issues that need to be addressed:
- axios does not handle multipart options. 
  - axios recommends to use `form-data` package for attachments, but I don't know if we should include another package or let the user handle this
- `requestStream`  method does not return a request stream but a data response as a stream (because axios does not return the request stream, if only has the option to return the response as a stream with options.responseType='stream')
- a few tests need some rewriting (I have adjusted a few to work with the new packages as best as I knew/could)

I have updated a few packages (those with minor changes)
Some files have a few more changes because of vs code text autoformat (mainly whitespaces)
On a personal note, I don't think Promises mix well together with callbacks, so maybe callbacks should be dropped in a future version if axios is adopted ?

PR #1154, opened 2021-08-02
Correctly add the https agent to axios

scagood:
According to the [axios Request Config](https://github.com/axios/axios#request-config) the agent needs to be passed using `httpsAgent`.

This is a follow up to https://github.com/vpulim/node-soap/pull/1146#issuecomment-889908849


jsdevel:
@scagood please update with the latest from master.  i had to migrate to github actions.

sm2017:
There is a bug in version Version 0.41.0, I am using Version 0.39.0 and it works well today I upgraded to latest version and found that there is a bug in my app, after inspecting I figured out that vpulim/node-soap#1154 is a bad PR, I am using https-proxy-agent package to send requests via a http proxy, the code works with proxy in version 0.39.0 but not latest (I am not testing 0.41.0 but I think the issue is here)

sm2017:
@jsdevel @scagood please fix proxy bug

scagood:
@sm2017 This is more likely another issue using axios, so you may want to start there (https://github.com/axios/axios/issues?q=is%3Aissue+is%3Aopen+https-proxy-agent) before making a PR to fix this.

sm2017:
@scagood It was working before

PR #1161, opened 2021-09-01
fix undefined targetNamespace

asxelot:


jsdevel:
@asxelot can you add a test?

asxelot:
@jsdevel done

PR #1163, opened 2021-09-06
minor fix for ntlm request

vision10:
1 minor change in code for ntlm requests,
bumped axios-ntlm package to latest version to fix an error, 
I also removed httpntlm because its not used anymore
also shouldn't the `if parseReponseAttachments { ... }`  be in the `handleResponse` function?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "soap",
-  "version": "0.43.0",
+  "version": "0.42.0",
   "description": "A minimal node SOAP client",
   "engines": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,19 +1,18 @@
 {
   "name": "soap",
-  "version": "0.42.0",
```
jsdevel, line 3: please revert this.  maintainers bump the version.

jsdevel (changes requested):

PR #1167, opened 2021-09-27
Publicly expose wsdl in Client

imgx64:
The `wsdl` property in `Client` is currently private. And while it's possible to manually construct it using `new WSDL(...)`, the actual implementation as done in `createClient(...)` is more complicated and does caching/fetching from file or url/etc.


imgx64:
I need to manually call wsdl.objectToXML to convert the request to XML without actually sending it.

onury:
We also have cases where we extend classes and need access to `wsdl`.

PR #1170, opened 2021-10-12
soap-header-fix

amalashenok:
There is an issue with appearing of empty soap:Header tag, although decodedHeaders are set to null it's being formed, because of this.security is defined, but this.security.toXML() returns empty string.


jsdevel:
@amalashenok please provide a test

PR #1171, opened 2021-10-21
fix error when soapaction header is not set

paulish:
Fixes error mentioned here: https://github.com/vpulim/node-soap/pull/1152#issuecomment-947758759


jsdevel:
@paulish thanks for submitting this!  is there a way you can test this?

MrOldProger:
For testing you can use SoapUI tool with disabled 'soapaction' property in request's property sheet.

paulish:
@jsdevel added the test

PR #1172, opened 2021-11-05
Fix missing parameter in example

dj95:
The example in the README does not include a required parameter. If this parameter is omitted, the client will always be `undefined`.


dj95:
Since you do not seem to care about pull requests I'll close this one one...

jsdevel:
@dj95 so harsh!  please re-open.  sorry, i usually go through the open PRs every month or so.

dj95:
Since there was progress on other issues I thought that you might not care. Sorry for my impatience!

dj95:
Thank you for reopening and merging the PR :)

PR #1173, opened 2021-11-12
correctly serialize MTOM into axios data and add Content-Encoding: gzip support in soap request

ischisan:


ischisan:
content was empty when using MTOM and Content-Lenfth was 0, so i added the logic for serializing the multiparts into the body.

jsdevel:
please undo all the style changes

ischisan:
> please undo all the style changes. i can't even see what you changed with all the whitespace changes...

sorry, prettier was activated ;-)

i removed the style changes.

BR

ischisan:
I added support to enforce gzip compression in soap request

Readme.md
```diff
@@ -522,6 +522,7 @@ Interesting properties might be:
   ]
   ```
 * `forceMTOM`: set to True if you want to send the request as MTOM even if you don't have attachments
+* `forceGzip`: set to True if you ant to force transfer-encoding in gzip (default: ```false```)
```
jsdevel, line 525: ant -> want

Readme.md
```diff
@@ -554,6 +555,7 @@ Interesting properties might be:
...
```
jsdevel, line 558: ant -> want

jsdevel (changes requested):

jsdevel:
please ensure the build passes

jsdevel (approved):

ischisan:
> @ischisan do you have a way of adding a test for this?

this is hardly possible when the soap server does not support MTOM. Maybe I can add that server functionality later, but not before next month. I would provide a test for server and client after that.

jsdevel:
that would be great if you could follow up when you can @ischisan

PR #1174, opened 2021-11-23
enabling async methods for postProcess

karlop:
Enabling resolving promises on postProcess

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -293,7 +293,7 @@ export class Client extends EventEmitter {
     }
   }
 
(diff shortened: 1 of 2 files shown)


jsdevel:
please ensure the build passes.

.github/workflows/pr.yml
```diff
@@ -1,5 +1,5 @@
 name: PR Build
-on: pull_request
```
jsdevel, line 2: please revert this.  i don't think it is building on your PR now.

jsdevel (changes requested):

PR #1176, opened 2021-12-23
add optional targetNamespace to output message name

Benomi:
By default, if the response builder is not RPC stylized. It will remove `targetNamespace` from the message `output` tag and the message parts. With this simple fix, it will add `targetNamespace` to the response name. It will help fix issues like [this](https://stackoverflow.com/questions/70453937/how-to-edit-the-response-fields-in-node-soap).


jsdevel:
@Benomi can you add a test please?

Benomi:
@jsdevel check now.

Benomi:
@jsdevel I've rebased to the latest master and fixed errors as well.

jsdevel:
thank you!

PR #1179, opened 2022-01-12
Change axios to be a peer dependency

ssiltanen:
The usage of axios in this repository is very basic, but its versioning
is defined very tightly. This causes problems, because the interface
IOptions uses AxiosInstance as its property, preventing the clients of
this library from updating their axios version beyond the limits set in
this library. Typescript compiler gives type error when giving higher
version AxiosInstance for IOptions.

By setting axios as a peer dependency, we would allow the clients of
this library to install other versions of axios if they so choose.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.43.0",
       "license": "MIT",
       "dependencies": {
-        "axios": "^0.21.1",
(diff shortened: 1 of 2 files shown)


ktj:
axios-ntlm (dependency of this project) also has "axios": "^0.21.3"
similar PR could be made there

PR #1182, opened 2022-02-03
adds support for fastify

flovilmart:
Hi! Great library! We're using it with fastify in a project and found it would be useful to have a 'native' implementation. 

What do you think?


jsdevel:
i'd rather see you make the library more pluggable if it's not already.  building in hooks for every framework out there isn't feasible.

flovilmart:
cc @rv1

PR #1180, opened 2022-01-19
Change axios back as dependency.

RopoMen:
Hi, 

The last PR (https://github.com/vpulim/node-soap/pull/1179) where axios moved as peer-dependency was a way to fix type issues with axios instance if project depending from node-soap would also use axios as their dependency and axios versions would be different.

The idea is correct, but I'm currently not sure what would be the correct fix. Correct fix is not to move axios as peer-dependency, because broke node-soap usage with npm versions lower than 7, because npm versions 3-6 will not install peer-dependencies by default. PR checks run by node-soap CI pipeline https://github.com/vpulim/node-soap/pull/1179/checks were green, because build pipeline is using node 16.13.2 which has npm 8.x and that will install peer-dependencies automatically.

---

Not PR related.
`GitHub issues have been disabled to focus on pull requests.`, I wish you would enable issues again because that would allow normal way of communication. If you do not like "please help me" issues, then close those automatically or guide users to use gitter for that purpose.

Axios released new 0.25.0 version that fixes security issue in 'follow-redirects' package, I did not update axios in this PR, because I'm not sure how `axios-ntlm` works with different axios versions.

From some PR comments my co-worker picked up phrase something like "Because node-soap is under version 1, we can make breaking changes.". This is at-least 7 year old project and it has no good alternatives so this package IS in production use in many places and I suggest that working habits in this project get more mature. Change from `request` to `axios` caused breaking change although it was not mentioned as such. My customers project issue was caused proxy handling differencies between request and axios and fix caused over 10 000 eur cost (consultant fixing code) after this node-soap patch version upgrade.

PR #1183, opened 2022-03-15
Upgrade formidable to v2

orgads:
v1 is deprecated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "axios-ntlm": "^1.2.0",
         "debug": "^4.3.2",
-        "formidable": "^1.2.2",
+        "formidable": "^2.0.1",
         "get-stream": "^6.0.1",
         "lodash": "^4.17.21",
         "sax": ">=0.6",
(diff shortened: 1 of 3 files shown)


orgads:
@vpulim ping

orgads:
The failure doesn't look related to my change. It looks the same as in #1180.

jsdevel:
@orgads i know you didn't cause it.  i'm mainly a PR reviewer and publisher.  if you have time to fix it on master i'd greatly appreciate it

alfaproject:
The update to this version now breaks our webpack build and accordingly to formidable, updating to version 3 fixes the issue. Is there any reason why you didn't choose to update to version 3 instead?

orgads:
I used `latest`. Wasn't aware of v3 (which is still not `latest` at this point).

alfaproject:
Fair enough, I've followed up with their team to see what's the status. Thanks!

kvetoslavnovak:
After upgrading to 0.45.0 there seems to be breaking change for me. 
node-soap v 0.45.0 update broke soap.createClientAsync()

I am not able to track were the problem might be. , see the description of the problem here in Stackoverflow:
[node-soap v 0.45.0 update broke soap.createClientAsync()](https://stackoverflow.com/questions/74339110/node-soap-v-0-45-0-update-broke-soap-createclientasync)

PR #1186, opened 2022-05-16
promisifyMethod return object instead of array to allow for destructuring

anasik:
Currently, if users do this:
> const result =  await client.MyFunction(args) 
The result will be equal to an array containing result, rawResponse, soapHeader, rawRequest, mtomAttachments. And if they want just the parsed JSON response they have to do result[0].

This is fine. But if we make promisify return an object instead of an array, they could do this:
> const { result } =  await client.MyFunction(args)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -236,7 +236,7 @@ export class Client extends EventEmitter {
           if (err) {
             reject(err);
           } else {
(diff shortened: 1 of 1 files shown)


smokhov:
@w666 -- could you please rebase this to run the checks?

w666:
I reviewed this again, while I think this may provide some benefits, it will break it for many users of this library, unfortunately I can't do this.

PR #1187, opened 2022-05-20
WSSecurity Protocol with user/pass token element.

OlafurTorfi:
Hi

I pulled the package, and the tests were failing out of the box. So I just made some minor hacks to get the tests running.

Then I updated all but two of the packages (the major versions of formidable and strip-bom required more changes than fixing the import and adding some types. So I left those.

Then I added a new security protocol, WSSecurityCertWithToken. Is basically the same as WSSecurityCert, except that it takes in a username and password as well. Which we need here in iceland.

I have used this WSSecurityCertWithToken protocol implementation in production, and can confirm that it works.


package.json
```diff
@@ -1,82 +1,84 @@
...
-    "serve-static": "^1.14.1",
-    "should": "^13.2.3",
-    "sinon": "^1.17.7",
-    "source-map-support": "^0.5.10",
-    "timekeeper": "^2.1.2",
-    "tslint": "^5.18.0",
-    "typedoc": "^0.20.37",
-    "typescript": "^3.9.10"
-  }
+    "name": "soap",
```
jsdevel, line 2: can you revert these changes or keep the changeset small?

jsdevel (changes requested):

package.json
OlafurTorfi: done :)

jsdevel:
please fix the build

package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "soap",
-  "version": "0.43.0",
```
jsdevel, line 3: please revert the version field.

jsdevel (changes requested):

package.json
OlafurTorfi: Ok.
Although I believe it is correct according to semver specification [https://semver.org/#spec-item-7](https://semver.org/#spec-item-7)
```Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API```

jsdevel:
@OlafurTorfi please fix the build

OlafurTorfi:
The merge conflict has been resolved.

jsdevel:
@OlafurTorfi please fix the build

w666 (approved):

w666:
I resolved conflict and seemed like it works. Wondering if @jsdevel wants to review it again? 
I personally don't see any problem merging it.

w666:
No objections on this PR, tests passed, merging.

PR #1188, opened 2022-06-09
Fix axios peer dependency version

mohd-akram:
This fixes the issue mentioned in #1180. It should remain a peer dependency since its types are exposed by this library, but the version restriction should be loosened. Hopefully this can be released soon as the current version of the library cannot be installed if a user has a direct dependency on a newer version of axios.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,7 @@
         "node": ">=12.0.0"
       },
       "peerDependencies": {
-        "axios": "^0.21.1"
+        "axios": ">=0.21.1"
       }
     },
     "node_modules/@babel/code-frame": {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "xml-crypto": "^2.1.3"
   },
   "peerDependencies": {
-    "axios": "^0.21.1"
+    "axios": ">=0.21.1"
   },
   "repository": {
     "type": "git",


jsdevel:
closes #1188

alfaproject:
I don't know exactly what you guys did, but the latest version of `soap` doesn't work for us anymore. I tried adding `axios` as explicit dependency but it doesn't seem to help either. Our repository has a few different versions of `axios` added by other dependencies and probably the wrong one is being picked up by webpack... Can you be explicit about exactly what your node module requires? Later versions of `axios` don't seem to be compatible either because they don't follow semver...

jsdevel:
alright, i think we need to add axios as an explicit dependency (i.e. merge #1180).  @alfaproject please monitor #1180 .

mohd-akram:
This doesn't seem related to axios, it seems related to formidable and webpack.

@alfaproject see node-formidable/formidable#337.

alfaproject:
Ok my bad, but either way I won’t be able to upgrade node-soap, meh ):

MegaSpaceHamlet:
```bash
node_modules/nestjs-soap/node_modules/axios
  soap  0.40.0 - 0.44.0 || >=1.0.0
  Depends on vulnerable versions of axios
  node_modules/nestjs-soap/node_modules/soap
    nestjs-soap  2.0.0 - 2.2.1 || >=3.0.1
    Depends on vulnerable versions of soap
    node_modules/nestjs-soap
```

`soap 0.40.0 - 0.44.0 || >=1.0.0` depends on axios 0.27.2 as a peer dependency, which is apparently vulnerable, and definitely outdated. Is it possible to update the peer dependency? Or at least make it `>=0.27.2`

PR #1189, opened 2022-06-22
feature to have the option to change the key of the wsdl 

Lucas-Mondini:
I was having a problem with a company, that the wsdl was returning the wrong data to the soapclient; but despite i knew the expected data, I can't change by the code.
So I made a small change that allow me to change the XML key before send to the server
I just have to pass an object with the key value that I want to change, and the value that I want that this key have
```
        const options = { 
            overrideElementKey: {
                oldKeyOnTheXML: "NewKeyOnTheXML"
            }
        }
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -188,6 +188,10 @@ export class Client extends EventEmitter {
         }
       }
     }
+
+    if(options.overrideElementKey !== undefined) {
+      this.wsdl.options.overrideElementKey = options.overrideElementKey;
+    }
     if (options.overrideRootElement !== undefined) {
       this.wsdl.options.overrideRootElement = options.overrideRootElement;
     }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -188,6 +188,10 @@ export class Client extends EventEmitter {
         }
       }
     }
+
+    if(options.overrideElementKey !== undefined) {
```
jsdevel, line 192: please document this in the README

jsdevel (changes requested):

src/client.ts
smokhov: @w666 -- is it only the README change that is required that's holding up this?

w666: Also needs a test. I added this one to my todo list.

PR #1190, opened 2022-07-18
Bumped Axios version to ">=0.25.0"

aniltallam:
If the response size for any SOAP API is greater than 0x1fffffe8 characters then getting the following uncaught error and node process is crashing.
```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at Buffer.utf8Slice (<anonymous>)
    at Buffer.toString (buffer.js:799:17)
    at IncomingMessage.handleStreamEnd (/ws/node_modules/soap/node_modules/axios/lib/adapters/http.js:262:41)
    at IncomingMessage.emit (events.js:327:22)
    at IncomingMessage.EventEmitter.emit (domain.js:467:12)
    at endReadableNT (internal/streams/readable.js:1327:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  code: 'ERR_STRING_TOO_LONG'
}
```

The issue is due to older version of Axios (https://github.com/axios/axios/blob/v0.21.1/lib/adapters/http.js#L253) not properly handling this error.

But this is fixed in v0.25.0 of Axios (https://github.com/axios/axios/blob/v0.25.0/lib/adapters/http.js#L302).  So bumped the axios version to >=0.25.0 in the package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "xml-crypto": "^2.1.3"
   },
   "peerDependencies": {
-    "axios": ">=0.21.1"
+    "axios": ">=0.25.0"
   },
   "repository": {
     "type": "git",


aniltallam:
@vpulim can you look at this PR .
btw some test cases are failing. not sure if we can ignore and comment them.

jsdevel:
@aniltallam please update with the latest from master

benasher44:
I think this is covered by #1192

becerriljc:
> Hope the axios in `node-soap` could updated!

I Hope too, the axios updated!

lehh:
> I have conflicts when using this package with the latest `axios@1.1.3` installed in my project.
> 
> Hope the axios in `node-soap` could updated!

I believe it's an axios problem. They fixed it in the [axios](https://github.com/axios/axios/pull/5225/commits) package itself, and it is supposed to be released on [v1.2.0](https://github.com/axios/axios/blob/v1.x/CHANGELOG.md).

jsdevel:
plz resolve the conflicts

andrew-dias:
> I think this is covered by #1192

Was it necessary to change the `axios` version resolution in #1192 from `>=` to `^`? This causes a regression where axios versions >=1.0 are no longer compatible (at least in npvm v7+ I think).

benasher44:
No it was a mistake. Didn't realize axios was 1.0. Should probably just upgrade this package?

andrew-dias:
I wouldn't upgrade the peer dependency to `1.0.0` because that would just break it for anyone still on `0.x`, and we want the peer dependency to reflect the lowest compatible version. We could even drop it back to `>=0.25.0` as this PR proposed, if that is the minimum compatible version (I'm not sure if you upgraded to `0.27.2` just because it was the latest available version, or if it was actually required for your PR).

benasher44:
It was just the latest available version at the time

andrew-dias:
I am able to work with axios v1.x by using the `overrides` field in `package.json`:

```
  "overrides": {
    "soap": {
      "axios": ">=0.25.0"
    }
  },
```

rmkanda:
Can we mark the peer dependency to support old and new versions? 
```
"peerDependencies": {
  "axios": "^0.27.2 || ^1.2.2"
}
```

andrew-dias:
> Can we mark the peer dependency to support old and new versions?
> 
> ```
> "peerDependencies": {
>   "axios": "^0.27.2 || ^1.2.2"
> }
> ```

No need to be any more restrictive than what this PR already proposes:

```
  "peerDependencies": {
    "axios": ">=0.25.0"
  },
```

w666:
Should not be problem anymore, new version released with updated axios.

PR #1193, opened 2022-08-09
Check if Schema has been defined

willmustafa:
**The Problem:**
Some Magento XML gets an error in the line ``this.element = schema.elements[nsName.name];`` because schema is undefined.

**The Solution:**
Simple check if schema has been defined ``if(schema) this.element = schema.elements[nsName.name];``

**Tests:**
All existing tests passed. I didn't create a new test because all worked well and new tests would be the same.


jsdevel:
@willmustafa it would be nice if you could cover the else condition.  i'm assuming existing tests didn't cover that

PR #1196, opened 2022-08-18
Improvement when there is no explicit Body in the request

aurelmegn:
I have improved the input request parsing when there is no explicit body in the initial request Xml.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -369,7 +369,7 @@ export class Server extends EventEmitter {
             try {
                 const soapAction = this._getSoapAction(req);
                 const messageElemName = (Object.keys(body)[0] === 'attributes' ? Object.keys(body)[1] : Object.keys(body)[0]);
-                const pair = binding.topElements[messageElemName];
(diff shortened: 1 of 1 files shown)


aurelmegn:
Please, which tool do you use to format the code. My IDE change much more than needed.

src/server.ts
```diff
@@ -27,38 +33,60 @@ type Request = http.IncomingMessage & { body?: any };
...
 
 function isPromiseLike<T>(obj): obj is PromiseLike<T> {
-  return (!!obj && typeof obj.then === 'function');
+  return !!obj && typeof obj.then === "function";
 }
 
 function getDateString(d) {
   function pad(n) {
-    return n < 10 ? '0' + n : n;
+    return n < 10 ? "0" + n : n;
```
w666, line 45: Code style should not be changed, this makes it hard to review and does not provide any benefits. Please, revert styling changes and leave only actual changes.

w666 (changes requested):

PR #1199, opened 2022-10-14
Upgrade XML Crypto

AdrianHL:
This fixes the following issue in a nested dependency https://github.com/advisories/GHSA-9pgh-qqpf-7wqj
The fix in xml/crypto has only been pushed in a major version release based on a breaking change; see https://github.com/yaronn/xml-crypto/releases

PR #1200, opened 2022-10-17
Bump xml-crypto

benasher44:
This gets the downstream `@xmldom/xmldom` update to resolve another dependabot alert.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "strip-bom": "^3.0.0",
         "uuid": "^8.3.2",
         "whatwg-mimetype": "3.0.0",
-        "xml-crypto": "^2.1.3"
+        "xml-crypto": "^3.0.0"
       },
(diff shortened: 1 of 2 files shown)


benasher44:
@vpulim this is done in #1199 as well. Only difference is that now that I'm a committer, CI has already passed here. Will let you decide next steps!

benasher44:
@jsdevel this is done in #1199 as well. Only difference is that now that I'm a committer, CI has already passed here. Will let you decide next steps!

RopoMen:
This xml-encryption update may potentially contain BREAKING CHANGES because
1. they self update major version
2. @xmldom/xmldom states that 0.8.x has BREAKING CHANGES https://github.com/xmldom/xmldom/blob/master/CHANGELOG.md#fixed-6

ellisium:
still warm topic?

PR #1202, opened 2022-11-16
Fix including xsd from another xsd while using inline xmlns

e6c31d:


jsdevel:
@e6c31d plz get the build to pass

e6c31d:
@jsdevel I don't think the build failure is related to this PR. It builds successfully for me on Node.js 14 and 16, but not on 18. The build also fails on the master branch on Node.js 18.

e6c31d:
@jsdevel I created another PR to fix the build on Node.js 18: #1206

w666:
Hi @e6c31d,

I rebased you branch to check if it still works, I recently released new version with some major changes. 

I will merge it and include into next release.

w666 (approved):

PR #1205, opened 2022-12-06
Fix names only in soap messages.

alex-eri:
Without namespace check this function fails with `Exception has occurred: TypeError: Cannot read properties of undefined (reading 'description')` if Message>Body tags deepper inside answer.

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -232,7 +232,7 @@ export class WSDL {
       const obj = {};
       const originalName = name;
 
-      if (!objectName && top.name === 'Body' && name !== 'Fault') {
+      if (!objectName && xmlns.soap && top.name === 'Body' && name !== 'Fault') {
         let message = this.definitions.messages[name];
         // Support RPC/literal messages where response body contains one element named
         // after the operation + 'Response'. See http://www.w3.org/TR/wsdl#_names


pestig (approved):
I had a same problem with deep Body tag. It's fixed my problem also.

w666:
Hi @alex-eri @pestig 

I am happy to merge it and include into next release, it just requires a test. If you can add one that would be great.

w666 (changes requested):
Please add a unit test

w666:
I rebased this, but some tests failed. I was about to approve it without a test but seems like this change breaks something.

smokhov:
Weird. Those appear timeout errors though. Possible infra issues of GitHub? Maybe can simply re-run the job in GitHub's GUI?

w666:
The test suite is a bit weird, it does log errors as timeouts, debugging it is a kinda pita.

smokhov:
it seems the errors are in `WSSecurityPlusCert.ts`. When running `npm cover` locally I get the same exact errors even without the change proposed here by removing it.

smokhov:
Very weird. Now most tests on my local system pass. It went down from 11  to 1 (`WSSecurityCert` rsa-sha1), then to 2 not passing ("before all" hook for "lastElapsedTime is computed"). And this includes master without the proposed change. It seems almost each time I run `npm run cover` the numer of tests failing changes. Something is off here...

I get no more time out errors (previously I tried with the timeout settings, as well as catcbing `done` (per [this post](https://stackoverflow.com/questions/44149096/for-async-tests-and-hooks-ensure-done-is-called-if-returning-a-promise-en)), but that led to more errors that I could not make sense of.

My current failures locally are:

```
  587 passing (3s)
  2 pending
  2 failing

  1) SOAP Client
       "before all" hook for "lastElapsedTime is computed":
     Uncaught Error: listen EADDRINUSE: address already in use :::54454
      at Server.setupListenHandle [as _listen2] (node:net:1908:16)
      at listenInCluster (node:net:1965:12)
      at Server.listen (node:net:2067:7)
      at Context.<anonymous> (test/client-response-options-test.js:33:12)
      at process.processImmediate (node:internal/timers:483:21)

  2) WSSecurityCert
       should use rsa-sha1 signature method when the signatureAlgorithm option is set to WSSecurityCert:
     Error: error:03000098:digital envelope routines::invalid digest
      at Sign.sign (node:internal/crypto/sig:128:29)
      at node-soap/node_modules/xml-crypto/lib/signature-algorithms.js:11:32
      at RsaSha1.getSignature (node_modules/xml-crypto/lib/types.js:52:20)
      at SignedXml.calculateSignatureValue (node_modules/xml-crypto/lib/signed-xml.js:327:42)
      at SignedXml.computeSignature (node_modules/xml-crypto/lib/signed-xml.js:763:18)
      at WSSecurityCert.postProcess (lib/security/WSSecurityCert.js:2:10622)
      at Context.<anonymous> (test/security/WSSecurityCert.js:285:26)
      at process.processImmediate (node:internal/timers:483:21)
```

but yesterday (1) was not there. And I changed nothing in this instance... (2) may be related to the recent update of `xml-crypto` in 0b159d17ab47869a40625657c2dee9415c4fc207 but that should have been caught sooner than now.

w666:
@smokhov 

Well, tests on main are fairly stable. This branch never worked for me.

PR #1209, opened 2022-12-25
honor httpAgent/httpsAgent when setting ntlm flag

unilynx:
Passing httpAgent/httpsAgent inside `wsdl_options` works until you add `ntlm: true` to the `wsdl_options`. This PR fixes this.


smokhov:
@w666 -- please force checks run on this

PR #1210, opened 2023-01-26
feat: add express request object as parameter to the log method.

jplusje:
This PR add's the express request object as a parameter to the log function.
With this change it's possible to (for example) create a unique id for the request and use it when performing logging in node-soap.


w666 (approved):

w666:
Looks good to me, thanks for the readme update. I will merge it when will be preparing new release.

PR #1212, opened 2023-02-10
updated axios and axios-ntlm to latest

vision10:
latest version of axios-ntlm fixes node v18 digital envelope error
Error: error:0308010C:digital envelope routines::unsupported

updated axios to latest: 1.3.2
updated axios-ntlm to latest: 1.4.0 (uses axios 1.3.2)

also 2 tests where failing because of hardcoded response Content-Length

this request needs to be accepted to pass all tests: https://github.com/vpulim/node-soap/pull/1206

PR #1216, opened 2023-03-28
Improve trim speed during XML parsing.

shellicar:
* Use native string.trim instead of regex replace for performance.
* Add tests surrounding trim function used in wsdl parsing.

## Issue with parsing large XML responses
I recently noticed an issue with slow parsing of large (1-2mb+) responses.
I narrowed the issue down to the trim(text) method in wdsl/index.ts
This uses regex to remove whitespace characters at the start and end of a string.
This has performance implications when the string is very long, as I believe the trimRight regex needs to scan the entire content before reaching the end and then backtrack.

This issue was so bad it was causing requests to timeout after (10+ minutes).

## Solution
I couldn't see any other tests testing the whitespace removal, so I added some.
I validated that it was extremely slow before the change (I gave up waiting after around 5 minutes).
I validated that after the change it was fast (~10ms).
And that the expected output is the same.

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -23,11 +23,8 @@ const debug = debugBuilder('node-soap');
 
 const XSI_URI = 'http://www.w3.org/2001/XMLSchema-instance';
 
-const trimLeft = /^[\s\xA0]+/;
-const trimRight = /[\s\xA0]+$/;
-
-function trim(text) {
-  return text.replace(trimLeft, '').replace(trimRight, '');
+export function trim(text) {
+  return text.trim();
 }
 
 function deepMerge<A, B>(destination: A, source: B): A & B {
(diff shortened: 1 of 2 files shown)


el-noppo (approved):
LGTM

frebos88 (approved):

cgerikj (approved):

frebos88:
Tested the PR out and also saw a huge improvement in performance. 👌

shellicar:
Just wondering if this project is being maintained anymore?
I don't see any activity from the listed maintainers.

Another PR that seems quite important and supported isn't merged either (another issue I just ran into which brought me back here) - #1212 

I ended up just forking the project and hosting it on my own package feed, but I figured there would have been some activity in the past few months.

w666:
Good one, thanks for submitting it, I will include in the next release.

PR #1218, opened 2023-05-04
Speed up WSDL parsing

tomquist:
We were running into performance issues while generating a SOAP client. I profiled the client generation and identified two bottlenecks in parsing WSDL files:
- When visiting an unexpected node an error got constructed that was discarded later on in non-strict mode. Constructing an error is pretty expensive because Node needs to capture the stack.
- splitQName split the entire nsName string even though it's only interested in the first element

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -65,6 +65,7 @@ export class Element {
   public $targetNamespace?;
   public children: Element[] = [];
   public ignoredNamespaces;
+  public strict: boolean;
   public name?: string;
   public nsName?;
   public prefix?: string;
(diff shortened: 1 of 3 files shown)


test/client-test.js
```diff
@@ -500,7 +500,7 @@ var fs = require('fs'),
             assert.ok(result);
             assert.ok(client.lastResponse);
             assert.ok(client.lastResponseHeaders);
-            assert.ok(client.lastElapsedTime);
+            assert.ok(client.lastElapsedTime !== undefined);
```
tomquist, line 503: This seems to be flaky on my machine because it's 0 sometimes (and more likely with this change), which is falsy. However, we're only interested in the existence of the value.

src/wsdl/elements.ts
```diff
@@ -171,14 +175,24 @@ export class Element {
...
     }
   }
 }
 
+export class UnexpectedElement extends Element {
+  public startElement(stack: Element[], nsName: string, attrs, options: IWsdlBaseOptions, schemaXmlns) {
+    const child = new UnexpectedElement(nsName, attrs, options, schemaXmlns);
+    child.init();
+    stack.push(child);
+  }
```
tomquist, line 193: This short-circuits the `allowedChildren` lookup.

w666 (approved):

w666:
Changes look good. I guess you run some performance tests, how much faster parsing now?

PR #1219, opened 2023-05-17
feature (xml key): allow in first level for rpc

jglongaron:
* allow xmlKey in first level for rpc

# Problem
Using xmlKey ('$xml') in the first level of args

```js
var soap = require('soap');
  var args = {$xml: '<action>1</action>'};

  soap.createClient(url, {}, function(err, client) {
      client.MyFunction(args, function(err, result) {
          ...
      });
  });
```

Request includes "$xml" tag instead of unescaped text

```xml
...<soap:body><MyFunction><$xml>&lt;tns:action&gt;&lt;...&lt;tns:/action&gt;&lt;</$xml></MyFunction></soap:body>...
```

# New Request 

```xml
...<soap:body><MyFunction><action>1</action></MyFunction></soap:body>...
```

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -604,17 +604,21 @@ export class WSDL {
       }
       if (key !== nsAttrName) {
         const value = params[key];
-        const prefixedKey = (isParts ? '' : nsPrefix) + key;
-        const attributes = [];
-        if (typeof value === 'object' && value.hasOwnProperty(this.options.attributesKey)) {
-          const attrs = value[this.options.attributesKey];
-          for (const n in attrs) {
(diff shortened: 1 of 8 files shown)


smokhov:
@w666 -- this is one also has tests and no merge conflicts; can re-run the check to see if the current ones pass

w666:
Interesting, I don't see that new test in the build output.

smokhov:
It's there, line 375-376:
```
    ✔ should use defined xml key in first level for rpc
    ✔ should use defined xml key in first level for rpc (promisified)
```

w666:
Ah, I see, I guess I was too tired when was looking on it last time. Will include this on onto next release. Thanks.

PR #1226, opened 2023-08-30
fix: serviceName and portName must correspond to port binding

paulish:
Soap server code makes an assumption that url may not match the port binding. In this case the first port is used. At the same time WSDL can have multiply services and ports and therefore not only the first port binding should be taken but also first port name and first service name. Without this patch the first port binding is used and the last serviceName and portName. With this patch applied port binding corresponds correct serviceName and portName.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -298,10 +298,7 @@ export class Server extends EventEmitter {
     const obj = this.wsdl.xmlToObject(input);
     const body = obj.Body;
     const headers = obj.Header;
-    let binding: BindingElement;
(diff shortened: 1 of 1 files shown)


smokhov:
@w666 -- could you force checks run for this?

w666:
https://github.com/vpulim/node-soap/pull/1337 adds support for this. It was not a bug, it just was not implemented. 
Also, changes in this PR do not fix the problem mentioned in the description. At least it did not work with the tests I created to replicate the issue.

Closing.

PR #1227, opened 2023-10-05
Adding advanced logging capability

G3RMAN:
For cases where is needed log xml text with transactional information allocated in headers for example


w666:
Hi @G3RMAN,

Seems like this is already implemented, could you please double check it?

PR #1228, opened 2023-10-27
Try to make error messages useful when using SOAP 1.2

nraynaud:
When using SOAP 1.2 the Fault error messages were `Uncaught Error: undefined: undefined`.

They are now of the form `Uncaught Error: s:Sender: w:InternalError: The WS-Management service cannot process the request because the XML is invalid.: {"FaultDetail":"http://schemas.dmtf.org/wbem/wsman/1/wsman/faultDetail/Invalid","WSManFault":{"attributes":{"Code":"2150858819","Machine":"dell"},"Message":"WinRM cannot process the request because the input XML uses an undefined XML namespace."}}`, while verbose, it's actually useful.


w666 (approved):

w666:
Yeah, this makes sense, approved, I will merge when will be preparing new release. Thanks for the improvement.

PR #1230, opened 2024-01-11
typeStorageを確実に参照するようにする

forcia-tinoue:
alteaに対するxmlToObjectで配列型を配列として変換してくれない問題を暫定的に解消。
node-soap的に望ましい実装ではなさそうだが...。

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -225,13 +225,22 @@ export class ElementElement extends Element {
     let type: any = this.$type || this.$ref;
     if (type) {
       type = splitQName(type);
+      const typeStorage = this.$type ? definitions.descriptions.types : definitions.descriptions.elements;
       const typeName: string = type.name;
+      if(typeName in typeStorage){// typeにtypeNameが入ってしまった場合の対応、本来はこのようなことが起きないように対処すべきかもしれない。
+        if (this.$ref) {
+          element = typeStorage[typeName];
+        }
+        else {
+            element[name] = typeStorage[typeName];
+        }
+        return element;
+      }
       const ns: string = xmlns && xmlns[type.prefix] ||
         ((definitions.xmlns[type.prefix] !== undefined || definitions.xmlns[this.targetNSAlias] !== undefined) && this.schemaXmlns[type.prefix]) ||
         definitions.xmlns[type.prefix];
       const schema = definitions.schemas[ns];
       const typeElement = schema && (this.$type ? schema.complexTypes[typeName] || schema.types[typeName] : schema.elements[typeName]);
-      const typeStorage = this.$type ? definitions.descriptions.types : definitions.descriptions.elements;
 
       if (ns && definitions.schemas[ns]) {
         xmlns = definitions.schemas[ns].xmlns;


w666:
To make it easier to understand could you translate this to English and explain why do you need this change? Also, if it is possible please cover changes with the tests.

smokhov:
From deepl:

"""
Ensure that typeStorage is referenced

Tentatively solved the problem that xmlToObject for ltea does not convert array types as arrays.
This may not be a desirable implementation from a node-soap point of view.
"""

Not sure if that helps :shrug: but I was curious :)

src/wsdl/elements.ts
```diff
@@ -225,13 +225,22 @@ export class ElementElement extends Element {
     let type: any = this.$type || this.$ref;
     if (type) {
       type = splitQName(type);
+      const typeStorage = this.$type ? definitions.descriptions.types : definitions.descriptions.elements;
       const typeName: string = type.name;
+      if(typeName in typeStorage){// typeにtypeNameが入ってしまった場合の対応、本来はこのようなことが起きないように対処すべきかもしれない。
```
smokhov, line 230: "// typeName in type, perhaps this should be handled to prevent this from happening in the first place."

src/wsdl/elements.ts
```diff
@@ -225,13 +225,22 @@ export class ElementElement extends Element {
...
+        if (this.$ref) {
+          element = typeStorage[typeName];
```
w666, line 232: Probably, need to check that `typeStorage[typeName]` has a value, could be undefined, null ... anything

w666 (changes requested):
Needs a test. While code makes sense, implementation should be locked, so other changes do not break this,

smokhov:
Upon reading further, unless I don't understand something, I believe actually the existing code already covers what the PR proposes just further down a few lines. Specifically the check's if() and the else() and the value is set properly.

https://github.com/vpulim/node-soap/blob/9c6ad017e61899b6912b1dabc7fc93b43bfb8285/src/wsdl/elements.ts#L261

https://github.com/vpulim/node-soap/blob/9c6ad017e61899b6912b1dabc7fc93b43bfb8285/src/wsdl/elements.ts#L292

If my understanding is correct, I think this PR can simply be closed.

smokhov:
@w666 -- per my earlier reasoning the current code I think covers what was proposed here due to another PR that got merged and covers element values in different configurations;  I think it was min/maxOccurs PR that refactored that code. Please review if you agree with me and simply close this PR if you do. I don't think there is much else to do here.

src/wsdl/elements.ts
smokhov: The current code guarantees this that `typeStorage[cacheKey]` (formerly `typeStorage[typeName]`) is set.

```javascript
        if (!(cacheKey in typeStorage)) {
          let elem: any = {};
          typeStorage[cacheKey] = elem;

          if (isMany && 'maxOccurs' in typeElement) {
            typeElement.maxOccurs = this.$maxOccurs;
          }
          if (Boolean(this.$minOccurs) && 'minOccurs' in typeElement) {
            typeElement.minOccurs = this.$minOccurs;
          }

          const description = typeElement.description(definitions, xmlns);
          if (typeof description === 'string') {
            elem = description;
          } else {
            Object.keys(description).forEach((key) => {
              elem[key] = description[key];
            });
            
            const $attributes = description[AttributeElement.Symbol];
            if ($attributes) {
              elem[AttributeElement.Symbol] = $attributes;
            }
          }

          if (this.$ref) {
            element = elem;
          } else {
            element[name] = elem;
          }

          if (typeof elem === 'object') {
            elem.targetNSAlias = type.prefix;
            elem.targetNamespace = ns;
          }
          
          typeStorage[cacheKey] = elem;
        } else {
          if (this.$ref) {
            // Differentiate between a ref for an array of elements and a ref for a single element
            if (isMany) {
              const refTypeName = typeName + '[]';
              const refCacheKey = ns ? `${ns}::${refTypeName}` : refTypeName;
              typeStorage[refCacheKey] = typeStorage[cacheKey];
              element[refTypeName] = typeStorage[refCacheKey];
            } else {
              element = typeStorage[cacheKey];
            }
          } else {
            element[name] = typeStorage[cacheKey];
          }
        }
```

smokhov:
I was trying to see if this PR still has any merit and can be salvaged, but the current code covers everything. In fact the PR has reduced basically to a single line move:
- https://github.com/smokhov/node-soap/pull/115
- Additionally, PR #1397 should have actually closed this.

w666:
@smokhov seems like this can be just closed as not needed anymore?

smokhov:
> @smokhov seems like this can be just closed as not needed anymore?

Precisely.

w666:
Closing as not required anymore

PR #1238, opened 2024-04-26
Prevent mutating $type in the schema while processing requests

simhnna:
I'm relying on the wsdl for typechecking and requests mutate `$type`.

I don't really understand much about the code, but with my change `$type` is no longer changed and all tests still pass.

This behavior was introduced with an old [commit](https://github.com/vpulim/node-soap/commit/9b6ba1fa39a4c15abf71d108cc7d159d5a8e3645) and I would consider changing the schema during request processing to be a bad habit. Ideally the namespaces would be set initially during schema parsing and not touched later.

### DIFF
--- a/test/client-schema-does-not-change-on-request-test.js
+++ b/test/client-schema-does-not-change-on-request-test.js
@@ -1,0 +1,125 @@
+const fs   = require('fs');
+const soap = require('../');
+const path = require('path');
+var http = require('http');
+var jsdiff = require('diff');
+var assert = require('assert');
+
+let server;
+let port;
+
+function normalizeWhiteSpace(raw)
+{
+  var normalized = raw.replace(/\r\n|\r|\n/g, '');  // strip line endings
+  normalized = normalized.replace(/\s\s+/g, ' '); // convert whitespace to spaces
+  normalized = normalized.replace(/> </g, '><');  // get rid of spaces between elements
+  return normalized;
+}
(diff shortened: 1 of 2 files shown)


test/client-schema-does-not-change-on-request-test.js
```diff
@@ -0,0 +1,125 @@
...
+
+function cbCaller(client, methodName, requestJSON, responseJSON, responseSoapHeaderJSON, options, attachmentParts, done){
+    client[methodName](requestJSON, function(err, json, body, soapHeader){
+      if(requestJSON){
+        if (err) {
+          assert.notEqual('undefined: undefined', err.message);
+          assert.deepEqual(err.root, responseJSON);
+        } else {
+          // assert.deepEqual(json, responseJSON);
+          assert.equal(JSON.stringify(typeof json === 'undefined' ? null : json), JSON.stringify(responseJSON));
```
w666, line 113: I am wondering what was the reason to use stringify and equal instead of deepEqueal?

simhnna: That's shamelessly copied from the request/response tests...

w666: Does it work with the deepEqual?

simhnna: done

PR #1239, opened 2024-05-16
add envelopeSoapUrl option

mkwiecinski:
This adds an option to change the URL in `xmlns:soap` attribute.

Some services require this Url to be changed. Until now, it was hardcoded.

Also requested on SO: https://stackoverflow.com/questions/68483250/node-soap-remove-xmlns-attribute-from-method


w666:
Hi @mkwiecinski, could you please add a test for that? While change seems simple we need a way to make sure it will not be broken in the future. Thanks.

mkwiecinski:
@w666 Sorry for the lack of tests. Let me know if this is enough.

w666 (approved):

w666:
Looks good, thanks for the test. I will include it to the next release.

PR #1242, opened 2024-06-14
Upgrade dependencies and refactor code to work with the latest xml-crypto

w666:
Refactor WSSecurityCert and WSSecurityCertWithToken to make it
 work with the latest xml-crypto.

 Fix invalid test xml strings in WSSecurityCert and
 WSSecurityCertWithToken tests.

 Update dependencies:
 - axios to 1.7.2
 - debug to 4.3.5
 - formidable to 3.5.1
 - sax to 1.4.1
 - uuid to 10.0.0
 - whatwg-mimetype to 4.0.0
 - xml-crypto to 6.0.0

 Update dev dependencies:
 - @types/debug to 4.1.12
 - @types/express to 4.17.21
 - @types/lodash to 4.17.5
 - @types/node to 14.0.0
 - @types/request to 2.48.12
 - @types/sax to 1.2.7
 - @types/uuid to 9.0.8
 - @types/whatwg-mimetype to 3.0.2
 - body-parser to 1.20.2
 - colors to 1.4.0
 - diff to 5.2.0
 - doctoc to 2.2.1
 - duplexer to 0.1.2
 - express to 4.19.2
 - finalhandler to 1.2.0
 - glob to 10.4.1
 - jshint to 2.13.6
 - mocha to 10.4.0
 - nyc to 17.0.0
 - semver to 7.6.2
 - serve-static to 1.15.0
 - sinon to 18.0.0
 - source-map-support to 0.5.21

PR #1243, opened 2024-06-14
updated packages

vision10:
I have updated most of the outdated packages that are relatively safe and dont break tests to latest

also removed types:
@types/axios - axios provides types
@types/request - no request package

`sinon` and `xml-crypto` require code changes

`whatwg-mimetype` and `get-stream` require min node version: 18
I think `axios-ntlm` also requires node v18
`get-stream` - latest version is now pure ESM. It cannot be require()d from CommonJS

`stip-bom`
stip-bom@4 requires esModuleInterop flag
stip-bom@5 requires entire package tuned to work with modules
my sugestion: dont use stip-bom package, make function in utilities


w666:
I raised a PR with the updated deps, including xml-crypto. But planning to keep it in PR for some time just to get feedback if there is any.

https://github.com/vpulim/node-soap/pull/1242

w666:
and I happy to get rid of strip-bom.

PR #1246, opened 2024-07-18
Pass the error object to log

ikonst:
The current code uses `suppressStack` to determine what is logged, while I think `suppressStack` was only intended to control whether we expose internals to the caller.

It would be preferable if the logging mechanism could see the original error for what it is, since then it could use the appropriate logging mechanism in the code base; for example, pass the error object to Sentry, or log it in a structured way (name and stack separately) [for Datadog to consume](https://docs.datadoghq.com/standard-attributes/), etc.

### DIFF
--- a/test/server-test.js
+++ b/test/server-test.js
@@ -463,9 +463,6 @@ describe('SOAP Server', function () {
         assert.ok(err);
         assert.strictEqual(err.response, response);
         assert.strictEqual(err.body, body);
-        // console.log(test.logMessages);
-        // assert.strictEqual(test.logMessages.length, 1);
-        // assert.notStrictEqual(test.logMessages[0].tyoe, {});
         done();
       });
     });
--- a/test/server-test.js
+++ b/test/server-test.js
@@ -100,9 +100,14 @@ describe('SOAP Server', function () {
       res.end();
     });
 
+    test.logs = [];
+
     test.server.listen(15099, null, null, function () {
       var testSv = test.server.address();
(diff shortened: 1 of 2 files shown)


w666:
Hi @ikonst,

Could you please add a test for this? Change looks good me.

ikonst:
@w666 take a look ^

test/server-test.js
```diff
@@ -444,9 +461,11 @@ describe('SOAP Server', function () {
...
       client.GetLastTradePrice({ tickerSymbol: 'trigger error' }, function (err, response, body) {
         assert.ok(err);
-        // what happens here (error and response ????)
-        // assert.strictEqual(err.response, response);
-        // assert.strictEqual(err.body, body);
+        assert.strictEqual(err.response, response);
+        assert.strictEqual(err.body, body);
+        // console.log(test.logMessages);
+        // assert.strictEqual(test.logMessages.length, 1);
+        // assert.notStrictEqual(test.logMessages[0].tyoe, {});
```
w666, line 468: I guess this and two lines above were used for debugging?  It would be good to remove unused code to keep it clean.

w666 (changes requested):
One minor comment, otherwise looks good to me.

w666 (approved):

w666 (approved):

w666:
Approved, thanks for the improvement, I will include it into next release. 

Please let me know if you need these changes asap, will see if I can make a release earlier.

ikonst:
Thank you. There's no urgency.

PR #1247, opened 2024-07-22
Totally remove lodash

JbIPS:
Replace lodash functions by ESNext style functions or by deepmerge-ts (lighter and more maintened)


pnpm-lock.yaml
w666, line 1: We use `npm`, please remove this file and run `npm install` so `package-lock` is in sync.

JbIPS: Sorry I missed that. It's fixed

w666:
Seems like tests do not quite work.

w666:
I was thinking about this change and don't see any benefits in removing lodash. While it was not updated for 4 years, seems like version 5 will be released at some point. 

After my recent changes node-soap does not contain any vulnerabilities.

If you have different point of view ii I happy to discuss it.

JbIPS:
I will update this PR after summer break.

Concerning the why, I have 2 reasons to remove this kind of dependency:
- It can be considered as "unmaintained" after 4 years, and I personally doubt it will have a rebirth. The discovery of a vuln seems bound to happen
- Most of the useful functions are now in the standard API
- The package weight 1.41MB and you're just cherry-picking a few functions

I understand those are very opinionated reasons and I'll understand if you do not wish to remove it.

w666:
Hi @JbIPS,

Okay, happy to discuss proposed solution.

w666:
I had another look on it ... that is not that easy, deepmerge-ts is not enough. Some stuff can be replaced with native JS features, but not all. Too much work to replace what works just fine.

smokhov:
Prior work on this: #1122

w666:
Closing as not active. Will create separate PR with different approach.

PR #1249, opened 2024-08-18
Speed up tests execution

w666:
- Override endpoints in some tests where it tried to send requests to the endpoints in the internet
 - Now tests work in offline


w666:
This PR is mostly to make life a bit easier for those who develop and run tests in this project.
Now it is possible to run tests offline, for me as a maintainer it was very important as I work on this project on my way to work with a bad network coverage.
Some tests do not try to send requests to hosts that do not exist anymore.
Some tests relied on error response from real hostnames. 

On my machine runtime was ~3 minutes before changes, now ~5 seconds.

PR #1254, opened 2024-09-09
Feature inform signaturealgorithm

JoseTM:
default algorithm seem to be removed  from xml-crypto; so now must it must be specified. WSSecurityCert is not setting singnatureAlgorithm actually so just rsa-sha256 is been used. I tried to fix this behaviour to be able of setting signature algorithm.

### DIFF
--- a/test/security/WSSecurityCert.js
+++ b/test/security/WSSecurityCert.js
@@ -234,4 +234,22 @@ describe('WSSecurityCert', function () {
     }
     should(xml).not.be.ok();
   });
+   
+  it('should use rsa-sha1 signature method when the signatureAlgorithm option is set to WSSecurityCert', function () {
+      var instance = new WSSecurityCert(key, cert, '', {
+         hasTimeStamp: false,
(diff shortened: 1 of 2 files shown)


test/security/WSSecurityCert.js
w666, line 1: Unfortunately, it is very hard to review changes in this file. I agree that code formatting is not the best in this project, but I don't think it should be addressed in this PR. Could you please revert code-reformatting and leave just actual test changes.

JoseTM: Yes,  of course. And I apollogize

w666 (approved):

w666:
@JoseTM let me know if you need this feature to be released asap, if no I will do at my normal release schedule in 1-2 weeks.

JoseTM:
In your normal workflow is fine. Pleased to contribute and thanks for your job

SzymonNadbrzezny:
Good to know it's going to be released soon. :D

w666:
1.1.4 has been released

PR #1255, opened 2024-09-14
WSSecurityCert KeyInfo tag wrapped around  SecurityTokenReference

somdeepjana:
`WSSecurityCert` `<wsse:SecurityTokenReference>` tag should be wrapped inside `<KeyInfo>` tag.

reference: https://docs.oasis-open.org/wss/v1.1/wss-v1.1-spec-pr-x509TokenProfile-01.htm

```xml
<wsse:Security>
  <ds:Signature>
  
    <ds:SignedInfo>...</ds:SignedInfo>
    <ds:SignatureValue>...</ds:SignatureValue>

    <ds:KeyInfo Id="keyinfo"> <---------------------------- This tag was missing

       <wsse:SecurityTokenReference>...</wsse:SecurityTokenReference>
    </ds:KeyInfo>

  </ds:Signature>
</wsse:Security>
```

### DIFF
--- a/src/security/WSSecurityCert.ts
+++ b/src/security/WSSecurityCert.ts
@@ -58,6 +58,7 @@ export class WSSecurityCert implements ISecurity {
   private publicP12PEM: string;
   private signer: any;
   private signerOptions: IXmlSignerOptions = {};
+  private keyInfoId: string;
   private x509Id: string;
   private hasTimeStamp: boolean;
   private signatureTransformations: string[];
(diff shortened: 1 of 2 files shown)


src/security/WSSecurityCert.ts
```diff
@@ -111,15 +112,19 @@ export class WSSecurityCert implements ISecurity {
       key: privatePEM,
       passphrase: password,
     };
+    this.keyInfoId = `KI-${generateId()}`;
     this.x509Id = `x509-${generateId()}`;
     this.hasTimeStamp = typeof options.hasTimeStamp === 'undefined' ? true : !!options.hasTimeStamp;
     this.signatureTransformations = Array.isArray(options.signatureTransformations) ? options.signatureTransformations
       : ['http://www.w3.org/2000/09/xmldsig#enveloped-signature', 'http://www.w3.org/2001/10/xml-exc-c14n#'];
 
     this.signer.getKeyInfo = (key) => {
```
w666, line 121: Looks like I missed this when was migrating to the latest XML-Crypto. I think you can set `getKeyInfoContent` and it will add `KeyInfo`. 

```suggestion
    this.signer.getKeyInfoContent = (key) => {
```

src/security/WSSecurityCert.ts
```diff
@@ -111,15 +112,19 @@ export class WSSecurityCert implements ISecurity {
       key: privatePEM,
       passphrase: password,
     };
+    this.keyInfoId = `KI-${generateId()}`;
```
w666, line 115: I don't think that `id` in `KeyInfo` is just a random string. My understanding from the spec is that this is an id that references element from `SignedInfo`.

test/security/WSSecurityCert.js
```diff
@@ -54,6 +54,7 @@ describe('WSSecurityCert', function () {
     xml.should.containEql('<Signature xmlns="http://www.w3.org/2000/09/xmldsig#">');
     xml.should.containEql('<wsse:SecurityTokenReference');
     xml.should.containEql('<wsse:Reference URI="#' + instance.x509Id);
+    xml.should.containEql('<KeyInfo Id="' + instance.keyInfoId);
```
w666, line 57: I think test should check both. opening tag, like you already do, but closing tag `</KeyInfo>` as well.

src/security/WSSecurityCert.ts
somdeepjana: used `getKeyInfoContent` from `xml-crypto` as requested.

src/security/WSSecurityCert.ts
somdeepjana: removed `keyInfoId` creation. letting `xml-crypto` handle it.

test/security/WSSecurityCert.js
somdeepjana: added starting and ending tag checks in the latest commit.

w666 (approved):

PR #1257, opened 2024-09-27
overrideImportLocation try/catch

simakvladimir:
Hi!

# Problem:

The node-soap library allows overriding the import location of WSDL files by defining an `overrideImportLocation` function in the options object. However, if the code within this function throws an error, the entire application crashes. This is because `overrideImportLocation` is invoked in the `process.nextTick` queue, and the `createClient` callback is not invoked.

Furthermore, in our project, we require complex logic to determine the import location. This necessitates the ability to throw user-defined exceptions within the `overrideImportLocation` function. It would be highly beneficial if these exceptions were returned by the client callback function.

# Proposed solution:

To address these issues, we propose the following:

1. Wrap `overrideImportLocation` in a try-catch block: This will ensure that any errors thrown within the function are caught and handled gracefully.

2. Extend `overrideImportLocation` arguments: instead of returning only the location (for backward compatibility), it should also return the URI, include path, and options. This extended set of arguments facilitates the implementation of complex logic to determine the sub-WSDL location. The options parameter can be used to redefine specific parameters for the sub-WSDL.

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -1209,14 +1209,18 @@ export class WSDL {
       includePath = url.resolve(this.uri || '', include.location);
     }
 
-    if (this.options.wsdl_options !== undefined && typeof this.options.wsdl_options.overrideImportLocation === 'function') {
-      includePath = this.options.wsdl_options.overrideImportLocation(includePath);
-    }
-
     const options = Object.assign({}, this.options);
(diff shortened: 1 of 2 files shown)


w666 (approved):
Yeah, makes sense. Thanks for the test.

w666:
Just released new version 1.1.5

PR #1261, opened 2024-10-31
Add custom WSDL cache option

millerdk12:
Currently the library caches WSDLs for the lifetime of the process. This is not ideal for long running processes like a backend service that connect to many SOAP webservices.

Adding an option to override the current lifetime cache with a custom cache so user can bring their own cache, e.g. TTL or LRU. I had considered just disabling the cache using the existing option, but this is not also not ideal.

Code changes:
* Adding `IWSDLCache` interface which can be extended with custom caching implementations
* Added `DefaultWSDLCache` implementation which matches current caching handling, ie global singleton which indefinitely caches WSDLs.
* Added `options.wsdlCache` to allow users to override the default cache.
* Added test cases for default cache and overriding option.

I also tested with npm link against an existing application that uses node-soap. By adding ?v=1, ?v=2 and so on to the WSDL URL, you can simulate loading many different WSDLs. Setting ` max-old-space-size` to something like 128, eventually the process will crash with the default cache. For my application, this occurred around 1000th WSDL with a 128mb limit. I updated to use a cache with a set max item limit and memory remained low and the process did not crash even when reaching the 1000 WSDL loaded mark.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -111,6 +111,7 @@ GitHub issues have been disabled to focus on pull requests. ([#731](https://gith
   - `wsdl_headers` (*Object*): Set HTTP headers with values to be sent on WSDL requests.
   - `wsdl_options` (*Object*): Set options for the request module on WSDL requests. If using the default request module, see [Request Config | Axios Docs](https://axios-http.com/docs/req_config).
   - `disableCache` (*boolean*): Prevents caching WSDL files and option objects.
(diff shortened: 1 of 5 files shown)


w666 (approved):

w666:
Hi @millerdk12,

Thanks for the improvement. I will include this into next release.

millerdk12:
Great, thanks for taking a look!

w666:
@millerdk12 version 1.1.6 has been released

PR #1273, opened 2025-02-11
Add option for digest algorithm

pzemljic-git:
Currently, the digest algorithm is hard coded to sha256 even when passing a different signature algorithm like sha1. However, some web services require differnt digest algorithms. This change adds a new option for passing the digest algorithm:

  client.setSecurity(new soap.WSSecurityCert(
    privateKey,
    publicKey,
    '',
    {
      signatureAlgorithm: 'http://www.w3.org/2000/09/xmldsig#rsa-sha1',
      **digestAlgorithm**   : 'http://www.w3.org/2000/09/xmldsig#sha1',
    }
  ));


w666:
HI @pzemljic-git,

I think default value is required, and it must be the same as the one currently used. As you can see tests are failing atm.

Could you also add a test that default value can be overridden?
Maybe doc changes are needed as well, but I can help with that,

Thanks.

pzemljic-git:
Hi @w666,

Good point, thanks! I added the default value, added a test, and updated the documentation.

Best regards,
Peter

w666:
Looks good to me, thanks for the contribution.
When do you need new release?

pzemljic-git:
No rush on the release. I can use my fork for now. Thanks for merging and feedback!

w666:
1.1.8 has been released.

Issue #1276, opened 2025-02-25
conflicting import namespace prefixes


icholy:
### Context:

I have a root wsdl file imports 2 additional wsdl files:

**root.wsdl**

```xml
<?xml version="1.0" encoding="utf-8"?>
<wsdl:definitions
    xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
    xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/"
    xmlns:tns="http://tempuri.org"
    name="TestService"
    targetNamespace="http://tempuri.org">

  <wsdl:import namespace="foo" location="foo.wsdl"/>
  <wsdl:import namespace="bar" location="bar.wsdl"/>

  <wsdl:service name="TestService">
    <wsdl:port name="TestPort" binding="tns:TestBinding">
      <soap:address location="http://example.com/TestService"/>
    </wsdl:port>
  </wsdl:service>
</wsdl:definitions>
```

The **bar.wsdl** file defines the `xmlns:tns="bar"`:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<wsdl:definitions
    xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
    xmlns:tns="bar"
    targetNamespace="bar">
</wsdl:definitions>
```

The **foo.xml** file defines `xmlns:tns="foo"`:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<wsdl:definitions
    xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
    xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/"
    xmlns:tns="foo"
    targetNamespace="foo">

  <wsdl:types>
    <xsd:schema targetNamespace="foo" elementFormDefault="qualified">
      <xsd:element name="TestElement">
        <xsd:complexType>
          <xsd:sequence>
            <xsd:element name="val" type="xsd:string" minOccurs="0"/>
          </xsd:sequence>
        </xsd:complexType>
      </xsd:element>
    </xsd:schema>
  </wsdl:types>

  <wsdl:message name="TestInputMessage">
    <wsdl:part name="params" element="tns:TestElement"/>
  </wsdl:message>

  <wsdl:message name="TestOutputMessage">
    <wsdl:part name="params" element="tns:TestElement"/>
  </wsdl:message>

[...]

PR #1278, opened 2025-02-28
keep track of effective xmlns for each element

icholy:
This is an attempt at fixing #1276. It adds a new `effectiveXmlns` property to each `Element` type which keeps track of the namespace mapping at the time the element was defined.

I'm pretty confident that we need to track the xmlns mapping separately, but I'm not sure if this is the best way to do it. However, people often say that the correct solution becomes obvious once you see an incorrect one, so I'm sharing it anyway.

**edit:**

Another approach I considered was giving each `Element` a reference the `WSDL` instance it was declared in. Implemented here: https://github.com/vpulim/node-soap/pull/1279

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -70,6 +70,7 @@ export class Element {
   public nsName?;
   public prefix?: string;
   public schemaXmlns?;
+  public effectiveXmlns?: IXmlNs;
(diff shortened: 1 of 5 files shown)


icholy:
Closing in favor of https://github.com/vpulim/node-soap/pull/1279

PR #1279, opened 2025-02-28
use wsdl specific xmlns prefix mappings

icholy:
This is an alternative to https://github.com/vpulim/node-soap/pull/1278 which is a fix for #1276 

Each `Element` is given a reference to the `xmlns` property of the `DefinitionsElement` it came from. The logic is completely the same except that we check in the element local `definitionsXmlns` before checking in the merged `definitions.xmlns`. A new `findNs` function was added to simplify the namespace lookups.

### DIFF
--- a/src/wsdl/elements.ts
+++ b/src/wsdl/elements.ts
@@ -70,6 +70,7 @@ export class Element {
   public nsName?;
   public prefix?: string;
   public schemaXmlns?;
+  public definitionsXmlns?: IXmlNs;
   public valueKey: string;
   public xmlKey;
   public xmlns?: IXmlNs;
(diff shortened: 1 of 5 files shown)


w666:
Hi @icholy, I had a quick look, looks good to me, thanks for the test.  All existing tests passed, so low risk that it can break something. 

When do you need next release? I will try to do it next week if it is not very urgent. Thanks.

icholy:
@w666 thanks for looking. I have a (hacky) work-around in my production code right now so it's not urgent. Next week is fine with me.

w666:
New release 1.1.9 has been published.

PR #1280, opened 2025-03-24
packages update

vision10:
updated axios, axios-ntlm and some other dev dependencies to latest
some dev deps seem to set minimum node version to 18, might need to update

modified tsconfig.json module to NodeNext
ts warning: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.

small fix for test: paths not correct on windows


w666 (approved):

w666:
Thanks for the updates @vision10

w666:
@vision10, I created #1281, probably it is time to bump min node version. feel free to help with that if you have time.

PR #1290, opened 2025-06-01
Removed the hardcoded ID of timestamp

rohangupta6807:
When we add additional references, we encounter a signature verification error because the `Timestamp` element has a hardcoded ID of `1`. The `signer.getSignatureXml()` method automatically assigns IDs sequentially (e.g., `0`, `1`, `2`) to the signed elements. As a result, the reference IDs become `0`, `1`, and `1`, causing a conflict due to the duplicate ID of `1` and leading to verification failure.

Removed the hardcoded timestamp ID and increases the package version.


smokhov:
@rohangupta6807 I think the unwrittien rule is that the official maintainers here bump the package version right before release when they are ready to release, in this case it's @w666 so the version change now is a bit premature to `1.2.0` here IMHO.

rohangupta6807:
@smokhov make sense, I have reverted the package version.

w666:
Hi, this makes sense. But could you add a test for this? I suppose you already know how to replicate the problem? Thanks.

rohangupta6807:
@w666  added the test case to verify the reference's Id.

w666 (approved):
Thanks for the test!

PR #1298, opened 2025-06-24
Axios update

Heymant7:
https://trello.com/c/b1o7CYu7/988-upgrade-axios-version-to-latest

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,8 @@
+0.45.0 / 2022-06-15
+===================
+* [FIX] axios peer dependency version (#1188)
+* [MAINTENANCE] Upgrade formidable to v2 (#1183)
+
 0.44.0 / 2022-06-06
 ===================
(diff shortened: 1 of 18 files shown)


smokhov:
Speaking as a user and an occasional contributor...

@Heymant7 this PR is a bit odd and covers more than Axios update...

- My suggestion would be first sync your fork from this repo via a rebase, and then focus only on the changes you want to contribute; this will avoid the conflicts you currently have
- This library's axios appears to be already a the updated version: https://github.com/vpulim/node-soap/blob/1754117bd4263efcbe6159a4976365098eb8d207/package.json#L11
- There are some custom body XML stuff as well; that is not discussed as to what and why
  - There are no tests covering that
- Please avoid bumping library versions and change author names and package names. The maintainer here @w666 would do that
- Your Trello board appears to be inaccessible as well.

w666:
Yeah, dodgy changes. Looks like a scam.

PR #1313, opened 2025-07-22
Handle deeply nested messages

### LINKED ISSUE #1284: Service can not identify operation / message
We have to provide both a client and a server based on an outside WSDL. We have created a node-soap client from the WSDL and validated it works against a mock generated by SoapUI. Then we created a node-soap server from the WSDL and tried to have the node-soap client talk to the node-soap server. That consistently fails in the xmlToObject of the WSDL: it looks like the server is unable to match the SOAP request to an operation / message in the WSDL. From the content of `this.definitions.messages` it looks like the server expects a `BerichtIndienenSoapIn` XML node instead of a `BerichtIndienen` XML node.
Client and server generated with SoapUI are consistent with the documentation of the WSDL and can talk to eachother. I'm not sure what exactly is going on, but since the node-soap client can not talk to the node-soap server I think this might be a node-soap issue.

I have attached a minimal reproduction. You can start the server using `npm start` and then run the testcase with `npm test` and it will throw the error 'TypeError: Cannot read properties of undefined (reading 'description')'. I have stripped out the majority of the definitions from the WSDL and XSD so they just contain a single operation and the needed types, but there might be some types left over in the XSD that are not referenced.

[reproduction.zip](https://github.com/user-attachments/files/19959830/reproduction.zip)

### LINKED ISSUE #1293: Result : Not showing Response Body
@smokhov 

After successful response , I am getting full stack of response including headers instead of only parsed JSON body in result

While printing error object I am getting

`TypeError: Cannot read properties of undefined (reading 'description')
    at SAXParser.p.onopentag (/var/task/node_modules/soap/lib/wsdl/index.js:261:37)
    at emit (/var/task/node_modules/sax/lib/sax.js:631:35)
    at emitNode (/var/task/node_modules/sax/lib/sax.js:636:5)
    at openTag (/var/task/node_modules/sax/lib/sax.js:832:5)
    at SAXParser.write (/var/task/node_modules/sax/lib/sax.js:1415:13)
    at WSDL.xmlToObject (/var/task/node_modules/soap/lib/wsdl/index.js:469:11)
    at parseSync (/var/task/node_modules/soap/lib/client.js:300:33)
    at /var/task/node_modules/soap/lib/client.js:506:24
    at handleBody (/var/task/node_modules/soap/lib/http.js:204:17)
    at /var/task/node_modules/soap/lib/http.js:236:24 {`

smokhov:
Closes #1205 
Closes #1293 
Closes #1284  

- make sure existing tests pass off PR #1205
- add new tests to cover #1205 and #1284 
- fix typos
- update a number of dependencies

### DIFF
--- a/test/client-test.js
+++ b/test/client-test.js
@@ -821,7 +821,8 @@ var fs = require('fs'),
               assert.ok(client.lastRequest);
               assert.ok(client.lastMessage);
               assert.ok(client.lastEndpoint);
-              console.log(client.lastMessage)
+              // Commented out for production test due to verbosity.
+              //console.log(client.lastMessage)
               assert.strictEqual(client.lastMessage, message);
               done();
             });
@@ -1824,7 +1825,8 @@ describe('Client posting complex body', () => {
         assert.ok(client.lastMessage);
         assert.ok(client.lastEndpoint);
 
-        console.log(client.lastMessage);
+        // Commented out for production test due to verbosity.
+        //console.log(client.lastMessage);
         const expectedBody = '<registrationMessages:registerUserRequest xmlns:registrationMessages="http://test-soap.com/api/registration/messages" xmlns="http://test-soap.com/api/registration/messages"><registrationMessages:id>ID00000000000000000000000000000000</registrationMessages:id><registrationMessages:lastName>Doe</registrationMessages:lastName><registrationMessages:firstName>John</registrationMessages:firstName><registrationMessages:dateOfBirth>1970-01-01</registrationMessages:dateOfBirth><registrationMessages:correspondenceLanguage>ENG</registrationMessages:correspondenceLanguage><registrationMessages:emailAddress>jdoe@doe.com</registrationMessages:emailAddress><registrationMessages:lookupPermission>ALLOWED</registrationMessages:lookupPermission><registrationMessages:companyAddress><ct:address xmlns:ct="http://test-soap.com/api/common/types"><ct:streetName>Street</ct:streetName><ct:postalCode>Code</ct:postalCode><ct:city>City</ct:city><ct:countryCode>US</ct:countryCode></ct:address><ct:companyName xmlns:ct="http://test-soap.com/api/common/types">ACME</ct:companyName></registrationMessages:companyAddress></registrationMessages:registerUserRequest>';
         assert.strictEqual(client.lastMessage, expectedBody);
 


smokhov:
Sigh... the timeouts seem to be related to GitHub runners' resources from my googling. Locally on my AlmaLinux 9.6 machine, all tests pass. So weird.

I have to run the tests like this locally in order for the ssh1-rsa test to work:

```
OPENSSL_ENABLE_SHA1_SIGNATURES=1 npm run cover
```

else it always fails.

w666:
I don't think this is related to node version. But I am getting timeouts in my environment too (WSL - Ubuntu 22.04 LTS, Node.js 22).

I will try to have a look.

smokhov:
I've found the reason. Will commit shortly.

w666:
@smokhov 

I released urgent fix, it conflicts with your branch now. Sorry about that.

smokhov:
@w666 TL;DR: after experimenting back and forth and adding some instrumentation it seems `xmlns.soap` was not enough -- the tests also use `xmlns.soapenv` and `xmlns.S` occasionally. Also some of the `xmlns` was not yet populated before that `if()` was made. In the process of doing all that, I've ended up updating the deps and node version to be min supported LTS version -- all these can be reverted. Now onto the issue on how to add a test for that change and what it should have...

w666:
I think there should be an example request that triggers that original problem mentioned in the #1205. 
Ideally test should fail on non-fixed version (like latest node-soap release) and should pass against this branch.

w666:
And I think I will ask to revert min node version related changes, but we can discuss it later.

w666:
I suppose that https://github.com/vpulim/node-soap/issues/1284 can be related, there is code example.

smokhov:
> I suppose that #1284 can be related, there is code example.

You are right, it may indeed be related! It does look relevant! I've added the reproduction content into the request-response samples al the wsdl and schemas, but having trouble getting the request working as it's giving me an error diff expecting something else vs. what's given in the SoapUI's samples in that issue. I'll commit a non-working test from that issue for further debugging for now so it doesn't sit just with me. Any comment is welcome @w666 as I don't have complete understanding of it yet.

w666:
Thanks, I pulled the latest changes, maybe will have time this week.

test/request-response-samples/IndienenBericht__should_work_with_deeply_nested_XML_messages/request.xml
```diff
@@ -0,0 +1,43 @@
...
+            <v32:BerichtInfo>
+               <!--You may enter ANY elements at this point-->
+            </v32:BerichtInfo>
+            <v32:Bericht/>
+         </v3:Request>
+      </v3:IndienenBericht>
+   </soap:Body>
+</soap:Envelope>
+
+<?xml version="1.0" encoding="utf-8"?>
```
smokhov, line 27: From this line downwards is a comparison text. It should not be in the code; committed by accident before removing. Please remove this fragment.

w666:
@smokhov 

I think I fixed the test. Also fixed that annoying issue in test runner, it does not throw timeout error anymore

.github/workflows/pr.yml
```diff
@@ -3,6 +3,7 @@ on: pull_request
 jobs:
   code-quality:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
```
w666, line 6: I think this can be something like 5 mins.

package.json
```diff
@@ -35,7 +35,7 @@
     "watch": "tsc -w -p .",
     "lint": "tslint -p tsconfig.json",
     "toc": "./node_modules/.bin/doctoc Readme.md --github --maxlevel 3",
-    "cover": "nyc --extension=.ts --reporter=lcov --reporter=html --reporter=text mocha --timeout 15000 --exit test/*-test.js test/security/*.js",
+    "cover": "OPENSSL_ENABLE_SHA1_SIGNATURES=1 nyc --extension=.ts --reporter=lcov --reporter=html --reporter=text mocha --no-parallel --timeout 30000 --exit test/*-test.js test/security/*.js",
```
w666, line 38: Would be good to double check if it works without this change.

src/wsdl/index.ts
```diff
@@ -232,7 +232,56 @@ export class WSDL {
...
+      // Debug instrumentation
+      console.log("==============================XMLNS/=================================");
+      console.log("top.name=" + top.name);
+      console.log("name=" + name);
+      console.log("objectName=" + objectName);
+      console.log("xmlns=");
+      console.log(xmlns);
+      console.log("xmlns.soap=" + xmlns.soap);
+      console.log("==============================/XMLNS=================================");
+      */
```
w666, line 282: I think this is not needed anymore.

test/client-test.js
```diff
@@ -1824,7 +1824,7 @@ describe('Client posting complex body', () => {
         assert.ok(client.lastMessage);
         assert.ok(client.lastEndpoint);
 
-        console.log(client.lastMessage);
+        //console.log(client.lastMessage);
```
w666, line 1827: this and one above are unrelated changes, should be reverted.

package.json
smokhov: This is not necessary for Ubuntu 20 or 22 I believe on GitHub or WSL. But on a distro that deprecated SHA1 and issues a warning, it failes sha1-rsa test without that env var. Without it I cannot run `npm run cover` without errors on master. I believe GitHub is switching to Ubuntu 24 soon for Actions, and it may exhibit the same error in that one test.

test/client-test.js
w666: right, maybe add a comment explaining that it was commented because it is very verbose?

package.json
w666: okay, would be good to raise a ticket to investigate that later.

smokhov: Put it here: #1323

w666:
Merged. Tried to clean up squashed commit message, hopefully I captured all the changes.

Thank you @smokhov for leading this.

PR #1324, opened 2025-08-04
Add `*.ts` to editorconfig

nora-heydecker-axa:
Most files in this project are now TypeScript, so the indentation configuration in the `.editorconfig` is important for that too.

My editor (WebStorm) wasn't correctly picking it up before and defaulting to 4 spaces, which made editing annoying.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,10 @@ insert_final_newline = false
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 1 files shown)


w666:
There can be other projects where code style does not match your local or default editor config. In such cases I just save a file without reformat.
But I am happy to approve this if it makes your life easier.

PR #1337, opened 2025-08-18
Add support for multi-service and multi-port binding WSDL files

smokhov:
- Closes #1226 

This PR supersedes PR #1226 (by @paulish). For now simply sync'ing it with `master`. Unsure yet what type of test it would have... It seems covered? But I don't have a concrete example in mind.

### DIFF
--- a/test/server-options-test.js
+++ b/test/server-options-test.js
@@ -429,7 +429,6 @@ describe('SOAP Server with Options', function () {
         })
         .catch((err) => {
           try {
-            console.log(`>>> data`, err.response.data);
             assert.equal(err.response.status, 500);
             assert.ok(err.response.data.match(/<faultcode>.*<\/faultcode>/g), 500);
             assert.ok(err.response.data.match(/<faultstring>.*<\/faultstring>/g), 'Invalid XML');


w666:
Will have a look in this one later. It is in my backlog.

w666:
> Will have a look in this one later. It is in my backlog.

fyi: I am working on a test for this one when have time, but due to other issues in this project this is not a first priority.

w666:
Was looking into this past week. Still not sure I understand the problem because I managed to create a WSDL file that breaks it. 
Basically, when there are two services and port bindings are in order of appearance it works, but once I change, for example, port order it breaks parsing. I assume this is what this PR was about. But it does not work, moreover it works same way before the fix.

Anyway, I continue investigating this. At least I have a good WSDL and a template for a test.

w666:
Reverted original change, I think it is not correct, moreover it does not fix anything. Instead fixed looked through service->port bindings. Added new multi-service/multi-port WSDL and new tests.

@smokhov would you be able to review this, in case I overlooked something?

smokhov:
> Reverted original change, I think it is not correct, moreover it does not fix anything. Instead fixed looked through service->port bindings. Added new multi-service/multi-port WSDL and new tests.
> 
> @smokhov would you be able to review this, in case I overlooked something?

I had a look at the changes, you've done quite a comprehensive test coverage for the issue, so I think it's good to go. I guess the code has improved since the times of the original PR so the original change became redundant, but it's nice now to have it fully covered as well.

PR #1338, opened 2025-08-18
 enabling async methods for postProcess (take 2)

smokhov:
- Closes #1174 

This PR supersedes PR#1174 (by @karlop). I've sync's master onto it and checked its tests pass.


test/client-test.js
```diff
@@ -1834,6 +1834,93 @@ describe('Client posting complex body', () => {
...
+            city: 'City',
+            countryCode: 'US'
+          },
+          companyName: 'ACME'
+        }
+      }
+
+      client.registerUser(requestBody, {
+        postProcess: async (xml) => {
+          const testString = await new Promise(resolve => setTimeout(resolve("Jane"), 3000));
```
w666, line 1867: This resolves immediately, so does not really simulate an async operation. 
I think it should be something like this 
```suggestion
          const testString = await new Promise(resolve => setTimeout(() => resolve("Jane"), 50));
```

What do you think?

w666:
Would be good to mention this in readme as well.

PR #1345, opened 2025-09-01
[dependabot] tune deps scheduling and updates

smokhov:
This adds ignore directives and introduces a cooldown period for updates to occur.


w666:
Thanks for this one!

PR #1347, opened 2025-09-07
fix harcoded namespace prefixes in parsing & add more tests

### LINKED ISSUE #1346: Possible regression in 1.3.0 from 1.2.1
We have a WSDL file that starting returning no data starting from upgrading from 1.2.1 to 1.3.0

After comparing the changes https://github.com/vpulim/node-soap/compare/v1.2.1...v1.3.0#files_bucket and checking with gpt5

The response was:

> Short answer: because a namespace / WSDL-parsing change in v1.3.0 broke how element names are matched — the refactor that landed in the v1.3.0 changeset (handling redefined namespace aliases and deeply-nested messages) changed the order/logic of how src/wsdl/index.ts discovers element names, xsi:type / xmlns mappings and id/href refs. That can make the parser fail to match the WSDL message/element names your WSDL uses (so the client ends up with no parsed response data). See the commits that introduced this behavior. 
> 
> Why that breaks your WSDL (short technical explanation)
> 
> The v1.3.0 changes moved and rewrote the code that: (a) collects xmlns declarations on an element, (b) handles id/href multiRef refs, and (c) decides when an element inside the SOAP Body should be treated as the response object name.
> 
> If the parser no longer resolves the exact local name/namespace the WSDL used to register in definitions.messages[...], the code that maps an incoming XML element to its WSDL type/response drops through and you get no parsed data (or empty client.describe() / empty result object).
> 
> The commit specifically mentions supporting “redefining of a namespace alias” — that change can change the computed name/objectName used for lookups. If your WSDL uses redefined/embedded namespace aliases (many SAP WSDLs do), this is the likely root cause. 
> [GitHub](https://github.com/vpulim/node-soap/commit/4343a09130644704c950f5f6d131e195cdfa7ab1)
> 
> How you can verify / debug quickly
> 
> Create a client and dump the WSDL messages to see whether operations/messages were parsed:
> 
> const soap = require('soap');
> soap.createClient(wsdlUrl, {}, (err, client) => {
>   if (err) return console.error(err);
>   console.log('describe():', client.describe());
>   // low-level: inspect parsed WSDL
>   console.log('wsdl messages keys:', Object.keys(client.wsdl.definitions.messages));
> });
> 
> 
> If client.describe() is empty or messages don’t look right, the WSDL parsing failed to create the mapping.
> 
> Capture raw response and compare lastResponse/lastRequest to see whether the XML body element name is different from keys under client.wsdl.definitions.messages.
> 
> What you can do right now (workarounds / fixes)
> 
> Pin / downgrade to the last known-good version: npm install soap@1.2.1. If that restores behaviour, it confirms the regression is in the 1.3.0 changes. (fastest mitigation)
> 
> Reproduce and open an issue on the repo with a minimal WSDL + sample request/response that demonstrates the failure — the maintainers already touched this area so they’ll likely accept a regression report. Link to the commits above when you file it. 
> 
> Temporary local patch: if you can edit node_modules/soap/src/wsdl/index.ts, revert the specific parsing logic change around element xmlns and the early objectName decision back to the old behavior (the diff is in the v1.3.0 commit). I can draft a small patch for you if you want — tell me whether you prefer a patch file or a one-off monkey patch for runtime.

smokhov:
- Closes #1346 

This PR simply creates two new request-response sample regression tests based on the WSDL from #1346. I test with various options and with and without security. The tests pass with the current branch tip. However, the OP would need to confirm if the samples are representative enough of the data they see.

@w666 -- over to you to make any adjutments as needed

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -22,6 +22,7 @@ import * as elements from './elements';
 const debug = debugBuilder('node-soap');
 
 const XSI_URI = 'http://www.w3.org/2001/XMLSchema-instance';
+const ENV_URI = 'http://schemas.xmlsoap.org/soap/envelope/';
 
 export function trim(text) {
   return text.trim();
@@ -215,6 +216,7 @@ export class WSDL {
     };
     const stack: any[] = [{ name: null, object: root, schema: schema }];
     const xsiPrefixes: Map<any, any> = new Map();
+    let envPrefix: string = "soap";
     const xmlns: any = {};
 
     const refs = {};
@@ -255,6 +257,9 @@ export class WSDL {
           if (value === XSI_URI) {
               xsiPrefixes.set(name, value);
           }
+          if (value === ENV_URI) {
+              envPrefix = name;
+          }
           continue;
         }
         hasNonXmlnsAttribute = true;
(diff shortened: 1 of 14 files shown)


w666:
Tests look good, I just think we should wait for the results from https://github.com/vpulim/node-soap/issues/1346.

Issue #1267, opened 2024-12-22
sequence is not respected


toughlex-donatas:
https://www.w3schools.com/xml/el_sequence.asp

> The sequence element specifies that the child elements must appear in a sequence.

https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms256089(v=vs.100)

> Requires the elements in the group to appear in the specified sequence within the containing element.

This library does not ensure this behavior.

Probably needed changes are somewhere here https://github.com/vpulim/node-soap/blob/15032f0484d606463684919c7aa1a8f9f77402bc/src/wsdl/index.ts#L748

w666:
Hi @toughlex-donatas,

Could you please a bit more details? How did you find this issue? Is there an example that you can share, like code snippet or xml file?

Thanks.

toughlex-donatas:
To replicate the problem you could create a test in https://github.com/vpulim/node-soap/blob/master/test/wsdl-test.js and use this wsdl https://github.com/vpulim/node-soap/blob/master/test/wsdl/marketo.wsdl
Try to call GetLeadChanges for example, you will see that SOAP XML ParamsGetLeadChanges will not have its children ordered in correct order.
There are APIs that don't work with unordered sequence elements.

w666:
Had a look on it, does not look like an easy fix.  Will have another look later when have more time.

kouak:
Here's a helper I've made for this purpose.

This function is a utility that recursively reconstructs a JavaScript object to match the exact property order defined in a schema. It also acts as a sanitizer by filtering out any properties from the input object that are not present in the schema.

```typescript
export function reorderKeys<T extends Schema, O extends { [key: string]: any }>(
  schema: T,
): (obj: O) => O {
  return (obj: O): O => {
    // Loop through schema, pull property from Object
    return Object.keys(schema).reduce<any>((prev, curr) => {
      const lookupKey: string = curr.replace(/\[\]$/, '');
      const isArrayExpected: boolean = curr.endsWith('[]');

      if (!(lookupKey in obj)) {
        return prev;
      }

      const currSchema = schema[curr];

      if (typeof currSchema === 'string') {
        prev[lookupKey] = obj[lookupKey];
        return prev;
      }

      if (typeof currSchema === 'object') {
        if (
          Object.keys(currSchema).filter(
            (k) => k !== 'targetNSAlias' && k !== 'targetNamespace',
          ).length
        ) {
          prev[lookupKey] =
[...]

PR #1348, opened 2025-09-10
Add support for the group element

rjordakiev:
Working on a change that allows respecting the order imposed by `xs:sequence` in schema, which is an existing issue (#1267). 

Respecting the `xs:group` element is also of importance since there are specific cases when the sequence appears as a child of the group.


rjordakiev:
Is there any intrinsic reason that the group element is commented out and parsed as `UnexpectedElement` that I may be messing or it simply is obsolete for the current implementation?

w666:
Could you please add some tests for this? it would help to understand the changes and lock down the implementation. Thanks.

smokhov:
> Is there any intrinsic reason that the group element is commented out and parsed as `UnexpectedElement` that I may be messing or it simply is obsolete for the current implementation?

IMO, it's history the OG devs would know. Whatever was the reason, we don't need to dwell on it, and push the project forward, so thank you for your contribution.

I see in your fork you also did commit an implementation of the sequence ordering; that other branch has tests. Do they cover this group element as well? if so, then why not include it into the PR as well? It seems from your implementation `group` element is required to enable proper sequence ordering anyway.

w666:
@rjordakiev

If you can't add a test for this change, I can do this, but I need example WSDL and ideally expected request response. Thanks.

smokhov:
> @rjordakiev
> 
> If you can't add a test for this change, I can do this, but I need example WSDL and ideally expected request response. Thanks.

They have a some test WSDL and a test for the squence ordering in their other branch, maybe can use that?
  - https://github.com/rjordakiev/node-soap/commit/c346fa20f3fd3684dc74632427ceddf30cfe6960

w666:
@smokhov 

Ah, I will have a look a bit later, thanks for checking this.

smokhov:
> > @rjordakiev
> > If you can't add a test for this change, I can do this, but I need example WSDL and ideally expected request response. Thanks.
> 
> They have a some test WSDL and a test for the squence ordering in their other branch, maybe can use that?
> 
>     * [rjordakiev@c346fa2](https://github.com/rjordakiev/node-soap/commit/c346fa20f3fd3684dc74632427ceddf30cfe6960)

@w666 pls rebase and merge when you have time. Then we can merge the rest

w666 (approved):

w666:
@smokhov done

PR #1365, opened 2025-10-04
Handle different namespace prefix for the same namespace

### LINKED ISSUE #1363: Incorrect parsing of complex types
I have been using the package for quite some time now with a fairly extensive API and had never had any problems until last week when there was a minor update to the service definition. Since then, boolean values in responses with complex types are no longer parsed correctly and remain strings (e.g., getMessageResponse).

I worked on this problem all weekend, but neither overwriting namespaces nor using a custom deserializer solved it, and crying didn't help either.

I'm absolutely stuck and it's driving me crazy. Does anyone have any hint for me?

[wsdl.zip](https://github.com/user-attachments/files/22583286/wsdl.zip)

w666:
- Add `forceUseSchemaXmlns` option to force to use schema xmlns, it contains all the prefixes defined in XSD files.
 - Add tests for both cases when `forceUseSchemaXmln=true` and not set.

 Close #1363


w666:
There is a very old change https://github.com/vpulim/node-soap/pull/923 that I think is not correct, but it is there for very long time so very likely some people may rely on this behavior. New option fixes that, but must be explicitly set.

PR #1378, opened 2025-11-01
Fix broken TOC links for italic headings

tzkz:
Fixed anchor links in table of contents for headings that use underscores for italic formatting (_request_, _message_, etc.). GitHub strips the underscores when generating section anchors, so the links need to reference the plain text versions.

See: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links


smokhov:
@tzkz thanks for the fix up. Have you checked it works with `npm run toc`? This is how that TOC is maintained. I am afraid your changes might get undone the next time some one runs that...

https://github.com/vpulim/node-soap/blob/91a3898f18e2636ef75c39b1ea7994c1f321ef54/package.json#L37

tzkz:
@smokhov thanks for reviewing. Yes, you're right, running `npm run toc` will undo my changes. Actually, the issue i'm fixing was already fixed in the `doctoc` https://github.com/thlorenz/doctoc/pull/235 but was never released.

Also, running `npm run toc` will undo the changes made by others in other sections, and will create broken links, and these are different issues that weren't fixed in `doctoc`, for example:
1. `- [Client._method_(args, callback, options) - call _method_ on the SOAP service.](#client_method_args-callback-options---call-_method_-on-the-soap-service)` - here, the problems is `doctoc`'s dependency `anchor-markdown-header` removes punctuation `.` before stripping markdown, `Client._method_` becomes `Client_method_`
2. `- [Client._service_._port_._method_(args, callback[, options[, extraHeaders]]) - call a _method_ using a specific _service_ and _port_](#client_service__port__method_args-callback-options-extraheaders---call-a-_method_-using-a-specific-_service_-and-_port_)` - here, the problem is `anchor-markdown-header`s dependency `remove-markdown` does not handle cases like `Client._service_._port_._method_` properly, where multiple italic words are separated by periods, becomes `Client.service._port_.method`

I can try fixing the dependencies but I'm not sure if they will be merged/released. They don't seem to be so actively maintained.

Or we should not use _italic_ in headings 🤷‍♂️

smokhov:
I personally agree about de-italicizing headings. It is indeed unclear when doctoc will release next as well as other deps. If @w666 agrees, can perhaps then strip italics from the markdown headings altogether.

On the other hand, the project has recently switched to `eslint` and `prettier`. Both support markdown. Maybe in that `toc` script, can try formatting `Readme.md` with `prettier` after the current `doctoc` run and see if this automatically fixes it as-is?

w666:
Hi all,

I am okay with de-italicizing them. I would remove that doctoc tbh. Feel free to do so, and as @smokhov suggested replace it with prettier. 

But I am also okay to merge it as it is.

w666 (approved):

smokhov:
> Hi all,
> 
> I am okay with de-italicizing them. I would remove that doctoc tbh. Feel free to do so, and as @smokhov suggested replace it with prettier.
> 
> But I am also okay to merge it as it is.

What I meant actually:

```json
    "toc": "./node_modules/.bin/doctoc Readme.md --github --maxlevel 3 && prettier ... Readme.md", 
```

This is because maintaining TOC accurately manually will be a PITA

w666:
@smokhov 

I will just merge it and create a separate PR with the suggested change.

w666:
actually, suggested changes does not work, it still breaks some links. And seems like there are no good alternatives. I guess I will park this issue for now.

PR #1387, opened 2025-11-18
Replace depricated url.parse with the WHATWG URL API

### LINKED ISSUE #1380: URL.PARSE() uses deprecation, security warning message in Node.js v24
Now I update my node version from 22 to 24, and when i try to use the soap feature on a failed test, this works, catching the error, but I receive a warning message in my shell:

```bash
(node:30416) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
```
Node.js v24.11.0

w666:
close #1380

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -8,7 +8,6 @@ import { NtlmClient } from 'axios-ntlm';
 import { randomUUID } from 'crypto';
 import debugBuilder from 'debug';
 import { ReadStream } from 'fs';
-import * as url from 'url';
 import { MIMEType } from 'whatwg-mimetype';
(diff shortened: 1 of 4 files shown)


smokhov:
How much do you really trust copilot to do the right thing?

w666:
Ah, it is kinda useless, was just experimenting with it past week. Also was wondering if it helps with this PR, as I am still not sure if I should release this.

PR #1398, opened 2025-12-02
Bump glob (via mocha) to fix glob CLI: Command injection via -c/--cmd executes matches with shell:true

smokhov:
- closes #1391
-  `glob` as a transitive dep to `mocha` had vulns that were fixed and mocha also bumped as a result. Perhaps it's time to release 1.6.1, @w666 ?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3429,15 +3429,15 @@
       }
     },
     "node_modules/glob": {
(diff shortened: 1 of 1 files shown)


w666:
I will try to do a maintenance release today.

PR #1413, opened 2025-12-28
fix: align deserialization of restricted strings with standard strings

### LINKED ISSUE #1412: Empty string elements with XSD restrictions are deserialized as `null` instead of `""`
I've encountered an inconsistency in how empty XML elements are deserialized when XSD restrictions are involved.

When an element is defined as a `simpleType` based on `xs:string` with a restriction (such as <xs:pattern>), receiving an empty tag in the SOAP response (`<myField></myField>`) results in the value being deserialized as `null`.

The value should be deserialized as an empty string "", which is the behavior observed for fields defined without restrictions.

kouak:
This PR aligns the deserialization behavior of restricted strings (e.g., pattern, length) with standard strings.

I updated the logic to normalize the schema name before checking the type. Now, restricted strings are correctly identified and behave exactly like standard strings.

Added a test.

Closes #1412

### DIFF
--- a/src/wsdl/index.ts
+++ b/src/wsdl/index.ts
@@ -367,7 +367,12 @@ export class WSDL {
       const topSchema = top.schema;
       const name = splitQName(nsName).name;
 
-      if (typeof cur.schema === 'string' && (cur.schema === 'string' || cur.schema.split(':')[1] === 'string')) {
+      /**
+       * When parsing a string element, we need to correctly transform `<tag></tag>`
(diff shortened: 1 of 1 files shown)


w666:
Hi @kouak,

Thank you for your contribution, especially for the tests.
Is this fix urgent for you. If yes, then I will try to allocate time tomorrow to build a new release.

kouak:
It can wait :)

PR #1418, opened 2026-01-01
Remove OPENSSL_ENABLE_SHA1_SIGNATURES=1 as it is not required anymore

### LINKED ISSUE #1323: investigate the need of OPENSSL_ENABLE_SHA1_SIGNATURES=1 in npm cover and npm test
On some distributions (e.g., RH-derivative EL9, such as AlmaLinux 9), SHA1 is deprecated by default and running e.g. `npm run cover` (and presumable `npm run test`) fails with:

```
  1) WSSecurityCert
       should use rsa-sha1 signature method when the signatureAlgorithm option is set to WSSecurityCert:
     Error: error:03000098:digital envelope routines::invalid digest
      at Sign.sign (node:internal/crypto/sig:128:29)
      at node-soap/node_modules/xml-crypto/lib/signature-algorithms.js:11:32
      at RsaSha1.getSignature (node_modules/xml-crypto/lib/types.js:52:20)
      at SignedXml.calculateSignatureValue (node_modules/xml-crypto/lib/signed-xml.js:327:42)
      at SignedXml.computeSignature (node_modules/xml-crypto/lib/signed-xml.js:763:18)
      at WSSecurityCert.postProcess (lib/security/WSSecurityCert.js:2:10622)
      at Context.<anonymous> (test/security/WSSecurityCert.js:285:26)
      at process.processImmediate (node:internal/timers:483:21)
```

running it as `OPENSSL_ENABLE_SHA1_SIGNATURES=1 npm run cover` makes it pass. Going to the future either both targets in `project.json` should have this option, esp. when GitHub Actions switches to Ubuntu 24 soon(tm), or the test should migrate to a newer algorithm.

w666:
Close #1323

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "toc": "./node_modules/.bin/doctoc Readme.md --github --maxlevel 3",
-    "cover": "OPENSSL_ENABLE_SHA1_SIGNATURES=1 TZ=UTC nyc --extension=.ts --reporter=lcov --reporter=html --reporter=text --reporter=json-summary mocha --no-parallel --timeout 30000 --exit test/*-test.js test/security/*.js",
+    "cover": "nyc --extension=.ts --reporter=lcov --reporter=html --reporter=text mocha --no-parallel --timeout 30000 --exit test/*-test.js test/security/*.js",
     "postcover": "node ./scripts/update-coverage-badge.js",
     "cover:badge": "npm run cover && npm run postcover",
     "docs": "typedoc --out docs",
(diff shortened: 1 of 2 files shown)


smokhov:
JFOI: it (the SSL test) still breaks without it on AlmaLinux 9, but I can workaround it by setting the var myself before running `npm cover`

w666:
Ah, okay. It is fine with me to keep it. 

I just investigated a bit and ubuntu 24 has this disabled as well and github actions has been migrated to ubuntu 24, so I assumed it should work in other distros too.

smokhov:
> Ah, okay. It is fine with me to keep it.
> 
> I just investigated a bit and ubuntu 24 has this disabled as well and github actions has been migrated to ubuntu 24, so I assumed it should work in other distros too.

I think it's still OK to proceed with the removal as you indicated.

PR #1423, opened 2026-01-10
Bump qs to 6.14.1 and body-parser to 2.2.2

smokhov:
- body-parser (and express, but that will be released later) have a transitive dep on qs
- qs 6.14.0 was found vulnerable to qs's `arrayLimit` bypass in its bracket notation allowing DoS via memory exhaustion (CVE-2025-15284)

@w666 -- please review and merge and I guess 1.6.3 is in order?


w666:
1.6.3 has been released.

PR #1431, opened 2026-01-31
Add test for datetime conversion from incomplete to full ISO date string

w666:
for example `2026-03-15T09:00:00` to `2026-03-14T20:00:00.000Z`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "cover": "OPENSSL_ENABLE_SHA1_SIGNATURES=1 nyc --extension=.ts --reporter=lcov --reporter=html --reporter=text mocha --no-parallel --timeout 30000 --exit test/*-test.js test/security/*.js",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js -v",
     "docs": "typedoc --out docs",
-    "test": "mocha --timeout 15000 --bail --exit test/*-test.js test/security/*.js"
+    "test": "TZ=UTC mocha --timeout 15000 --bail --exit test/*-test.js test/security/*.js"
   },
   "keywords": [
     "soap"
--- a/test/request-response-samples/Dummy__should_handle_datetime_transformation/request.json
+++ b/test/request-response-samples/Dummy__should_handle_datetime_transformation/request.json
@@ -1,0 +1,5 @@
+{
+  "EventName": "Product Launch",
+  "EventDateTime": "2026-02-15T14:30:00.000Z",
+  "CreatedAt": "2026-01-29T10:15:30.500Z"
+}
--- a/test/request-response-samples/Dummy__should_handle_datetime_transformation/request.xml
+++ b/test/request-response-samples/Dummy__should_handle_datetime_transformation/request.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:tns="http://www.example.com/datetime">
+    <soap:Body>
+        <DummyRequest xmlns="http://www.example.com/datetime">
+            <EventName>Product Launch</EventName>
+            <EventDateTime>2026-02-15T14:30:00.000Z</EventDateTime>
+            <CreatedAt>2026-01-29T10:15:30.500Z</CreatedAt>
+        </DummyRequest>
+    </soap:Body>
+</soap:Envelope>
--- a/test/request-response-samples/Dummy__should_handle_datetime_transformation/response.json
+++ b/test/request-response-samples/Dummy__should_handle_datetime_transformation/response.json
@@ -1,0 +1,5 @@
+{
+  "EventId": "EVT-2026-001",
+  "ProcessedDateTime": "2026-01-29T12:00:45.000Z",
+  "NextScheduledDateTime": "2026-03-15T09:00:00.000Z"
+}
--- a/test/request-response-samples/Dummy__should_handle_datetime_transformation/response.xml
+++ b/test/request-response-samples/Dummy__should_handle_datetime_transformation/response.xml
@@ -1,0 +1,12 @@
+<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/"
+    xmlns:tns="http://www.example.com/datetime"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <soap:Header></soap:Header>
+    <soap:Body>
+        <tns:DummyResponse>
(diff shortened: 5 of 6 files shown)


jmechevarria:
Here's my feedback after analysing the tests.

I checked out tag `v1.1.1`, and ran test file `request-response-samples-test.js` using the tests you added in this PR.

The fact that the test passes **is** the actual issue here.

**response.xml**:
`<tns:NextScheduledDateTime>2026-03-15T09:00:00</tns:NextScheduledDateTime>`
**response.json**:
`"NextScheduledDateTime": "2026-03-15T09:00:00.000Z"`

That test should not pass ever, because then it means that the library is doing exactly what I mentioned in the original post, that the original XML date strings are being transformed (`2026-03-15T09:00:00` -> `2026-03-15T09:00:00.000Z`) instead of mapped as-is (`2026-03-15T09:00:00` -> `2026-03-15T09:00:00`).

The test I would expect to see pass, is one that uses:

**response.xml**:
`<tns:NextScheduledDateTime>2026-03-15T09:00:00</tns:NextScheduledDateTime>`
**response.json**:
`"NextScheduledDateTime": "2026-03-15T09:00:00"`

------

Something I noticed is that the behavior with 1.1.0 is the same, which honestly is a mystery to me, because I even went back to my project code and did multiple clean installs of 1.1.0 and 1.1.1, testing each time, and each time obtaining a different result: 1.1.0 keeps the string as is, and 1.1.1 appends the `.000Z`.

------

I **think** the cause is this line:
[value = new Date(text);](https://github.com/vpulim/node-soap/blob/7a620c31caaeee9835040d9cd63c78f2bf735ad7/src/wsdl/index.ts#L469)

It turns the date string into an object, and of course if that's used with `toISOString` for example, it will generate a string with `.000Z` at the end. This is definitely not something I would expect as a user of the library, because it's taking a date string with no timezone information, and it's assuming it's always UTC by appending `.000Z` at the end.

I hope I explained myself clearly.

w666:
@jmechevarria 

How does your WSDL look like? Is it possible to get at least a snippet with the datatype? 
It must be a different problem, because the test I created works same way in all versions.

jmechevarria:
> @jmechevarria
> 
> How does your WSDL look like? Is it possible to get at least a snippet with the datatype? It must be a different problem, because the test I created works same way in all versions.

Sorry for the delay.

Here's an excerpt:

```
    <xs:complexType name="Object1">
        <xs:complexContent mixed="false">
            <xs:extension base="q25:AMessage" xmlns:q25="http://schemas.example.com/WebServices/DataContracts/Common">
                <xs:sequence>
                    <xs:element minOccurs="0" name="Prop1" nillable="true" type="xs:string"/>
                    <xs:element minOccurs="0" name="Prop2" nillable="true" type="xs:string"/>
                    <xs:element minOccurs="0" name="STA" type="xs:dateTime"/>
                    <xs:element minOccurs="0" name="STD" type="xs:dateTime"/>
                    <xs:element minOccurs="0" name="Prop3" type="xs:boolean"/>
                    <xs:element minOccurs="0" name="Prop4" nillable="true" type="xs:string"/>
                </xs:sequence>
            </xs:extension>
        </xs:complexContent>
    </xs:complexType>
```

w666:
Hi @jmechevarria,

Thanks for additional details, but unfortunately, I can't replicate this.  Also, I don't see what changes between 1.1.0 and 1.1.1 can cause this behaviour. 

I think I need your help with to create reproduction steps or maybe update the test I created in this MR so it can replicate the behaviour you described in the issue.

Thanks.

smokhov:
There are many occurrences of `toISOString()` in `node_modules` in various deps. Also `JSON.stringify()` calls `toISOString()`. One way to avoid it is to add an `wsdl_option`, e.g., `raw_date: true`, so that the `value` here: https://github.com/vpulim/node-soap/blob/7a620c31caaeee9835040d9cd63c78f2bf735ad7/src/wsdl/index.ts#L469 is treated as text instead of `Date()` since `Date()`'s internal storage is always UTC.

w666:
@smokhov,

Don't you think it would be more like a workaround to ignore datetime data type?
I am happy to provide such workaround if we really broke it at some previous version, the only problem we can't replicate the issue. Would be good if @jmechevarria can help us with this.

smokhov:
> @smokhov,
> 
> Don't you think it would be more like a workaround to ignore datetime data type? I am happy to provide such workaround if we really broke it at some previous version, the only problem we can't replicate the issue. Would be good if @jmechevarria can help us with this.

You are correct. It is a workaround. Like you, I can't explain [the differences between 1.1.0 and 1.1.1](https://github.com/vpulim/node-soap/compare/v1.1.0...v1.1.1) as there are no related changes there nor dependency updates.

I also did baby tests with the following output:

```javascript
const date3 = new Date("2026-03-15T09:00:00");
console.log(date3.toString());
console.log(date3.toISOString());
console.log("stringify: " + JSON.stringify(date3));

const date4 = new Date("2026-07-04T15:30:00-04:00");
console.log(date4.toString());
console.log(date4.toISOString());

const date5 = new Date("2026-07-04T19:30:00.000Z");
console.log(date5.toString());
console.log(date5.toISOString());

const date6 = new Date("2026-01-23T16:13:49.7494192");
console.log(date6);
console.log(date6.toString());
console.log(date6.toUTCString());
console.log(date6.toISOString());

...

"Sun Mar 15 2026 09:00:00 GMT-0400 (Eastern Daylight Time)"
"2026-03-15T13:00:00.000Z"
'stringify: "2026-03-15T13:00:00.000Z"'
"Sat Jul 04 2026 15:30:00 GMT-0400 (Eastern Daylight Time)"
"2026-07-04T19:30:00.000Z"
"Sat Jul 04 2026 15:30:00 GMT-0400 (Eastern Daylight Time)"
"2026-07-04T19:30:00.000Z"
Fri Jan 23 2026 16:13:49 GMT-0500 (Eastern Standard Time)
"Fri Jan 23 2026 16:13:49 GMT-0500 (Eastern Standard Time)"
"Fri, 23 Jan 2026 21:13:49 GMT"
"2026-01-23T21:13:49.749Z"
```

However, we are not unique with this issue. Apparently strong-soap had similar problems (as well as a reverse problem), and they had a patch for this along the same lines with an option for a workaround:

- https://github.com/loopbackio/strong-soap/issues/289
- https://github.com/loopbackio/strong-soap/pull/525
- https://github.com/loopbackio/strong-soap/issues/94

w666:
@smokhov 

Okay, if this is a real issue for so many people lets just add an option to return date in this weird format. 
I will try to allocate time for this, unless you have a particular interest in this and want to pick this up.

jmechevarria:
Sorry for the late reply.

We decided to go with the configuration option mentioned [here](https://github.com/vpulim/node-soap?tab=readme-ov-file#custom-deserializer), and it has allowed us to continue using the latest version of this library. I'll also take a look into `raw_date: true` since it sounds more straightforward.

Thanks for looking into this

w666:
Okay, I will just merge this test then and close the issue.

PR #1435, opened 2026-02-06
Handle undefined toXml()

### LINKED ISSUE #1422: Error with  the inclusion of validations of:  (decodedHeaders || (this.security && this.security.toXML()) in client.ts
Hi, currently I have a problem with validation added in the commit:
Commit 05eb9bd

https://github.com/vpulim/node-soap/commit/05eb9bd34cf8bf0e3150d81b9f35948a1e9dacab#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R446:~:text=(decodedHeaders%20%7C%7C,())

In that commit was added : ` (decodedHeaders || (this.security && this.security.toXML())`

When I generate  a `wsSecurityCert `with this configuration :
```
    const wsSecurityCert = new soap.WSSecurityCert(key, cert, keystoreSslName, {
      signatureAlgorithm: 'http://www.w3.org/2000/09/xmldsig#rsa-sha1',
      signatureTransformations: ['http://www.w3.org/2001/10/xml-exc-c14n#'],
      hasTimeStamp: false
    });
const client = await soap.createClientAsync(wsdlURL, wsdlOptions);
    client.setSecurity(wsSecurityCert);
```

service client valitation reject to me with error `toXML is not a function.`
if you can check it, I appreciate

### LINKED ISSUE #1434: WSSecurityCert: this.security.toXML is not a function after upgrading to 1.6.4
After upgrading from soap@1.1.11 to soap@1.6.4, using WSSecurityCert throws the error: **this.security.toXML is not a function**                                                                                                                                                             
                                                                                                                                                                                                    
  Environment:                                                                                                                                                                                      
  - soap version: 1.6.4                                                                                                                                                                             
  - Node.js version: 22.x                                                                                                                                                                           
  - OS: Linux (also reproducible on macOS)     

Steps to reproduce:                                                                                                                                                                               
                                                                                                                                                                                                    
```
  import * as soap from "soap";                                                                                                                                                                     
                                                                                                                                                                                                    
  const client = await soap.createClientAsync(wsdlUrl);                                                                                                                                             
                                                                                                                                                                                                    
  const security = new soap.WSSecurityCert(privateKey, publicCert, password, {                                                                                                                      
    hasTimeStamp: true,                                                                                                                                                                             
    signatureAlgorithm: "http://www.w3.org/2000/09/xmldsig#rsa-sha1",                                                                                                                               
    digestAlgorithm: "http://www.w3.org/2000/09/xmldsig#sha1",                                                                                                                                      
    signatureTransformations: ["http://www.w3.org/2001/10/xml-exc-c14n#"],                                                                                                                          
  });                                                                                                                                                                                               
                                                                                                                                                                                                    
  client.setSecurity(security);                                                                                                                                                                     
                                                                                                                                                                                                    
  // Any SOAP method call triggers the error                                                                                                                                                        
  await client.SomeMethodAsync(params);     
```                                                                                                                                                        
 The SOAP request should be signed with WS-Security and sent successfully (works in 1.1.11), but it throws `TypeError: this.security.toXML is not a function.`                                                                                                                                           
 Looking at the ISecurity interface, toXML() is documented as "not executed if postProcess is also defined." It appears that WSSecurityCert in 1.6.4 implements postProcess() instead of toXML(), but there's code internally that still calls toXML() directly, causing this error.

w666:
close #1434
close #1422

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -407,6 +407,9 @@ export class Client extends EventEmitter {
     if (this.security && this.security.addOptions) {
       this.security.addOptions(options);
     }
+    if (this.security && typeof this.security?.toXML !== 'function') {
+      this.security.toXML = () => '';
+    }
 
     if (style === 'rpc' && (input.parts || input.name === 'element' || args === null)) {
       assert.ok(!style || style === 'rpc', 'invalid message definition for document style binding');
--- a/test/security/WSSecurityCert.js
+++ b/test/security/WSSecurityCert.js
@@ -1,7 +1,11 @@
 'use strict';
 
+var soap = require('../..');
+
 var fs = require('fs'),
   join = require('path').join;
+
+var assert = require('assert');
 
 describe('WSSecurityCert', function () {
   var WSSecurityCert = require('../../').WSSecurityCert;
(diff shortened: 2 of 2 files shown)


smokhov:
> #close 1434

Should say:

- Closes #1434 
- Closes #1422 

as both cover the same case.

PR #1447, opened 2026-02-16
Replace get-stream and strip-bom with local impl

### LINKED ISSUE #1446: Cannot run with jest since version 1.6.5
Hi, 

Since version 1.6.5, I cannot make jest + ts-jest work. The issue might be a jest configuration, but I cannot figure what is missing by myself. 

I have the following error: 

```
    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation, specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    .../node_modules/get-stream/source/index.js:1
    import {on} from 'node:events';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module
```

This is most probably linked to get-stream upgrade.

To reproduce, I created a new npm folder, with jest, ts-jest, soap and the following test file: 

```
import * as soap from 'soap';

export const create = () => soap.createClientAsync('https://localhost:8080').then(console.log)
```

Any idea on how to solve this?

w666:
Close #1446

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,9 +14,7 @@
         "debug": "^4.4.3",
         "follow-redirects": "^1.15.11",
         "formidable": "^3.5.4",
-        "get-stream": "^9.0.1",
         "sax": "^1.4.1",
-        "strip-bom": "^5.0.0",
         "whatwg-mimetype": "4.0.0",
         "xml-crypto": "^6.1.2"
       },
@@ -803,12 +801,6 @@
       "engines": {
         "node": ">=14"
       }
-    },
-    "node_modules/@sec-ant/readable-stream": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/@sec-ant/readable-stream/-/readable-stream-0.4.1.tgz",
-      "integrity": "sha512-831qok9r2t8AlxLko40y2ebgSDhenenCatLVeW/uBtnHPyhHOvG0C7TvfgecV+wHzIm5KUICgzmVpWS+IMEAeg==",
-      "license": "MIT"
     },
     "node_modules/@shikijs/engine-oniguruma": {
       "version": "3.13.0",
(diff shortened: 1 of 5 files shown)


calvin-cdev:
@w666 thanks for opening this so quickly.
i tried yesterday by installing the branch direcly into my project instead of a tagged version, but the branch doesn't have a build.
i can try later today by cloning the repo, building, and referencing it from my machine.
or if it's easy enough for you, you could publish a pre-release or something.

anyway, yeah i'll try this out later today 👍

w666:
@calvin-cdev

I also replaced strip-boom lib.

w666:
@calvin-cdev,

To test it locally you just need to checkout this branch run `npm i && npm run build` and then `cp -R lib/* PATH_TO_YOUR_PROJECT/node_modules/soap/lib/`

calvin-cdev:
@w666 Okay, i just tested and this branch works for me 🥳 👍

PR #1489, opened 2026-05-07
Fix WSSecurityCert header placement when no SOAP headers exist

### LINKED ISSUE #1487: WSSecurityCert: Security header placed outside Envelope when no other SOAP headers exist
## Description

When using `WSSecurityCert` without any other SOAP headers (`addSoapHeader`), the `<wsse:Security>` block and `<ds:Signature>` are placed outside the `<soap:Envelope>` instead of inside `<soap:Header>`.

## Root Cause

In `client.ts`, the `<soap:Header>` element is only created if `this.security.toXML()` returns a truthy value:

https://github.com/vpulim/node-soap/blob/78a71cba52b134a6a675b755b9daed48ea9766b9/src/client.ts#L447-L458

`WSSecurityCert` does not implement `toXML()`, so the fallback `() => ''` is used (line 358):

https://github.com/vpulim/node-soap/blob/78a71cba52b134a6a675b755b9daed48ea9766b9/src/client.ts#L409-L411

Since `''` is falsy, the Header element is never created. Then `WSSecurityCert.postProcess` looks for `</${envelopeKey}:Header>` to insert the Security block, but it doesn't exist. The `indexOf` returns `-1` and `insertStr` places the content at the wrong position — outside the Envelope.

## Steps to Reproduce

1. Create a SOAP client with `WSSecurityCert` as the only security mechanism.
2. Do NOT call `addSoapHeader()`.
3. Invoke any SOAP method.
4. Inspect the raw XML request — the `<wsse:Security>` block is outside `<soap:Envelope>`.

## Expected Behavior

The `<soap:Header>` element should always be created when a security mechanism with `postProcess` is set, so that `WSSecurityCert.postProcess` can correctly insert the Security block inside it.

## Suggested Fix

Either:

- Have `WSSecurityCert` implement `toXML()` returning a truthy placeholder (e.g. `' '`) that triggers Header creation, or
- Change the condition in client.js to also check for the presence of `postProcess`:

```js
(decodedHeaders || (this.security && (this.security.toXML() || this.security.postProcess))
    ? '<' + envelopeKey + ':Header>...</' + envelopeKey + ':Header>'
    : '')
```

## Workaround

```js
const security = new soap.WSSecurityCert(privatePEM, publicPEM, password, options);
security.toXML = () => ' ';
client.setSecurity(security);
```

## Environment

- `soap` version: 1.9.1
- Node.js version: 22.x
- OS: Windows

---

rubenaguadoc:
Fixes #1487

When using `WSSecurityCert` without additional SOAP headers, the generated security header could be placed outside the SOAP `Envelope`, resulting in invalid XML structure.

This change ensures the security header is always inserted inside the SOAP `Envelope`, regardless of whether other SOAP headers are present.

Includes a test covering the scenario where no additional SOAP headers exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced SOAP security header generation with improved fallback handling to support edge cases.

* **Tests**
  * Added test coverage verifying correct security header insertion in SOAP requests.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -444,7 +444,7 @@ export class Client extends EventEmitter {
       encoding +
       this.wsdl.xmlnsInEnvelope +
       '>' +
-      (decodedHeaders || (this.security && this.security.toXML())
+      (decodedHeaders || (this.security && (this.security.toXML() || this.security.postProcess))
         ? '<' +
           envelopeKey +
           ':Header' +
--- a/test/security/WSSecurityCert.js
+++ b/test/security/WSSecurityCert.js
@@ -330,25 +330,38 @@ describe('WSSecurityCert', function () {
     xml.should.containEql(instance.signer.getSignatureXml());
   });
 
-  it('can handle undefined toXML in WSSecurity', async function () {
+  it('should produce valid XML with WSSecurityCert when no SOAP headers are added', async function () {
     const baseUrl = 'http://localhost:80';
+    const client = await soap.createClientAsync(__dirname + '/../wsdl/default_namespace.wsdl', {}, baseUrl);
 
-    try {
-      const client = await soap.createClientAsync(__dirname + '/../wsdl/default_namespace.wsdl', {}, baseUrl);
+    const security = new soap.WSSecurityCert(key, cert, '', {
+      hasTimeStamp: true,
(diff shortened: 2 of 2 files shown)


w666 (approved):

w666:
@rubenaguadoc thanks for the fix, sorry for the late review.

PR #1496, opened 2026-06-04
Adds support for <xs:choice minOccurs="0" maxOccurs="unbounded"/> cases.

4name:
That PR fixes <xsd:choice minOccurs="0" maxOccurs="unbounded"/> behavior. Before that, the WSDL marshaller could resolve this only as two (or more) separate type lists. The following JSON object has two separate lists of As and Bs:

Assuming we have the following definitions of some Document:
```
<>
    <!-- Parent document WSDL definition -->
    <xsd:element name="someOtherProperty" type="ct:SomeOtherType"/>
    <xsd:choice minOccurs="0" maxOccurs="unbounded">
        <xsd:element name="a" type="ct:A"/>
        <xsd:element name="b" type="ct:B"/>
    </xsd:choice>
</>
```

// Old behavior
```
{
    // Parent document JSON representation
    someOtherProperty: { ... SomeOtherType-type object ... },
    a: [ { ... A-type object 1 ... }, { ... A-type object 2 ... }, ...],
    b: [ { ... B-type object 1 ... }, { ... B-type object 2 ... }, { ... B-type object 3 ... }, ...],
}
```

that resulted into:
```
<>
    <!-- Parent document XML representation -->
    <someOtherProperty> ... SomeOtherType properties ... </someOtherProperty>
    <a> ... A-type properties 1 ... </a>
    <a> ... A-type properties 2 ... </a>
    <b> ... B-type properties 1 ... </b>
    <b> ... B-type properties 2 ... </b>
    <b> ... B-type properties 3 ... </b>
</>
```

In case we want to preserve the choice children order for the following JSON object:

// New behavior
```
{
    // Parent document JSON representation
    someOtherProperty: { ... SomeOtherType-type object ... },
    '$sequence': [ // <- Note this special transparent key that wraps a flat array object
        a: { ... A-type object 1 ... },
        b: { ... B-type object 1 ... },
        b: { ... B-type object 2 ... },
        a: { ... A-type object 2 ... },
        b: { ... B-type object 3 ... },
    ]
}
```

to be encoded as:
```
<>
    <!-- Parent document XML representation -->
    <someOtherProperty> ... SomeOtherType properties ... </someOtherProperty>
    <a> ... A-type properties 1 ... </a> <!-- Note that As and Bs tags are on the same level with someOtherProperty -->
    <b> ... B-type properties 1 ... </b>
    <b> ... B-type properties 2 ... </b>
    <a> ... A-type properties 2 ... </a> <!-- Note that we can mix As and Bs order now -->
    <b> ... B-type properties 3 ... </b>
</>
```

In case of interfere with $sequence property of some SOAP Service property, the special key name can be replaced with options arrayWithChoiceTag: string value.

## Release Notes

* **New Features**
  * Added `arrayWithChoiceTag` configuration option for customizing XML serialization of polymorphic array elements in SOAP operations.


w666:
Hi @4name. thanks for the PR. I am a bit overcomited atm, but this PR is on my backlog. I will have a look as soon as I can, Thanks.

Also, need to silent that CodeRabbit, too much noise from it.

4name:
Hi @w666! Thank you for the notice, take your time. Please let me know if there is anything that needs to be improved in that PR.

src/wsdl/index.ts
```diff
@@ -750,31 +750,36 @@ export class WSDL {
 
       for (i = 0, n = obj.length; i < n; i++) {
         const item = obj[i];
-        const arrayAttr = this.processAttributes(item, nsContext);
+        const isArrayWithChoiceTagContainer = name === this.options.arrayWithChoiceTag;
+        const arrayAttr = isArrayWithChoiceTagContainer ? '' : this.processAttributes(item, nsContext);
```
w666, line 754: New behaviour should be disabled by default, because it may break someones code. This, unfortunately, will disable default value, but will keep this backwards compatible with current implementation.

src/types.ts
```diff
@@ -99,6 +99,8 @@ export interface IWsdlBaseOptions {
   preserveWhitespace?: boolean;
   /** provides support for nonstandard array semantics. If true, JSON arrays of the form {list: [{elem: 1}, {elem: 2}]} are marshalled into xml as <list><elem>1</elem></list> <list><elem>2</elem></list>. If false, marshalls into <list> <elem>1</elem> <elem>2</elem> </list>. Default: true. */
   namespaceArrayElements?: boolean;
+  /** provides support for array with choice semantics. If array key matches, JSON arrays of the form {$sequence: [{elem: 1}, {elem: 2}]} are marshalled into xml as <elem>1</elem><elem>2</elem>. Default: $sequence. */
```
w666:
```suggestion
  /** provides support for array with choice semantics. If array key matches, JSON arrays of the form {$sequence: [{elem: 1}, {elem: 2}]} are marshalled into xml as <elem>1</elem><elem>2</elem> where $sequence is example value for the option. Disabled if option is not set. Example value: $sequence. */
```

test/wsdl-parse-test.js
```diff
@@ -144,4 +144,151 @@ describe(__filename, function () {
...
+          requestAsXML,
+          '<acme:getDataResponse xmlns:acme="http://test-soap.com/api/mixedsequence" xmlns="http://test-soap.com/api/mixedsequence"><acme:getDataResult><acme:a>0</acme:a><acme:b>10</acme:b><acme:c><acme:id>1</acme:id><acme:value1>test 1</acme:value1></acme:c><acme:d><acme:id>3</acme:id><acme:value2>test 3</acme:value2></acme:d><acme:c><acme:id>2</acme:id><acme:value1>test 2</acme:value1></acme:c></acme:getDataResult></acme:getDataResponse>',
+        );
+
+        done();
+      },
+    );
+  });
+
+  it('should parse complex wsdls with mixed choice and minOccurs maxOccurs with default $sequence', function (done) {
```
w666: This test should be converted to `option is not set` to make sure we keep original behaviour by default.

w666 (changes requested):
Thanks for the PR, changes make sense. Unfortunately we should keep this disabled by default, so it does not break someones code.

4name:
Hi @w666 Thank you for the code review, and sorry for the delayed response. The suggestion makes sense to me. I’ve also updated the documentation, which I had initially overlooked.

Please let me know if there’s anything else I’ve missed.

w666:
Hi @4name, thanks for the fixes, looks good to me.

There is just a linter issue in tests.

4name:
Thank you so much for review!

w666 (approved):

PR #1096, opened 2019-11-08
Allow options in objectToDocumentXML when calling a method 

ndeniche:
### Allows overriding the InputMessage "Request" element for custom elements based on args param

* Options is optional field for `objectToDocumentXML` function
* Replicates `_xml` parameter functionality (sending a custom XML structure as a literal) with JSON arguments.

PR #1509, opened 2026-07-20
Allow options in objectToDocumentXML when calling a method (take 2)

smokhov:
- Closes #1096 

This PR reformats #1096 (by @ndeniche) using current prettier rules and fixes its test to do what it's supposed to do clearing an earlier objection [here](https://github.com/vpulim/node-soap/pull/1096#issuecomment-2820362694), so the test only passes if `overrideBaseElement` is `true`.


test/client-test.js
```diff
@@ -2130,6 +2130,37 @@ describe('Uncategorised', function () {
...
+      },
+      // p3 -- callback
+      function (err, client) {
+        assert.ok(client);
+        assert.ifError(err);
+
+        client.MyService.MyServicePort.MyOperation(
+          { parameter: 'dummy', marameter: 'mummy' },
+          function (err, result, resp, soap, req) {
+            assert.ok(req.indexOf('</Request>') === -1);
```
w666, line 2153: Thanks for fixing this. Just two minor comments, if you have time. Though feel free to pass this to me.

I think need to check that open tag is not present and and it was replaced with expected value.

Also, would be good to check that it does not replace when `overrideBaseElement=false`.

smokhov: @w666 so I've covered the tags and expected content as well as `overrideBaseElement=false` as a separate test

smokhov:
@w666 -- added more coverage as you suggested and also realized that we had no mention of that in the README, so added there a section as well. Over to you.

w666 (approved):

PR #1522, opened 2026-08-25
fix(types): add parseResponseAttachments, keep deprecated alias

### LINKED ISSUE #1521: Typo in parseReponseAttachments, should be parseResponseAttachments
Renaming this would be a breaking change but can probably deprecate the typo but keep it for backwards compatibility?

MeGaurav4:
Closes #1521

Reponse -> Response typo in the public option. Renaming outright
is a breaking change, so this keeps parseReponseAttachments as a
deprecated alias and makes the implementation check either:

  parseResponseAttachments || parseReponseAttachments

Types: adds parseResponseAttachments as primary, marks the old
name @deprecated. Docs and tests updated to the correct spelling;
the old name remains in types and http as an alias check.

Also fixes enconding -> encoding in the encoding option docs.


w666 (changes requested):
@MeGaurav4 thanks for spotting and fixing it. Just one comment from me, could you please revert one couple of the tests changes just to make sure old deplrecatad one is used, and keep other tests use new one. This is just to make sure it is not broken in the future. Thanks.

w666:
Seems like code formatting is broken.

w666:
Hi @MeGaurav4,

I do understand that agentic coding is very tempting thing, but if it does not help you fixing the issues that is prabably not a good thing. 

While in the last commit you fixed the formatting, the main comment left unresolved.

MeGaurav4:
Thanks for the review! Good call on keeping the deprecated alias covered. I kept one wsdl_options.json and one response-encoding test on the old parseReponseAttachments name so the alias stays tested, while the other cases use the correct parseResponseAttachments. Let me know if you would like both wsdl_options kept on the old name or any other tweak.

w666 (approved):

PR #1525, opened 2026-09-06
Fix XML-to-object parsing for WSDL responses where a schema-declared array appears with only one item

### LINKED ISSUE #1523: Regression since 1.3.0: type resolution fails for elements whose type is in another namespace via inline xmlns prefix (WCF WSDLs) — untyped, un-arrayed response parsing
Since **1.3.0** (bisected; 1.2.1 is the last good version, still broken in **1.11.0**), `wsdl.xmlToObject()` fails to resolve an element's schema type when the element declaration references a type in a *different* target namespace through an inline `xmlns` prefix on the element itself:

```xml
<xs:element name="GetItemsResponse" nillable="true" type="q1:ArrayOfItem"
  xmlns:q1="http://example.com/data/"/>
```

This is the standard shape of WSDLs generated by .NET WCF services (`svcutil`/WCF emits `q1`, `q2`, … prefixes declared inline per element, pointing at `http://schemas.datacontract.org/...` namespaces), so any WCF service with data contracts in a separate namespace is affected.

When resolution fails, response parsing silently falls back to untyped mode:
- `maxOccurs="unbounded"` elements are **not wrapped in arrays** (single-element responses become bare objects)
- primitive values are **not type-coerced** (`xs:decimal`/`xs:int`/etc. come back as strings)

Code that navigates `result.Items[0]` or does arithmetic on parsed values breaks. We hit this in production against a WCF payments service.

Suspected cause: the namespace-parsing rework in #1322 ("Speed up parsing with many namespaces", first released in 1.3.0).

## Reproduction

Output of the script below:

```
soap 1.2.1:  {"Item":[{"amountField":12.34,"countField":5}]}   | Item is array: true
soap 1.11.0: {"Item":{"amountField":"12.34","countField":"5"}} | Item is array: false
```

Expected (1.2.1 behavior): typed array. Actual on 1.3.0+: bare object, string values.

<details>
<summary>Self-contained repro script (uses npm aliases soap121 / soap1110)</summary>

```js
// npm i soap121@npm:soap@1.2.1 soap1110@npm:soap@1.11.0
const WSDL = `<?xml version="1.0" encoding="utf-8"?>
<wsdl:definitions name="ItemService"
  targetNamespace="http://example.com/service/"
  xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
  xmlns:soap12="http://schemas.xmlsoap.org/wsdl/soap12/"
  xmlns:xs="http://www.w3.org/2001/XMLSchema"
  xmlns:tns="http://example.com/service/">
  <wsdl:types>
    <xs:schema elementFormDefault="qualified"
      targetNamespace="http://example.com/data/"
      xmlns:xs="http://www.w3.org/2001/XMLSchema"
      xmlns:tns="http://example.com/data/">
      <xs:complexType name="ArrayOfItem">
        <xs:sequence>
          <xs:element minOccurs="0" maxOccurs="unbounded" name="Item" nillable="true" type="tns:Item"/>
        </xs:sequence>
      </xs:complexType>
      <xs:complexType name="Item">
        <xs:sequence>
          <xs:element name="amountField" type="xs:decimal"/>
          <xs:element name="countField" type="xs:int"/>
        </xs:sequence>
      </xs:complexType>
    </xs:schema>
    <xs:schema elementFormDefault="qualified"
      targetNamespace="http://example.com/service/"
      xmlns:xs="http://www.w3.org/2001/XMLSchema">
      <xs:element name="GetItemsRequest" nillable="true" type="q1:ArrayOfItem"
        xmlns:q1="http://example.com/data/"/>
      <xs:element name="GetItemsResponse" nillable="true" type="q2:ArrayOfItem"
        xmlns:q2="http://example.com/data/"/>
    </xs:schema>
  </wsdl:types>
  <wsdl:message name="GetItemsRequest">
    <wsdl:part name="GetItemsRequest" element="tns:GetItemsRequest"/>
  </wsdl:message>
  <wsdl:message name="GetItemsResponse">
    <wsdl:part name="GetItemsResponse" element="tns:GetItemsResponse"/>
  </wsdl:message>
  <wsdl:portType name="IItemService">
    <wsdl:operation name="GetItems">
      <wsdl:input message="tns:GetItemsRequest"/>
      <wsdl:output message="tns:GetItemsResponse"/>
    </wsdl:operation>
  </wsdl:portType>
  <wsdl:binding name="ItemBinding" type="tns:IItemService">
    <soap12:binding transport="http://schemas.xmlsoap.org/soap/http"/>
    <wsdl:operation name="GetItems">
      <soap12:operation soapAction="http://example.com/service/IItemService/GetItems" style="document"/>
      <wsdl:input><soap12:body use="literal"/></wsdl:input>
      <wsdl:output><soap12:body use="literal"/></wsdl:output>
    </wsdl:operation>
  </wsdl:binding>
  <wsdl:service name="ItemService">
    <wsdl:port name="ItemPort" binding="tns:ItemBinding">
      <soap12:address location="https://example.com/service"/>
    </wsdl:port>
  </wsdl:service>
</wsdl:definitions>`;

const XML = `<s:Envelope xmlns:s="http://www.w3.org/2003/05/soap-envelope">
  <s:Body>
    <GetItemsResponse xmlns="http://example.com/service/" xmlns:b="http://example.com/data/">
      <b:Item>
        <b:amountField>12.34</b:amountField>
        <b:countField>5</b:countField>
      </b:Item>
    </GetItemsResponse>
  </s:Body>
</s:Envelope>`;

(async () => {
  for (const [name, mod] of [['1.2.1', 'soap121'], ['1.11.0', 'soap1110']]) {
    const soap = require(mod);
    const client = await soap.createClientAsync(WSDL, { forceSoap12Headers: true });
    const body = client.wsdl.xmlToObject(XML).Body;
    const items = body.GetItemsResponse && body.GetItemsResponse.Item;
    console.log(`soap ${name}:`, JSON.stringify(body.GetItemsResponse), '| Item is array:', Array.isArray(items));
  }
})();
```

</details>

- node-soap: broken in 1.3.0, 1.4.1, 1.11.0; works in 1.2.1
- Node.js: 22.14.0

Happy to test a fix branch — we have a production WCF service pinned to 1.2.1 because of this.

w666:
- The parser was resolving the response element name too aggressively, which dropped the array marker and caused single-item collections to be treated as plain objects instead of arrays.
 - Preserves primitive types for child values, so numeric fields remain numbers instead of being turned into strings.

 Close #1523


w666:
Breaking change was introduced in `70fe10c35a674d91c8399dba328065ce3e184ab5`. We did not notice the issue because there was a test gap. Now covered with tests.

w666:
Okay, no feedback so far, I am pretty confident with these changes, merging.

PR #1184, opened 2022-05-11
Added override constructor to ClientSSLSecurityPFX class

willemruys:
The declaration of [constructor overloads](https://www.typescriptlang.org/docs/handbook/2/classes.html#constructors) for ClientSSLSecurityPFX was missing one constructor definition. As a consequence, `ClientSSLSecurityPFX.d.ts` defined only one constructor:
```
../ClientSSLSecurityPFX.d.ts

export declare class ClientSSLSecurityPFX implements ISecurity {
    private pfx;
    private defaults;
    private passphrase;
    constructor(pfx: string | Buffer, defaults?: any);
    toXML(): string;
    addOptions(options: any): void;
}
```
I've added the constructor definition and the definition file showed both constructor methods.

PR #1137, opened 2021-02-18
Arrays with proper prefix and namespace

vimalraj-a:
When Array has specific prefix and namespace as per the WSDL, the namespace reference is ignored completely. 

The current fix makes the array elements with prefix and namespace when available. (Not only during the first element and don't conflict with **_overrideRootElement_** option)

PR #1529, opened 2026-09-21
Repair skipped tests

smokhov:
This PR re-enables 2 long-skipped tests. To make them more valid, it required to update the validity of the certs in `test/certs`.

- The test about namespaces and arrays was disabled in #1184 (originally introduced in #1137) by removing [PurchaseRequestService.wsdl](https://github.com/willemruys/node-soap/blob/99832a1d4e74991d8d5ea8c48410128834b1ccd1/test/wsdl/PurchaseRequestService.wsdl) and [PurchaseRequestService.xsd](https://github.com/willemruys/node-soap/blob/99832a1d4e74991d8d5ea8c48410128834b1ccd1/test/wsdl/PurchaseRequestService.xsd) apparently because their other referenced `.xsd`s stopped fetching. It simply turns out that Oracle changed the URL from `ecex-test.prc.us2.oraclecloud.com` to `ecex-test.fa.us2.oraclecloud.com`. So I simply restored these files, updated their URLs, and cached local copies in `test/wsdl` for referenced schemas in case when/if  the URL goes bad.
- For `ClientSSLSecurityPFX` test I needed to chage `client-password.pfx` and `client-password.pfx`using `AES-256-CBC` for Node/OpenSSL to accept as the previous one was old and complaining about old PFX data. To enable strict client and server cert tests in that  test, I renewed `agent2-cert.pem` for 10 years. It expired in 2020. The procedure was:
```shell
openssl x509 -x509toreq -in agent2-cert.pem -signkey agent2-key.pem -out temp.csr
openssl x509 -req -days 3650 -in temp.csr -signkey agent2-key.pem -out agent2-cert.pem   -extfile <(printf "subjectAltName=DNS:localhost,DNS:node-soap")
openssl pkcs12 -export   -certpbe AES-256-CBC   -out server-password.pfx   -inkey agent2-key.pem   -in agent2-cert.pem   -passout pass:test2test
openssl pkcs12 -export   -certpbe AES-256-CBC   -out client-password.pfx   -inkey agent2-key.pem   -in agent2-cert.pem   -passout pass:test2test
```
- I've updated the badge coverage

Now there are no more skipped tests.


test/wsdl/PurchaseRequestService.wsdl
```diff
@@ -0,0 +1,1021 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<wsdl:definitions
+     name="PurchaseRequestService"
+     targetNamespace="http://xmlns.oracle.com/apps/prc/po/editDocument/purchaseRequestService/"
+     xmlns:plnk="http://schemas.xmlsoap.org/ws/2003/05/partner-link/"
+     xmlns:errors="http://xmlns.oracle.com/adf/svc/errors/"
+     xmlns:orafault="http://xmlns.oracle.com/oracleas/schema/oracle-fault-11_0"
+     xmlns:tns="http://xmlns.oracle.com/apps/prc/po/editDocument/purchaseRequestService/"
```
w666, line 8: I am not sure we can include anything from oracle into a project under MIT license. They will sue us just because there is a work oracle.

https://docs.oracle.com/en/industries/communications/convergent-charging-controller/15.2/wsdl-rg/web-services-description-language-reference-guide.pdf

smokhov: > I am not sure we can include anything from oracle into a project under MIT license. They will sue us just because there is a work oracle.
> 
> https://docs.oracle.com/en/industries/communications/convergent-charging-controller/15.2/wsdl-rg/web-services-description-language-reference-guide.pdf

So the project would become extremely popular! :)

Alright, joking aside, this file exits in the prior project releases included in #1137 and only removed in #1184 only because the URL changed and they stopped resolving while a contributor was working on an unrelated PR.

Certaintly, can re-remove them, but arrays and namespaces test then needs to be changed to something else. As always suggestiins and direct edits are welcome :)

